PR #137, opened 2025-08-05
Synthetic SearchIndex classes for custom objects

bctiemann:


arthanson (changes requested):
This doesn't handle changes to search weight, for example if a field is marked as searchable, and COs are added, then the field is marked as not-searchable it will still show up in the search index.  Also if you change the weight it doesn't seem to have an effect on existing data.  I think it should?  or is this a known limitation?

arthanson (approved):

PR #156, opened 2025-08-15
Cleanup table name resolution

arthanson:
Cleans up  / centralizes the code to convert content type "tables{id}model" name to CO ID.


bctiemann (changes requested):
See inline comment

PR #179, opened 2025-08-22
Pluralized URL slugs for Custom Object Types

bctiemann:
Introduces `verbose_name` and `slug` fields which allow URL patterns (both for UI and API) to be pluralized, and template usages to adhere to a user-controllable pattern.

`name` is the "internal" lowercase model name (required)
`verbose_name` is the human-readable, capitalized and spaced name, optional
`verbose_name_plural` is the plural version of the above, optional, (templates will use `verbose_name` + "s" if not set)
`slug` is the slugified version of `verbose_name_plural`, with a `SlugField` and picker similar to that used in Sites


netbox_custom_objects/models.py
```diff
@@ -164,9 +164,15 @@ class CustomObjectType(PrimaryModel):
     _through_model_cache = (
         {}
     )  # Now stores {custom_object_type_id: {through_model_name: through_model}}
-    name = models.CharField(max_length=100, unique=True)
+    name = models.CharField(
+        max_length=100,
+        unique=True,
+        help_text=_("Internal lowercased object name, e.g. \"vendor_policy\""),
+    )
```
arthanson, line 171: It says lowercased but there is no validation or conversion on this, should we enforce that?

netbox_custom_objects/models.py
```diff
@@ -164,9 +164,15 @@ class CustomObjectType(PrimaryModel):
...
     schema = models.JSONField(blank=True, default=dict)
+    verbose_name = models.CharField(max_length=100, blank=True)
     verbose_name_plural = models.CharField(max_length=100, blank=True)
+    slug = models.SlugField(max_length=100, null=True, unique=True, db_index=True)
```
arthanson, line 175: I think we can add the help_text here?  same comment above, do we want to enforce lowercase?

Also - do we want this auto-generated off of name (can change but gives a default)?

arthanson (approved):

PR #185, opened 2025-08-25
Additional updates to use slug instead of name in URL/view resolution

bctiemann:
Some additional fixes to URL/view resolvers using `slug` instead of `name`.


netbox_custom_objects/templates/netbox_custom_objects/customobject.html
```diff
@@ -19,7 +19,7 @@
 {% endblock breadcrumbs %}
 
 {% block object_identifier %}
-{{ object|meta:"app_label" }}.{{ object.custom_object_type.name }}:{{ object.pk }}
+{{ object|meta:"app_label" }}.{{ object.custom_object_type.slug }}:{{ object.pk }}
```
bctiemann, line 22: I'm not sure if this one is correct

arthanson (approved):

PR #209, opened 2025-09-05
Refresh stale model if field is missing in get_field_value template tag

bctiemann:
In the `get_field_value` template tag, If `field` is not present on `obj`, generate a fresh version of the model from the current state and retry the operation.


mrmrcoleman:
Tested at:

```
NetBox: 4.4
COs: commit 9c84f5d7f54b6edae3664140d90e6a39881f65aa (HEAD -> 189-stale-model-in-template, origin/189-stale-model-in-template)
```

---

Screenshot 2025-09-06 at 11 05 05

PR #314, opened 2025-12-04
Fixes: #313 - Cap related custom objects table size

### LINKED ISSUE #313: "Custom Objects linking to this object" should be capped
### Plugin Version

0.4.1

### NetBox Version

4.4.7

### Python Version

3.12

### Steps to Reproduce

View the detail page for any object that has multiple Custom Objects linking to it via Object or Multiobject fields

### Expected Behavior

The "Custom Objects linking to this object" table should be capped at a reasonable length.

### Observed Behavior

There is no limit on the table length, so the rendered page can be arbitrarily large.

bctiemann:
### Fixes: #313

Adds a cap of 20 rows to the "Custom Objects linking to this object" table.

### DIFF
--- a/netbox_custom_objects/template_content.py
+++ b/netbox_custom_objects/template_content.py
@@ -48,7 +48,7 @@ class CustomObjectLink(PluginTemplateExtension):
         linked_custom_objects = []
         for field in custom_object_type_fields:
             model = field.custom_object_type.get_model()
-            for model_object in model.objects.all():
+            for model_object in model.objects.all()[0:19]:
                 model_field = getattr(model_object, field.name)
                 if model_field:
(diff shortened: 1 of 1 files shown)


arthanson (changes requested):
I think a problem with this is there is no indication that there are more linked objects that aren't shown. There is no indication of a limit, they just aren truncated. Also, there is no easy way then to get to the other linked objects that aren't shown as there is currently no way to filter the custom-object table via a linked object param.

They do now appear in the Related Objects table as a count- but we don't filter the results of the linked custom-object table as we don't dynamically create filters for this, so there is no way currently beyond this table to get to all the linked objects.  If we could get the filters working it seems more consistent to remove this table entirely.

arthanson (approved):

PR #316, opened 2025-12-05
Fixes: #224 - Use a ContentTypeField in CustomObjectTypeFieldSerializer

### LINKED ISSUE #224: `app_label` and `model` on API POST `custom-object-type-fields` is redundant
### Plugin Version

0.3.1

### NetBox Version

4.4.0

### Python Version

3.12.3

### Steps to Reproduce

API endpoint POST `custom-object-type-fields` requires `model` and `app_label`, despite at the same time requiring `related_object_type` which is an ID. `model` and `app_label` can be instead determined by the `related_object_type`, based on DB table `django_content_type`. I believe this is how it's done for other endpoints.

```python
custom_object_field_relational = {
    'custom_object_type': custom_obj1_id,
    'name': 'related_obj2',
    'label': 'Related Obj 2',
    'type': 'object',
    'primary': False,
    'related_object_type': custom_obj2_id,
    # Paremeters below should not be needed, as they can be determined by 'related_object_type'
    'app_label': 'netbox_custom_objects',
    'model': f'table{custom_obj1_id}model',
}
```

To reproduce:
1. Create Custom Object1
2. Create Custom Object2
3. Try creating via API a field for Custom Object1 that related to Custom Object2

### Expected Behavior

Expected behavior:
API call does not need `app_label` and `model`

### Observed Behavior

Observed behavior:
`app_label` and `model` need to be passed in the call

bctiemann:
### Fixes: #224

Aligns the `CustomObjectTypeFieldSerializer` with the behavior of `CustomFieldSerializer` by allowing the user to specify a `related_object_type` in the POST body rather than `app_label` and `model` fields:

```
{
    "custom_object_type": 26,
    "type": "multiobject",
    "name": "test2",
    "default": [14],
    "related_object_type": "dcim.device"
}
```

### DIFF
--- a/netbox_custom_objects/api/serializers.py
+++ b/netbox_custom_objects/api/serializers.py
@@ -4,6 +4,7 @@ import sys
 from core.models import ObjectType
 from django.contrib.contenttypes.models import ContentType
 from extras.choices import CustomFieldTypeChoices
+from netbox.api.fields import ContentTypeField
 from netbox.api.serializers import NetBoxModelSerializer
 from rest_framework import serializers
 from rest_framework.exceptions import ValidationError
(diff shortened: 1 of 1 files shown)


bctiemann:
This is technically a breaking change, but since #224 is characterized (correctly) as a bug this is really just aligning to "correct" (core) behavior by using the NetBox-standard `ContentTypeField`.

arthanson (approved):

PR #324, opened 2025-12-10
Fixes #310 - Linkify primary field

### LINKED ISSUE #310: Linkify primary field in custom object table view
### Plugin Version

0.4.1

### Proposed functionality

Currently to enter custom objects detailed view need to hit object id in custom objects table, better approach would be to always display primary key field in table view in a first column and linkify primary field with custom object.

### Use case

To make it easier to navigate custom objects table

### External dependencies

_No response_

bctiemann:
### Fixes: #310

Introduces a `render_table_column_linkified` method on the base `FieldType` which during dynamic table generation is attached to the "primary" column and provides a "linkified" value which leads to the target Custom Object.

Upshot is that in a Custom Object list view table, the "primary" field (i.e. "name" below) is automatically made a link, making it easier to create a customized table view with clickable object names.

Screenshot 2025-12-10 at 9 06 41 AM

**Note:**
For all non-primary fields, custom column renderers (following the `render_foo` pattern of django-tables2) are still used for overriding the displayed value for multiobject, JSON, etc. Note that if a JSON or other formatted field is made the "primary", formatting will be lost due to the `render_table_column_linkified` behavior taking precedence.

### DIFF
--- a/netbox_custom_objects/field_types.py
+++ b/netbox_custom_objects/field_types.py
@@ -110,6 +110,9 @@ class FieldType:
     def get_table_column_field(self, field, **kwargs):
         raise NotImplementedError
 
+    def render_table_column_linkified(self, record):
+        return linkify(record)
+
     def after_model_generation(self, instance, model, field_name): ...
 
     def create_m2m_table(self, instance, model, field_name): ...
--- a/netbox_custom_objects/views.py
+++ b/netbox_custom_objects/views.py
@@ -127,9 +127,13 @@ class CustomObjectTableMixin(TableMixin):
                         field.name
                     )
                 )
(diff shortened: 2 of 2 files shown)


arthanson (changes requested):
I'd argue that boolean, json, multi-object, object, multi-select, select should not be allowed to be primary name field, maybe run by Mark but (unless this was already discussed) but it seems strange and the loss of formatting just makes it more so.

If we do change this I'd suggest maybe re-arranging the edit form so "Type" field comes after name and depending on the toggle the "Primary name field" is shown or not (or disabled).

arthanson (approved):

PR #361, opened 2026-01-10
#296 filter custom type fields

### LINKED ISSUE #296: Add filter by attributed for custom-object-types
### Plugin Version

v0.4.1 - 2025-11-10

### Proposed functionality

### Use case

### External dependencies

_No response_

ifoughal:
Fixes #296
This pull request enhances custom field filtering support in NetBox custom objects by adding filter form fields for several custom field types, including decimal, object relationship, and multi-object relationship fields.
**Filter form field support for custom field types:**

* Added `get_filterform_field` to `DecimalFieldType`, returning a `forms.DecimalField` for filtering decimal custom fields.
* Implemented `get_filterform_field` in `ObjectFieldType` and `MultiObjectFieldType`, returning `DynamicModelChoiceField` and `DynamicModelMultipleChoiceField` respectively, allowing filtering by related objects.

**Dynamic filterset generation:**

* Updated `get_filterset_class` in `filtersets.py` to dynamically add Django filter fields for each custom field, selecting the appropriate filter type (e.g., `CharFilter`, `NumberFilter`, `BooleanFilter`, `DateFilter`, `ModelChoiceFilter`) based on the field's type.

### DIFF
--- a/netbox_custom_objects/filtersets.py
+++ b/netbox_custom_objects/filtersets.py
@@ -71,7 +71,6 @@ def get_filterset_class(model):
             attrs[field.name] = django_filters.DateFilter(field_name=field.name, lookup_expr='exact', label=field.label)
         elif field.type == CustomFieldTypeChoices.TYPE_URL:
             attrs[field.name] = django_filters.CharFilter(field_name=field.name, lookup_expr='icontains', label=field.label)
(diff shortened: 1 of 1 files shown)


ifoughal:
reopening a new PR

PR #365, opened 2026-01-12
296 custom type object dynamic fields filter

### LINKED ISSUE #296: Add filter by attributed for custom-object-types

ifoughal:
closes #296

### DIFF
--- a/netbox_custom_objects/filtersets.py
+++ b/netbox_custom_objects/filtersets.py
@@ -88,8 +88,7 @@ class CustomObjectTypeFilterSet(NetBoxModelFilterSet):
 
 
 def build_filter_for_field(field) -> Optional[django_filters.Filter]:
-    spec = FIELD_TYPE_FILTERS.get(field.type)
-    if not spec:
+    if not (spec := FIELD_TYPE_FILTERS.get(field.type)):
         return None
 
     queryset = None


netbox_custom_objects/field_types.py
```diff
@@ -799,7 +820,18 @@ def get_form_field(self, field, for_csv_import=False, **kwargs):
             )
 
     def get_filterform_field(self, field, **kwargs):
-        return None
+        """
```
arthanson: Can probably move this and the one above in ObjectFieldType to a new base class mixin, bit better for DRY.

netbox_custom_objects/filtersets.py
```diff
@@ -13,6 +16,67 @@
...
+
+
+FIELD_TYPE_FILTERS = {
+    CustomFieldTypeChoices.TYPE_TEXT: FilterSpec(django_filters.CharFilter, lookup_expr="icontains"),
+    CustomFieldTypeChoices.TYPE_LONGTEXT: FilterSpec(django_filters.CharFilter, lookup_expr="icontains"),
+    CustomFieldTypeChoices.TYPE_INTEGER: FilterSpec(django_filters.NumberFilter, lookup_expr="exact"),
+    CustomFieldTypeChoices.TYPE_DECIMAL: FilterSpec(django_filters.NumberFilter, lookup_expr="exact"),
+    CustomFieldTypeChoices.TYPE_BOOLEAN: FilterSpec(django_filters.BooleanFilter),
+    CustomFieldTypeChoices.TYPE_DATE: FilterSpec(django_filters.DateFilter, lookup_expr="exact"),
+    CustomFieldTypeChoices.TYPE_DATETIME: FilterSpec(django_filters.DateTimeFilter, lookup_expr="exact"),
```
arthanson: Not sure on here - normally we do something like:
    created = django_filters.DateTimeFilter()
    created__before = django_filters.DateTimeFilter(
        field_name='created',
        lookup_expr='lte'
    )
    created__after = django_filters.DateTimeFilter(
        field_name='created',
        lookup_expr='gte'
    )
the base does not have a lookup_expr.  if we could add the __before, __after that would be great but probably beyond the scope of this - maybe add that as a FR.

PR #410, opened 2026-02-19
Closes: #193 - Allow Custom Object Types to specify a group_name for navigation menu grouping

### LINKED ISSUE #193: Grouping custom objects
### Plugin Version

0.2.0

### Proposed functionality

The current UI structure has been split into 2 sections:

- **Object types**: custom object and fields definitions
- **Objects**: instances of custom objects

Image  

The proposed enhancement introduces the concept of custom object groups, enabling a hierarchical structure among custom objects to improve organisation and usability.  

Image

This can be achieved by capturing the group name when defining the custom object.

Image

### Use case

Creating multiple custom object types with similar names but distinct purposes can easily lead to confusion for users.

For instance, we’re currently evaluating how to structure our backup documentation to support disaster recovery efforts. Most organisations implement backups across several layers, such as network configurations, virtual machine images, and application data like databases or web directories. Each layer serves a specific function and typically requires its own documentation format.

Attempting to capture all these scenarios within a single `Backup` custom object by adding numerous (but rarely used) fields is not consistent with object-oriented design principles. This approach quickly becomes complex, difficult to maintain, and hard for users to interpret.

A more effective solution is to define multiple, purpose-specific custom objects and group them accordingly. This would allow us to:

- Clearly differentiate between backup types and their intended use
- Maintain consistency and clarity across documentation
- Simplify navigation and reduce ambiguity for users
- Accurately associate backup objects with the relevant device, service, or virtual machine

As you can imagine, the more we focus on object-oriented principles, the more custom objects definitions we need to create. The more we create, the more complex the solution becomes for users.

### External dependencies

_No response_

bctiemann:
### Closes: #193

This adds a `group_name` field to `CustomObjectType` which allows COTs to be grouped in the navigation menu. An unspecified `group_name` results in COTs being grouped under "Objects" as before.

Screenshot 2026-02-19 at 6 04 47 PM

**Note**: This solution requires that NetBox be restarted in order to re-populate the groupings in the nav menu if they have changed. This is because only the list of items within a group is iterable; the groupings themselves are populated concretely at app startup time.


arthanson (changes requested):
Failing test for some reason: django.db.utils.ProgrammingError: relation "netbox_custom_objects_customobjecttype" does not exist

arthanson (changes requested):
Couple issues:

I had two existing custom object types, I added a a1 group label to the first, b2 to the second one.

1. When I add b2 the custom object in the sidebar disappears - forcing you to restart.  I know you stated that it requires a restart to show, but this forces a restart as soon as you add a group name.  I don't think that is acceptable, or if okay with PM then should at least have a big warning on the field that you must restart if you change this item - IMHO I wouldn't ship with this restriction.

2. The ordering of the menu items is non-deterministic.  Each time you restart the group a1 and b2 can shift places.  a1 is first on some restarts, b1 is first on others.

3. Should probably add group-name to filters

I also think there should be some minimal tests:
- A test that a COT with group_name appears in the correct group in get_grouped_menu_items()
- A test that group_name is serialized in the API response

arthanson (approved):

Issue #366, opened 2026-01-13
Add filtering by objects, boolean and selection


ultrapista:
### Plugin Version

0.4.4

### Proposed functionality

Currently the filtering of custom objects only work for attributes of type **Text** or **Integer**. If the attributes are of type **Object**, **Selection** or **Boolean**, they are not displayed in the object filtering form. I propose to add the ability to filter by these attributes.

<img width="262" height="144" alt="Image" src="https://github.com/user-attachments/assets/52a748bf-188a-44cb-8680-85adc2d0304b" />

<img width="423" height="227" alt="Image" src="https://github.com/user-attachments/assets/1736e275-72e8-4e40-8a84-dedaba0f83b5" />

### Use case

This would allow the user to better filter the instances that he wants to see.

### External dependencies

N/A

PR #437, opened 2026-03-27
Closes: #296/#366 - Add filter by attributed for custom-object-types

### LINKED ISSUE #296: Add filter by attributed for custom-object-types

arthanson:
### Fixes: #296 #366

From @ifoughal original PR: #365 - added tests and cleanup.

This pull request adds dynamic filtering capabilities for custom object type fields, allowing users to filter custom objects based on their field values via the API and UI.

  Changes:
  - Introduced a FilterSpec dataclass and field type-to-filter mapping system in filtersets.py
  - Added ArrayContainsFilter for multiselect fields, fixing filtering on PostgreSQL ArrayField
  - Implemented get_filterform_field for Decimal, Object, and MultiObject field types
  - Fixed incorrect module name from "database.filtersets" to "netbox_custom_objects.filtersets"
  - Added test coverage for all custom field types

Monosnap Cot1 | NetBox 2026-04-16 16-15-13
Monosnap Cot1s | NetBox 2026-04-16 16-15-29

Also needed to add new CustomObjectAPITestCaseMixin as previously was using NetBox APIViewTestCases.APIViewTestCase which plugins shouldn't do - this was causing errors from NetBox recent test changes.

### DIFF
--- a/.github/workflows/lint-tests.yaml
+++ b/.github/workflows/lint-tests.yaml
@@ -34,7 +34,7 @@ jobs:
         run: ruff check
   tests:
     runs-on: ubuntu-latest
-    timeout-minutes: 10
+    timeout-minutes: 20
     strategy:
       fail-fast: false
       matrix:
--- a/netbox_custom_objects/field_types.py
+++ b/netbox_custom_objects/field_types.py
@@ -256,8 +256,6 @@ class DecimalFieldType(FieldType):
         return forms.DecimalField(
             label=field,
             required=False,
-            max_digits=12,
-            decimal_places=2,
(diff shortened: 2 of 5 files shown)


arthanson:
@claude /review

bctiemann:
@claude /review

bctiemann:
Fixed items 1-4.

bctiemann:
Added unit tests per item 5.

bctiemann (approved):

PR #442, opened 2026-04-02
Closes: #31 - Polymorphic object fields

### LINKED ISSUE #31: Allow GenericForeignKey Custom Object Type Fields
### Plugin Version

commit 722bbd624c8af515dc584b719e2864cbfcc14dd0 (HEAD -> public-preview, origin/public-preview) Author: btiemann [btiemann@netboxlabs.com](mailto:btiemann@netboxlabs.com) Date:   Thu Jun 26 15:31:01 2025 -0400

### Proposed functionality

Multiple Private Preview participants have requested the ability to use GenericForeignKeys for Custom Object Type fields

* Currently Custom Object Type fields can only point to one type
* It is common for users to to want to model "polymorphic" type behaviour. e.g. a Custom Object Type "dataflow" with fields of type Multiobject but that could point to either IPAddresses, IPRanges, or Prefixes

***Note** We should add an index across the fields* [https://github.com/netbox-community/netbox/blob/develop/docs/development/extending-models.md](https://github.com/netbox-community/netbox/blob/develop/docs/development/extending-models.md)

### Use case

* User creates a Custom Object Type X
* User adds a field Y of type Object or MultiObject
* User can select one or more related types instead of just one

### External dependencies

*No response*

bctiemann:
### Closes: #31 

This PR addresses a major feature addition: polymorphic object/multi-object fields (using a `GenericForeignKey` pattern). This means that the related object type for these fields can be any of a selected list, rather than just a single related object type.

Screenshot 2026-04-01 at 8 43 41 PM

Editing a custom object with such a field shows an individual field in the edit form for each object type supported by the object/multiobject field, allowing the selection lists to work with existing object-type-based selection lists: 

Screenshot 2026-04-01 at 8 44 37 PM

Unit tests have been added covering all UI and API routes, including validating that only the selected related object types can be assigned to the field.

**NOTE:**

```
        # KNOWN LIMITATION: polymorphic fields (is_polymorphic=True) store their
        # allowed target types on the related_object_types M2M, not on the
        # related_object_type FK.  This DFS therefore does not traverse edges
        # introduced by polymorphic fields.  A cycle that passes entirely through
        # polymorphic legs (e.g. A →(poly) B →(poly) A) will go undetected.
        #
        # Fixing this requires also iterating field.related_object_types.filter(
        # app_label=APP_LABEL) and recursing into each.  The check_polymorphic_recursion
        # signal already guards the direct A→B assignment, but cannot see multi-hop
        # cycles that depend on polymorphic fields already on intermediate types.
```


bctiemann:
@claude /review

bctiemann:
@claude /review

arthanson (changes requested):
The caching stuff and extract_cot_id_from_model_name can be separated out into a separate PR as they are functionally separate and would make this very large PR easier to review / more targeted.

suppress_clear_cache is only used in utilities.py and can be a private method.  install_clear_cache_suppressor is only Called in __init__ so that code can be separated out into another PR (the stuff in utilities and __init__).

bctiemann:
- Moved `cot_field_id` resolution and cache clearing out into an (already-merged) other PR
- Renumbered migrations
- Added missing internationalization
- Fixed several other gaps identified via local review post merging `feature` into this branch

arthanson (changes requested):
feature now has the changes for the tests to pass but has merge conflicts.

arthanson (changes requested):
This is returning double form-fields for items (see two form items for Circuits, two for Region, etc..):

Image

Also, I think a better UI for this would be like we do for scope (see below) where we have a dropdown for the type and dynamically switch the field under it. Also in the list view probably want to show the Poly Type next to Poly so you know what it actually is.

Image

Claude also came up with a couple points:

1. through_table_name can exceed PostgreSQL's 63-char identifier limit

 IMPORTANT — data integrity risk on large deployments.

 through_table_name is computed as:
 f"custom_objects_{self.custom_object_type_id}_{self.name}"

 The name field allows up to 50 characters (max_length=50). A realistic worst-case (e.g. COT id = 99999, name = 50 chars): 15 + 5 + 1 + 50 = 71 chars — 8 over PostgreSQL's hard 63-byte
 limit. PostgreSQL silently truncates identifiers, which can cause silent name collisions between tables sharing a long prefix.

 _safe_index_name() already exists in field_types.py for this purpose. An analogous _safe_table_name() should be applied to through_table_name (and correspondingly to
 through_model_name).

 Note: This doesn't require a data migration for any existing installations that haven't hit the limit yet, but should be addressed before GA.

 ---
 2. Orphaned field row if check_polymorphic_recursion signal raises

 The save path for a new polymorphic field via API:

 1. CustomObjectTypeField.save() — creates DB columns/through table (committed, or at least schema-altered)
 2. DRF create() → instance.related_object_types.set(resolved) — fires m2m_changed
 3. check_polymorphic_recursion (pre_add) raises ValidationError → DRF surfaces 400

 At step 3, the field row + DB schema already exist, but related_object_types is never set. The schema is in a partially-initialized state: the GFK columns (or through table) exist, but
  the field instance has no allowed types.

 The serializer's validate() catches some problems early, but it does NOT check for circular references (it can't — it runs before any DB writes). The signal is the only guard.

 Mitigation options:
 - Wrap the save() + set() in the DRF create() in an atomic block, catching ValidationError from the signal and deleting the field row in the exception handler.
 - Or document clearly that related_object_types.set() must succeed for the field to be usable, and add a check in get_model() that skips is_polymorphic fields with zero
 related_object_types.

bctiemann:
We now use the "scope type" style context-aware picker for single object polymorphic types:

Screenshot 2026-04-28 at 9 07 13 PM

Also, related inputs for a polymorphic field are grouped together under an `h3` subheading in the edit/bulk edit forms. Feedback from the above comments are also addressed.

arthanson (changes requested):
Several brought up by Claude review - the double save would crate two ObjectChange items so would be good if we could address that.

HIGH

  Double-save window for GFK in form custom_save — views.py ~858

  custom_save() calls instance.save() to create the row, then does a second instance.save() to write the poly GFK values. Between saves, the row exists in the DB with
  poly_obj_content_type_id = NULL. The fix is to set GFK attributes on the instance before the first save() call — the same way the API serializer's create() already does it.

  ---
  MEDIUM

  Nested connection.schema_editor() inside an outer one — field_types.py ~1321

  CustomObjectTypeField.save() already holds an outer with connection.schema_editor() context, then create_polymorphic_m2m_table opens its own. On PostgreSQL, exiting the
  inner editor flushes deferred SQL prematurely. The drop_polymorphic_m2m_table method already accepts schema_editor as a parameter — create_polymorphic_m2m_table should
  match that pattern.

  _original_is_polymorphic uses __dict__ instead of attribute access — models.py ~1134

  self.__dict__.get("is_polymorphic", False) returns False if the field is deferred. self.is_polymorphic is the safe accessor. Inconsistent with how _original_name is set.

  Missing ValueError/TypeError catch on ContentType lookup — serializers.py ~82

  ContentType.objects.get(pk=data["content_type_id"]) is user-controlled input with no type validation. A non-integer value raises ValueError/TypeError which surfaces as a
  500. The model_class.objects.get() call below correctly catches these — the ContentType lookup above does not.

  Two extra .filter() queries per serializer generation — serializers.py ~368

  get_serializer_class() builds _poly_obj_fields and _poly_m2m_fields via two extra DB queries per request. The main model_fields queryset is already fetched on line 340 —
  derive the poly sets from that in Python instead.

  ListField missing allow_null for non-required MultiObject fields — field_types.py ~1078

  A PATCH sending null for a non-required polymorphic MultiObject field fails with a 400 because the ListField wrapper doesn't set allow_null=not field.required. The
  single-object path correctly sets this on line ~606.

arthanson (changes requested):
Changes per mark - need to list the type of field next to each item on the detail view.

arthanson (approved):

bctiemann:
@claude /review all changes since the last review

bctiemann:
Issues 2 and 3 have been addressed. The rest are low-impact and left as-is.

arthanson (changes requested):
If you create a poly field, link it to an object then go to that object it doesn't show that it is linked.

Note I have Butler Communications linked in this custom object:
Image

But Butler Communications is showing now custom objects linked:
Image

arthanson (approved):

PR #444, opened 2026-04-02
Fixes: #440 - Support typeahead search for integer/decimal (and select/multiselect) fields

### LINKED ISSUE #440: Typeahead search returns no results for non-text primary fields
### Plugin Version

0.4.6 (also confirmed on current `main`)

### NetBox Version

4.4.10

### Python Version

3.12

### Steps to Reproduce

1. Create a Custom Object Type called "netarea"
2. Add an Integer field called "Major" and set it as the primary field
3. Create a netarea instance with Major = 311 (it displays as "311" in the UI)
4. On any core NetBox model (e.g. Prefix), create a Custom Field of type "Object" that references the "netarea" custom object type
5. Edit a Prefix and click on the custom field dropdown to assign a netarea
6. Type "311" into the typeahead search box

### Expected Behavior

The typeahead should find and display the netarea with Major = 311, since "311" is the value the object displays as.

### Observed Behavior

"No results found" is displayed. The search method in `filtersets.py` (lines 57-69 in `get_filterset_class`) only queries fields of type `TYPE_TEXT`, `TYPE_LONGTEXT`, `TYPE_JSON`, and `TYPE_URL` using `icontains`. Integer, decimal, boolean, date, and selection fields are completely excluded from the `Q()` filter. When no text fields exist (or none match), the `Q()` is empty and `.filter(Q())` returns no results.
```python
# filtersets.py lines 57-69 — only text-type fields are searched
for field in model.custom_object_type.fields.all():
    if field.type in [
        CustomFieldTypeChoices.TYPE_TEXT,
        CustomFieldTypeChoices.TYPE_LONGTEXT,
        CustomFieldTypeChoices.TYPE_JSON,
        CustomFieldTypeChoices.TYPE_URL,
    ]:
        q |= Q(**{f"{field.name}__icontains": value})

### PR DESCRIPTION
**Pull request by bctiemann**
### Fixes: #440 

**`filtersets.py` — `search()` method in `get_filterset_class()`:**                                                                                                            
  - Added `TYPE_SELECT` and `TYPE_MULTISELECT` to the `icontains` branch (stored as `CharField`, so substring search works)                                                    
  - Added `TYPE_INTEGER` / `TYPE_DECIMAL`: parse the search value as a number and use `__exact` match; silently skip if the value isn't numeric                              
  - Added `TYPE_DATE` / `TYPE_DATETIME`: parse with Django's `parse_date`/`parse_datetime` and use `__exact` match; skip if unparseable                                          
  - Added `if not q: return queryset.none()` — fixes a pre-existing bug where searching with no applicable field types (e.g. non-numeric string against an integer-only COT) returned all objects instead of none                                                                                                                            

Also adds unit tests exercising these filtering behaviors.

### DIFF
--- a/netbox_custom_objects/tests/test_filtersets.py
+++ b/netbox_custom_objects/tests/test_filtersets.py
@@ -2,6 +2,7 @@
 Tests for filtersets used by the plugin's UI and API views.
 """
 import datetime
+from decimal import Decimal
 
 from django.test import TestCase
 
@@ -10,6 +11,7 @@ from dcim.models import Device, DeviceRole, DeviceType, Manufacturer, Site
 from netbox_custom_objects.field_types import MultiObjectFieldType, ObjectFieldType
 from netbox_custom_objects.filtersets import get_filterset_class
 from netbox_custom_objects.models import CustomObjectTypeField
+from extras.models import CustomFieldChoiceSet
 from utilities.forms.fields import (
     DynamicModelChoiceField,
     DynamicModelMultipleChoiceField,
@@ -460,7 +462,6 @@ class DecimalPrimaryFieldSearchTestCase(CustomObjectsTestCase, TestCase):
         )
 
         model = cls.cot.get_model()
-        from decimal import Decimal
         cls.obj_11 = model.objects.create(price=Decimal("1.1"))
         cls.obj_03 = model.objects.create(price=Decimal("0.3"))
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by bctiemann** [comment]
Addressed all findings and added tests for datetime and decimal.

File: netbox_custom_objects/tests/test_filtersets.py [hunk 1]
```diff
@@ -400,3 +402,248 @@ def test_filter_multiple_targets_no_duplicates(self):
...
+        self.assertNotIn(self.obj_03.pk, pks)
+
+
+class SelectPrimaryFieldSearchTestCase(CustomObjectsTestCase, TestCase):
+    """Typeahead search finds objects when the primary field is a Select."""
+
+    @classmethod
+    def setUpTestData(cls):
+        super().setUpTestData()
+        from extras.models import CustomFieldChoiceSet
```
arthanson: move to top

arthanson (approved):
approved - just move the two imports to the top.

PR #445, opened 2026-04-02
Fixes: #441 - Patch ObjectSelectorView to support targeting custom objects from core custom fields

### LINKED ISSUE #441: Object selector (magnifying glass) broken for custom object types referenced from core custom fields
### Plugin Version

0.4.6 (also confirmed on current `main`)

### NetBox Version

4.4.10

### Python Version

3.12

### Steps to Reproduce

1. Create a Custom Object Type (e.g. "netarea")
2. On a core NetBox model (e.g. Prefix), create a Custom Field of type "Object" that references the "netarea" custom object type
3. Edit a Prefix
4. Click the magnifying glass icon next to the custom field to open the full object selector

### Expected Behavior

A selector dialog should open showing netarea objects, with the ability to browse and filter them.

### Observed Behavior

The selector either shows stale content from a previously loaded selector (e.g. "Select IP Address" from another field on the same form), or shows an empty/broken modal. The underlying cause is a 500 error from the HTMX content load.
Two issues combine:
1. Core NetBox's `CustomField.to_form_field()` (`customfields.py` line 593) unconditionally sets `selector=True` for all object-type custom fields. The plugin disables the selector for its own forms (`field_types.py` line 490: `selector=model._meta.app_label != APP_LABEL`), but this only applies to fields defined *within* custom object types — not to core NetBox custom fields on models like Prefix that reference a custom object type.
2. When the magnifying glass is clicked, `ObjectSelectorView._get_form_class()` (`htmx.py` lines 45-50) tries to import `netbox_custom_objects.forms.Table<N>ModelFilterForm` by convention, which does not exist for dynamically generated models. The `import_string` call raises an `ImportError`, causing a 500 error. The Bootstrap modal opens regardless (client-side `data-bs-toggle`), but the HTMX content request fails, leaving stale or empty content in the dialog.

bctiemann:
### Fixes: #441 

Fixes the issue of core Custom Fields being unable to use the Object Selector (magnifying glass) on object fields pointing to Custom Objects (this would previously raise an `ImportError`).

This change patches the `get_form_class` and `get_filterset_class` methods from `ObjectSelectorView` upon load of the `ready` method in Custom Objects. The patched functions come with custom-object-aware `FilterSetForm` and `FilterSet` classes that enable the Object Selector to work.


arthanson (changes requested):
I tink this can be made into a common routine called both here and in `CustomObjectListView.get_filterset_form` to be DRY.

The form-building logic in `_patched_get_form_class` duplicates `CustomObjectListView.get_filterset_form()` (views.py:374-394), with two subtle
  divergences: the class name uses `model.__name__` here vs `model._meta.object_name` in the view, and NotImplementedError is silently swallowed here vs logged at debug in the view.

arthanson (approved):

Issue #388, opened 2026-02-03
Custom Object Type exporter


bctiemann:
### Plugin Version

v0.4.6

### Proposed functionality

**Depends on #386**

When a suitable Custom Object Type schema export format has been developed, there should be functionality built into the `CustomObjectType` class or other utility functions to export a given COT to that format.

### Use case

This functionality is key to further development of COT versioning; exported COT schemas will be used as the input feedstock for the comparator developed in #387 .

PR #449, opened 2026-04-06
Closes: #386 - Exportable schema support

### LINKED ISSUE #386: Define a schema format for portable Custom Object Type definitions
### Plugin Version

v0.4.6

### Proposed functionality

To support Custom Object Type versioning, we need to begin by defining a mechanism for exporting the current definition of a Custom Object Type (and all its fields) to a versioned, portable format such as YAML or JSON, together with parameters capturing all configured attributes of all fields, and a field-order-preserving ID system to ensure that a COT can be imported to a new system identically to its source, as well as to allow upgrade and downgrade of a COT within a system.

This is one proposed example of such an exported COT, in YAML format:

```yaml
# Example: my_custom_type.yaml
name: data_flow
version: "1.2"
verbose_name: Data Flow
description: Tracks data flows between systems
fields:
  - id: 1  # Stable numeric ID for rename detection
    name: source_system
    type: object
    related_object_type: dcim.device
    required: true
  - id: 2
    name: source_system
    type: object
    related_object_type: netbox_custom_objects.table9model
    required: true
  - id: 3
    name: destination_system
    type: object
    related_object_type: dcim.device
    required: true
  - id: 4
    name: protocol
    type: select
    choice_set: protocols
    required: false
```

**Note** that in the above example, the `id: 2` field which points to another Custom Object Type (rather than a NetBox core model) must be expressed in terms of the internal model name, and this name may vary from one NetBox installation to another. To fully port a set of Custom Objects from one system to another, a tiered approach may be necessary:

1. Export/import all Custom Object Types with no field dependencies on other COTs
2. Export/import all COTs with dependencies on the first set but which might be dependencies of another set
3. Export/import all COTs with dependencies on the second set, etc.

The exporter component (see [NPL-715](https://linear.app/netboxlabs/issue/NPL-715/custom-object-type-exporter)) could resolve this dependency tree automatically and group COT definitions into dependency tiers for batched export/import.

### Use case

COTs will change schemas over time, e.g. by adding fields, renaming fields, changing field attributes, removing fields, etc. The long-term ability to support Custom Objects depends on there being a defined mechanism for performing and tracking these changes as they are made, exporting each version to a portable or archivable format, and importing them for use in a new system.

Without this infrastructural support, it will be very difficult for users to properly leverage Custom Objects across multiple NetBox instances. It will also be difficult or impossible for NetBox to be able to apply migrations (schema changes) to existing objects in the database when the code models defining those objects change (i.e. the core mixins upon which custom objects are built), or to upgrade custom objects on System B whose schema is of a different version from the custom objects on System A such that their schemas predictably match.

### External dependencies

*No response*

bctiemann:
### Closes: #386

Provides the groundwork and core data model for the portable COT schema format. Introduces stable field identifiers (`schema_id`), field lifecycle metadata (deprecation / scheduled removal), the `schema_document` store, and the canonical JSON Schema definition (`cot_schema_v1.json`) against which schema documents are validated.

## Model changes — `CustomObjectTypeField`

| Field | Type | Purpose |
|-------|------|---------|
| `schema_id` | `PositiveSmallIntegerField` (nullable) | Stable numeric identifier — never reused within a COT, even after deletion |
| `deprecated` | `BooleanField` | Mark a field as read-only / soft-retired in the UI |
| `deprecated_since` | `CharField` | Schema version in which the field was deprecated |
| `scheduled_removal` | `CharField` | Schema version in which the field is planned for removal |

## Model changes — `CustomObjectType`

| Field | Type | Purpose |
|-------|------|---------|
| `schema_document` | `JSONField` (nullable) | Last-applied schema document for this COT; source of truth for tombstoned field records |
| `next_schema_id` | `PositiveIntegerField` (default 0) | Monotonically increasing counter; auto-incremented on field creation via `QuerySet.update()` to avoid clearing the model cache |

`schema_id` is protected by a `UniqueConstraint(fields=["schema_id", "custom_object_type"], condition=Q(schema_id__isnull=False))`.

## `next_schema_id` design note

`schema_id` auto-assignment in `CustomObjectTypeField.save()` increments `next_schema_id` via `CustomObjectType.objects.filter(pk=...).update(next_schema_id=new_id)` rather than `cot.save()`. This bypasses the `post_save` signal that clears the in-process model cache, preventing a race where `contribute_to_class()` targets a freshly regenerated model class rather than the one `get_model()` already returned to the field-save caller.

## `cot_schema_v1.json`

Formal JSON Schema (draft 2020-12) for the portable document format. Key design points:

- Top-level document requires `schema_version: "1"` and a `types` array.
- Each field variant (`text_field`, `integer_field`, `object_field`, …) uses `unevaluatedProperties: false` to prevent extra keys from silently passing validation — mirroring the strictness already present on `removed_field`.
- `removed_field` tombstone entries carry `id`, `name`, `type`, and optional `removed_in` version.
- A `discriminator` hint (OpenAPI 3.x) is included for tooling; standard JSON Schema validators use the `oneOf` branches.

## `schema_format.py`

Shared constants consumed by the exporter, comparator, and executor:

- `SCHEMA_FORMAT_VERSION` — current format version string (`"1"`)
- `CHOICES_TO_SCHEMA_TYPE` / `SCHEMA_TYPE_TO_CHOICES` — bidirectional mapping between `CustomFieldTypeChoices` values and schema type names
- `FIELD_DEFAULTS` — canonical default values for all field attributes (omitted from schema documents when equal to default)
- `FIELD_TYPE_ATTRS` — which type-specific attributes are valid per field type

## Tests added

- `FieldDefaultsConsistencyTestCase` — cross-checks every key in `FIELD_DEFAULTS` against the corresponding model field default; catches drift between the constants and the model.
- `SchemaIdBackfillTestCase` — validates the `0005` migration logic: sequential assignment, monotonic counter update, gap-after-deletion prevention.
- Extended `SchemaFormatTestCase` — `unevaluatedProperties` rejects extra keys on all 9 field subtypes; `removed_field` accepts `removed_in`.

**Note:** `schema_document` is added in this PR but left `null`; the exporter (#388, PR #450) populates it.


arthanson (changes requested):
1. Model validators vs. identifier pattern mismatch

  CustomObjectType.name and CustomObjectTypeField.name both validate to ^[a-z0-9_]+$ (no double underscores). The JSON Schema identifier pattern (cot_schema_v1.json:675) is ^[a-z0-9]+(_[a-z0-9]+)*$, which additionally disallows leading and trailing underscores.

So a COT or field named _foo or bar_ is valid in the database but would produce an export that fails its own schema validator. The comment on identifier calls this out for field names, but cot_definition.name (schemas/cot_schema_v1.json:943) uses the same pattern — the gap applies to COT names too.

Seems like the validators should match.

2. schema_id - I'm a bit confused on this.  On one hand it is an auto-incrementing integer value and it is exported into the json schema.  It looks like this is to avoid conflicts between different people making updates to the schema but I think that would break down if they make two completely independent changes at the same time, the generated schema_id would be the same.

If it is not to avoid conflicts between different people, then it seems we could just use the fields pk. If it is to avoid conflicts it seems a UUID would be much better / safer for this.

bctiemann:
@arthanson Design discussion has settled that we are going with a single-source-of-truth model, i.e. many NetBox instances pull down an authoritative COT schema from a central repository, rather than trying to merge many peer schema changes. So for the `schema_id` question, we will stick with integers.

Note, this generally follows the pattern established by GRPC/protobuf, which also uses integers for its field IDs and places the onus on users to increment and curate them properly.

bctiemann:
@arthanson I addressed your concern about the regex validation on the `name` fields. Also added a documentation file.

The migrations will cause failures until we've merged https://github.com/netboxlabs/netbox-custom-objects/pull/472, so that should go in next.

arthanson (changes requested):
feature now has the Django 6 change from main so the test should pass but merging feature gets some merge conflicts that need to be cleaned up.   Also for #442

arthanson (approved):
The basics look good - but can't really test much.  One note to change before merge - the schema_id is not read-only in the serializer and it should be.  It was stated it can't change, but right now someone could post/put to it and change it.

Issue #389, opened 2026-02-03
Custom Object Type upgrade executor


bctiemann:
### Plugin Version

v0.4.6

### Proposed functionality

**Depends on** [NPL-714](https://linear.app/netboxlabs/issue/NPL-714/custom-object-type-state-comparator-functionality)

When a suitable list of schema change operations has been generated via the comparator developed in [NPL-714](https://linear.app/netboxlabs/issue/NPL-714/custom-object-type-state-comparator-functionality), we need a utility function that applies these changes in prescribed order via the schema editor:

```python
def apply_schema_upgrade(cot: CustomObjectType, operations: list[SchemaOperation]):
    with transaction.atomic():
        model = cot.get_model()
        with connection.schema_editor() as schema_editor:
            for op in operations:
                op.execute(schema_editor, model)
        cot.version = new_version
        cot.save()
```

### Use case

This functionality is needed for actually executing a planned and previewed set of schema migrations that alter the database on-disk to support a new version of a Custom Object Type.

### External dependencies

*No response*

Issue #390, opened 2026-02-03
Custom Object Type upgrade plan/execute API endpoints


bctiemann:
### Plugin Version

v0.4.6

### Proposed functionality

**Depends on** [NPL-716](https://linear.app/netboxlabs/issue/NPL-716/custom-object-type-upgrade-executor)

When we have a comparator class able to produce an ordered set of operations from two states of a given COT schema, we will need two API endpoints, one for a read-only preview and another for an "apply":

```
POST /api/plugins/custom-objects/custom-object-types/{id}/schema/preview/
  Body: { schema: <yaml/json> }
  Response: { operations: [...], warnings: [...], breaking_changes: [...] }

POST /api/plugins/custom-objects/custom-object-types/{id}/schema/apply/
  Body: { schema: <yaml/json>, confirmed: true }
  Response: { success: true, applied_operations: [...] }
```

### Use case

It will be necessary to apply Custom Object Type schema migrations during runtime, via an API endpoint, in an atomic way.

### External dependencies

*No response*

Issue #387, opened 2026-02-03
Custom Object Type state comparator functionality


bctiemann:
### Plugin Version

v0.4.6

### Proposed functionality

**Depends on** [NPL-713](https://linear.app/netboxlabs/issue/NPL-713/define-a-schema-format-for-portable-custom-object-type-definitions)

Implement a `COTStateDiff` class that:

* Loads current COT state from database
* Parses incoming schema definition
* Matches fields by stable ID (or name if no ID)
* Produces ordered list of operations

For precedent, Django's `MigrationAutodetector` compares two `ProjectState` objects (from_state, to_state) and generates Migration objects containing operations like:

* `AddField`
* `RemoveField`
* `AlterField`
* `RenameField`
* `CreateModel`
* `DeleteModel`

Custom Objects already performs these operations imperatively in CustomObjectTypeField.save():

```python
# models.py ~line 1504-1606
with connection.schema_editor() as schema_editor:
    if self._state.adding:
        schema_editor.add_field(model, model_field)
    else:
        schema_editor.alter_field(model, old_field, model_field)
```

To support versioned schemas, we need:

* A way to represent "desired state" (new schema version)
* A way to represent "current state" (existing COT + COTFs)
* A diff algorithm producing ordered operations
* Execution of operations in correct order (dependencies matter)
* Consider reversed operation (rollback/downgrade)

The below is a skeletal example of what the comparator class might look like:

```python
class COTStateDiff:
    def __init__(self, cot: CustomObjectType, new_schema: dict):
        self.cot = cot
        self.new_schema = new_schema
    
    def compute(self) -> list[SchemaOperation]:
        ops = []
        # Compare and generate AddField, RemoveField, AlterField, RenameField ops
        # Handle dependency ordering (e.g., add fields before constraints)
        return ops
```

**Problem:** Without stable field identifiers, renames are indistinguishable from remove+add.

Options:

* Numbered fields (protobuf-style) - Each COTF gets a stable numeric ID; renames are detected by ID match
* Explicit rename hints - Schema diff includes rename: old_name → new_name directives
* Name-based matching only - Accept that renames = remove + add (data loss)

### Use case

To support Custom Object Type Versioning, it is necessary for the plugin code to be able to parse a COT definition (as produced in [NPL-713](https://linear.app/netboxlabs/issue/NPL-713/define-a-schema-format-for-portable-custom-object-type-definitions)) and compare it against an existing COT that is present in the currently-running NetBox instance.

[...]

PR #453, opened 2026-04-08
Closes: #389 - Upgrade executor tool

bctiemann:
### Closes: #389

Implements the COT upgrade executor: a public Python API that takes a schema document (or pre-computed `COTDiff` list from the comparator, #387) and applies all changes to the live DB atomically.

## Public API — `executor.py`

```python
apply_document(schema_doc, *, allow_destructive=False) → list[COTDiff]
apply_diffs(diffs, type_defs_by_slug, *, allow_destructive=False) → None
```

`apply_document` is the primary entry point. `apply_diffs` is the lower-level primitive used by the schema apply API endpoint (#390, PR #454) when diffs are pre-computed for preview then applied separately.

## Two-phase execution

All DB writes are wrapped in a single `transaction.atomic()` block. Any exception causes a full rollback — including DDL (table creates/drops) since PostgreSQL supports transactional DDL.

**Phase 1 — COT records**
New `CustomObjectType` instances are created (which triggers `CustomObjectType.create_model()` and the backing `CREATE TABLE`). Existing COTs have their top-level attributes updated. New COTs are created in topologically-sorted order with respect to cross-COT `related_object_type` references, so a referenced COT's table always exists when the referencing COT's field is added in Phase 2.

**Phase 2 — Fields**
All `ADD`, `ALTER`, and `REMOVE` field operations are applied using the existing `CustomObjectTypeField.save()` / `delete()` mechanisms, which encapsulate all required DDL (`add_field`, `alter_field`, `remove_field`). By this point all COT tables are guaranteed to exist, so cross-COT FK resolution never fails due to ordering.

**Finalisation**
After Phase 2, for each COT: `schema_document` is updated (via `QuerySet.update()` to avoid clearing the model cache) and `next_schema_id` is synced to cover any `schema_id` values explicitly assigned during the apply.

## Dependency ordering

`_build_dep_order(diffs)` performs a DFS-based topological sort over the new-COT subset of the diff list. Only `custom-objects/<slug>` references within the document are treated as ordering constraints; references to existing COTs and built-in NetBox models are ignored. Self-references are not treated as cycles.

Raises `CircularDependencyError` if a cycle is detected before any DB writes.

## Exceptions

| Exception | Raised when |
|-----------|------------|
| `DestructiveChangesError` | `REMOVE` operations present and `allow_destructive=False` (checked before the transaction opens) |
| `CircularDependencyError` | Cross-COT object-field references form a cycle among new COTs |
| `UnknownChoiceSetError` | A `choice_set` name cannot be resolved to a `CustomFieldChoiceSet` |
| `UnknownObjectTypeError` | A `related_object_type` string cannot be resolved to an `ObjectType` |

## Tests — `test_executor.py` (50 tests, 11 classes)

| Class | Count | What it covers |
|-------|-------|---------------|
| `ExecutorBuildDepOrderTestCase` | 9 | `_build_dep_order` logic without DB: empty list, single diff, chain ordering, circular detection, self-reference, built-in ROT ignored |
| `ExecutorDestructiveGuardTestCase` | 4 | `allow_destructive` flag, `DestructiveChangesError.diffs` attribute, field survives rejected apply |
| `ExecutorNewCOTTestCase` | 8 | COT created in DB with correct attrs, fields created, `next_schema_id` synced, `schema_document` persisted |
| `ExecutorCOTAttrsTestCase` | 4 | Description, version, verbose_name updated; no-op leaves COT unchanged |
| `ExecutorFieldAddTestCase` | 7 | ADD with all field types; choice_set / ObjectType resolution; `UnknownChoiceSetError` / `UnknownObjectTypeError`; `next_schema_id` update |
| `ExecutorFieldAlterTestCase` | 5 | ALTER description, weight, required; rename; `validation_minimum` |
| `ExecutorFieldRemoveTestCase` | 3 | Field removed from DB; surviving fields unaffected; tombstone persisted in `schema_document` |
| `ExecutorSchemaDocumentTestCase` | 3 | `schema_document` set after no-op, contains correct slug and `schema_version` |
| `ExecutorDependencyTestCase` | 3 | Two new COTs with cross-reference succeed in both document orderings; object field resolves correctly |
| `ExecutorAtomicityTestCase` | 2 | Failed ADD rolls back new COT; failed mid-apply leaves field count unchanged |
| `ExecutorApplyDiffsTestCase` | 2 | `apply_diffs` low-level interface works; raises `DestructiveChangesError` correctly |


bctiemann:
@claude /review

arthanson (changes requested):
Blocked by #449

bctiemann:
No longer blocked, but can be merged into 387 branch or feature.

netbox_custom_objects/schema/executor.py
```diff
@@ -0,0 +1,530 @@
...
+    """
+    from netbox_custom_objects.schema.comparator import FieldOp  # noqa: PLC0415
+
+    for diff in ordered_diffs:
+        cot = cot_map[diff.slug]
+        for fc in diff.field_changes:
+            if fc.op is FieldOp.ADD:
+                _apply_field_add(cot, fc)
+            elif fc.op is FieldOp.REMOVE:
+                assert allow_destructive, (
```
arthanson: probably want that to raise an exception

arthanson (changes requested):
change the one place there is an assert, also.  ExecutorFieldAlterTestCase covers description, weight, rename, required, validation_minimum, choice_set, and related_object_type - but not a field type change (e.g., text → longtext or text → integer). Probably worth adding at least one test to confirm the existing CustomObjectTypeField.save() path handles it correctly through the executor.

netbox_custom_objects/schema/exporter.py
```diff
@@ -0,0 +1,224 @@
...
+    FIELD_DEFAULTS,
+    FIELD_TYPE_ATTRS,
+    SCHEMA_FORMAT_VERSION,
+)
+
+logger = logging.getLogger(__name__)
+
+# Matches the generated model name produced by CustomObjectType.get_table_model_name().
+# Capturing group 1 is the numeric COT id.
+_TABLE_MODEL_RE = re.compile(r'^table(\d+)model$', re.IGNORECASE)
```
arthanson: This is also defined in comparator.py, can we just define it once in constants.py?

netbox_custom_objects/schema/exporter.py
```diff
@@ -0,0 +1,224 @@
...
+    ``schema_document``.  Returns an empty list if the document is absent or
+    does not reference this COT.
+    """
+    if not cot.schema_document:
+        return []
+    # NOTE: matches by COT name. If the COT is renamed after tombstones
+    # are persisted, they will not be found. This will be addressed when
+    # #390 (apply) is implemented and tombstones are managed more explicitly.
+    for type_def in cot.schema_document.get("types", []):
+        if type_def.get("name") == cot.name:
```
arthanson: can't that just match on slug, that will bypass the rename issue?

arthanson (changes requested):

PR #454, opened 2026-04-08
Closes: #390 - Schema validation/apply endpoints

bctiemann:
### Closes: #390

Adds two new API endpoints for previewing and applying COT schema documents.

## New endpoints

| Method | URL | Description |
|--------|-----|-------------|
| `POST` | `/api/plugins/custom-objects/schema/preview/` | Diffs a schema document against live DB state; returns structured diffs **without any DB changes** |
| `POST` | `/api/plugins/custom-objects/schema/apply/` | Applies a schema document atomically; returns the applied diffs |

---

## `POST /schema/preview/`

**Request body** — a raw COT schema document:
```json
{
  "schema_version": "1",
  "types": [ ... ]
}
```

**Response `200`:**
```json
{
  "diffs": [
    {
      "slug": "my-cot",
      "name": "my_cot",
      "is_new": false,
      "has_changes": true,
      "has_destructive_changes": false,
      "cot_changes": { "description": ["old", "new"] },
      "field_changes": [
        {
          "op": "add",
          "schema_id": 5,
          "db_name": null,
          "schema_def": { "id": 5, "name": "new_field", "type": "text" }
        }
      ],
      "warnings": []
    }
  ]
}
```

Preview never modifies the database and never returns `409` — destructive changes are flagged via `has_destructive_changes: true` so the caller can decide whether to proceed.

---

## `POST /schema/apply/`

**Request body** — schema document wrapped with an optional `allow_destructive` flag:
```json
{
  "allow_destructive": false,
  "schema": {
    "schema_version": "1",
    "types": [ ... ]
  }
}
```

**Response `200`** (success):
```json
{
  "applied": true,
  "diffs": [ ... ]
}
```

**Response `409 Conflict`** — document contains `REMOVE` operations and `allow_destructive` was not set:
```json
{
  "error": "destructive_changes",
  "detail": "Schema contains destructive field removals for COT(s): my-cot. Pass allow_destructive=True to apply them.",
  "destructive_slugs": ["my-cot"]
}
```

**Response `400 Bad Request`** — circular COT dependency, unresolvable FK / choice-set reference, missing `schema` key, or invalid schema document structure:
```json
{
  "error": "circular_dependency" | "unresolvable_reference",
  "detail": "..."
}
```

---

## Implementation notes

- Both views are backed by the comparator (`diff_document`) and executor (`apply_document`) from issues #387 and #389.
- Input is validated against `cot_schema_v1.json` using `jsonschema.Draft202012Validator` before any DB access. Errors return `400` with `{"schema_errors": [{"path": [...], "message": "..."}]}`. Degrades gracefully if `jsonschema` is not installed.
- Both views use `_ignore_model_permissions = True`; authentication is still enforced.
- Apply is fully atomic — any error rolls back all changes including newly created COT tables.
- `schema_document` is persisted on each affected COT after a successful apply, so tombstones are available for future export/diff cycles.

## Tests — `tests/test_schema_api.py` (24 tests)

**`SchemaPreviewTestCase` (10):** 200 response, diff keys present, no DB mutation, field-add/alter detection, destructive preview returns 200 (not 409), auth enforcement, invalid schema returns 400.

**`SchemaApplyTestCase` (14):** new COT creation, field add, 409 guard without flag, 200 with `allow_destructive`, missing/non-dict `schema` key returns 400, unresolvable FK/choice-set returns 400, auth enforcement, `schema_document` persisted.


arthanson (changes requested):
Blocked by #449

bctiemann:
No longer blocked.

netbox_custom_objects/api/views.py
```diff
@@ -1,26 +1,94 @@
...
 
+logger = logging.getLogger(__name__)
+
+# ---------------------------------------------------------------------------
+# Schema document helpers
+# ---------------------------------------------------------------------------
+
+_SCHEMA_FILE = Path(__file__).parent.parent / "schema" / "cot_schema_v1.json"
+
+try:
```
arthanson: I'm leery of opening the file at import time, can this be a function / property that is cached?

arthanson (changes requested):
merge conflict that needs to be fixed. see comment re the opening of the file in views.py.  Also, The docstring says: `Raises ValidationError (DRF 400) if validation fails or if jsonschema is not installed.`

But the implementation does the opposite — if jsonschema is not installed, it silently returns early and lets the request through. The "or if not installed" clause should be removed.

arthanson (approved):

PR #467, opened 2026-04-14
Add thread-safe clear_cache suppressor and extract_cot_id_from_model_name

bctiemann:
## Summary

Separates two infrastructural utilities from PR #442 (polymorphic object fields) into a standalone, reviewable PR.

### `utilities.py`

**`extract_cot_id_from_model_name(model_name)`**
Replaces the fragile chained `.replace("table", "").replace("model", "")` pattern used throughout the codebase to extract a COT primary key from an internal model name (e.g. `table3model` → `"3"`). The old approach silently corrupts names that happen to contain the substrings `"table"` or `"model"`. The new function uses a strict regex (`^table(\d+)model$`) and returns `None` for non-matching names.

**Thread-safe `apps.clear_cache` suppressor**
Django's `apps.register_model()` calls `apps.clear_cache()`, which triggers our `get_models()` override → `get_model()` for every known COT → potential infinite recursion and spurious cache invalidations during dynamic model registration.

The old fix was a non-thread-safe module-level monkey-patch (`apps.clear_cache = lambda: None`) with a broken restore (`apps.clear_cache = apps.clear_cache`). The new approach:

- `install_clear_cache_suppressor()` — called once from `AppConfig.ready()`, installs a single wrapper on `apps.clear_cache` (idempotent).
- `_suppress_clear_cache()` — private context manager using a thread-local depth counter. Only the calling thread sees suppression; all other threads continue to receive real cache invalidation.
- `generate_model()` — updated to use `_suppress_clear_cache()` instead of the lambda patch.

### `__init__.py`

- Calls `install_clear_cache_suppressor()` at the top of `ready()`.
- Replaces the chained `.replace()` calls in `get_model()` with `extract_cot_id_from_model_name()`.

## Relationship to PR #442

PR #442 (polymorphic object fields) uses both of these utilities extensively. Once this PR merges, #442 should be rebased so it picks them up from `feature` rather than defining them itself. Note: `models.py` in #442 imports `suppress_clear_cache` directly — that import should be updated to use the private `_suppress_clear_cache` after rebase.

## Migrations

This PR also fixes the migration numbering conflict present in `feature`.


bctiemann:
@claude /review

bctiemann:
Addressed all surfaced issues. PR #442 will need to rebased on `feature` once this is merged.

arthanson (approved):
Just a question on if assert is correct of if we want to Raise the error on these checks.

bctiemann:
Changed the asserts to `ValueError`s.

arthanson (approved):

PR #469, opened 2026-04-14
Fixes #468: Suppress spurious warning from migration 0003 on fresh install

### LINKED ISSUE #468: Migrations for CO fail on clean database
### Plugin Version

0.4.10

### NetBox Version

v4.5.8

### Python Version

3.13

### Steps to Reproduce

1. Have a clean database (no migrations) with netbox and custom objects installed
2. run migrations

Note: I also tried it with the demo data for NetBox 4.5 loaded first and ran migrations after that and it produced the same error, so doesn't have to be a clean database - just no previous custom objects installed.

### Expected Behavior

Migrations should run without errors.

### Observed Behavior

Custom Objects migrations fail `Could not query CustomObjectType during migration (skipping FK constraint check): column `

```
  Applying ipam.0086_gfk_indexes... OK
  Applying netbox_custom_objects.0001_initial... OK
  Applying netbox_custom_objects.0002_customobjecttype_cache_timestamp... OK
  Applying netbox_custom_objects.0003_ensure_fk_constraints...Warning: Could not query CustomObjectType during migration (skipping FK constraint check): column netbox_custom_objects_customobjecttype.group_name does not exist
LINE 1: ... "netbox_custom_objects_customobjecttype"."slug", "netbox_cu...
                                                             ^
 OK
  Applying netbox_custom_objects.0004_customobjecttype_group_name... OK
  Applying sessions.0001_initial... OK
```

bctiemann:
## Summary

- Migration `0003_ensure_fk_constraints` imports the live `CustomObjectType` model to check FK constraints on existing rows. On a fresh install, `0003` runs before `0004`, so the live model already declares `group_name` but the column does not exist in the DB yet. The `SELECT … group_name …` query raised a `ProgrammingError`, which was caught by a savepoint/rollback — but the `print()` in the except handler fired anyway, producing a confusing warning on every clean install.
- Add a `SELECT EXISTS(…)` raw-SQL check at the top of the migration function. When the table is empty (fresh install), return immediately without touching the ORM. The savepoint fallback is retained for the genuine edge case: an upgrade where `0003` and `0004` are applied together against a non-empty table.

## Test plan

- [ ] Run `manage.py migrate` on a clean database — confirm no warning is printed for `0003`
- [ ] Run `manage.py migrate` upgrading from a database with existing `CustomObjectType` rows (pre-`0003`) — confirm FK constraint check still executes

### DIFF
--- a/netbox_custom_objects/migrations/0003_ensure_fk_constraints.py
+++ b/netbox_custom_objects/migrations/0003_ensure_fk_constraints.py
@@ -6,11 +6,27 @@ def ensure_existing_fk_constraints(apps, schema_editor):
     Go through all existing CustomObjectType models and ensure FK constraints
     are properly set for any OBJECT type fields.
     """
+    # Fast path: if there are no rows, there is nothing to do.  We check via
+    # raw SQL (SELECT id only) to avoid touching columns that may not exist in
+    # the DB schema yet when this migration runs — e.g. group_name is added by
+    # the *next* migration (0004), so the live model class already declares it,
+    # but the column is absent from the table until 0004 runs.  A plain
+    # CustomObjectType.objects.all() would generate SELECT … group_name … and
+    # raise a ProgrammingError even when the table is empty, which causes a
+    # confusing warning on every fresh install.
+    with schema_editor.connection.cursor() as cursor:
+        cursor.execute(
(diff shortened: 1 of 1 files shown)


arthanson:
@claude /review

arthanson (approved):

PR #474, opened 2026-04-17
Fixes #473: Skip absent fields when registering search index for stub models

### LINKED ISSUE #473: AttributeError on save when target COT has searchable object field and is cached as stub
## Plugin Version

0.4.10 (present since 0.4.5)

## NetBox Version

4.5.x (also affects 4.6.x)

## Python Version

3.12

## Steps to Reproduce

1. Create **COT B** with a text primary field `name` (searchable, `search_weight=500`) and an object field `device` pointing to `dcim.Device` (also `search_weight=500`).
2. Create **COT A** with a text primary field and an object field pointing to **COT B**.
3. Navigate to any view that triggers COT A model generation (e.g. the COT A list view) — this internally calls `B.get_model(skip_object_fields=True)`, which generates and caches a stub model for COT B that lacks the `device` field.
4. Save any **COT B** instance (create or edit).

## Expected Behavior

The instance saves successfully and the search index is updated for the fields present on COT B.

## Observed Behavior

An `AttributeError` is raised during the `post_save` search-indexing handler:

```
AttributeError: 'TableNModel' object has no attribute 'device'
```

**Root cause:** `register_custom_object_search_index()` (introduced in commit `e431cda`, Aug 2025) unconditionally includes every `CustomObjectTypeField` with `search_weight > 0` in the registered `SearchIndex`, regardless of whether those fields are present on the generated model class. When COT B is first generated as a stub (via `skip_object_fields=True`) to break the cross-COT FK recursion, object/multiobject fields are intentionally absent. The stub is cached, so subsequent calls to `B.get_model()` return the stub. When a COT B instance is saved, the search handler tries to read `instance.device` on a class that has no such attribute.

**Fix:** Guard each field in `register_custom_object_search_index()` with `model._meta.get_field(field.name)` and skip any field absent from the generated model.

bctiemann:
## Summary

- Guards `register_custom_object_search_index()` with `model._meta.get_field()` so fields absent from a stub model (generated via `skip_object_fields=True`) are excluded from the search index — fixes the `AttributeError` crash reported in #473.
- Populates `prechange_data` in the changelog for update and delete operations by snapshotting the instance before the serializer is constructed (mirrors NetBox's `get_object_with_snapshot()` pattern).
- Adds two regression tests covering the stub-index crash.

Closes #473

## Test plan

- [x] `test_save_does_not_crash_when_stub_cached_before_full_model` — creates a cross-COT FK so one COT's stub is cached; verifies `.save()` on the target COT instance does not raise.
- [x] `test_stub_model_search_index_excludes_absent_fields` — asserts the registered search index only contains fields present on the generated model.
- [x] `test_update_object` (inherited from NetBox's `APIViewTestCases`) — verifies that an `ObjectChange` record with correct `prechange_data` and `message` is created after a PATCH.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/netbox_custom_objects/api/views.py
+++ b/netbox_custom_objects/api/views.py
@@ -85,6 +85,18 @@ class CustomObjectViewSet(ModelViewSet):
             raise ValidationError(BRANCH_ACTIVE_ERROR_MESSAGE)
         return super().partial_update(request, *args, **kwargs)
 
+    def perform_update(self, serializer):
+        # Take a pre-change snapshot so prechange_data is populated in the changelog.
+        if hasattr(serializer.instance, 'snapshot'):
+            serializer.instance.snapshot()
+        super().perform_update(serializer)
+
+    def perform_destroy(self, instance):
+        # Take a pre-change snapshot so prechange_data is populated in the changelog.
+        if hasattr(instance, 'snapshot'):
+            instance.snapshot()
+        super().perform_destroy(instance)
+
 
 class CustomObjectTypeFieldViewSet(ModelViewSet):
     queryset = CustomObjectTypeField.objects.all()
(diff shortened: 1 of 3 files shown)


bctiemann:
@claude /review

arthanson (approved):

PR #475, opened 2026-04-20
Closes #425: Update and pin GitHub Actions for Node 24 compatibility

### LINKED ISSUE #425: Update and pin remaining GitHub Actions for Node.js 24 compatibility
### Proposed Changes

Review the repo-managed GitHub Actions workflows on `main` and update the remaining actions that still rely on deprecated Node.js 20 runtimes. As part of the same cleanup, pin the remaining version-tag references to full commit SHAs.

Affected workflows:
- `lint-tests.yaml`
- `release.yaml`

This primarily includes:
- `actions/checkout@v4`
- `actions/setup-python@v5`
- `actions/upload-artifact@v4`
- `actions/download-artifact@v4`

`lint-tests.yaml` already runs Ruff directly, so there is no separate Ruff action to update here. `release.yaml` already pins `pypa/gh-action-pypi-publish` to a commit SHA, so that step likely just needs validation as part of this work.

This change should be limited to the GitHub Actions used by the workflows themselves; it does not change the Python version used for testing or the release process.

### Justification

GitHub has deprecated Node 20 on Actions runners and will begin using Node 24 by default. Updating the remaining workflow dependencies now should help avoid warnings and future breakage while keeping `netbox-custom-objects` aligned with the rest of the NetBox ecosystem.

Pinning the remaining GitHub Actions to full commit SHAs also improves workflow security by reducing supply chain risk from tag retargeting and makes the referenced action versions easier to audit.

pheus:
### Fixes: #425

This PR updates the affected GitHub Actions for GitHub Actions' Node 20 deprecation and pins all referenced actions to full commit SHAs instead of version tags. This helps reduce supply chain risk from tag retargeting while preparing the workflows for the transition to Node 24.

The Node.js version used for CI testing is unchanged; this update only affects the GitHub Actions used by the workflows themselves.

Updated actions:
- `actions/checkout` to `v6.0.2`
- `actions/setup-python` to `v6.2.0`
- `actions/download-artifact` to `v8.0.1`
- `actions/upload-artifact` to `v7.0.1`
- `pypa/gh-action-pypi-publish` to `v1.14.0`

### DIFF
--- a/.github/workflows/claude.yaml
+++ b/.github/workflows/claude.yaml
@@ -26,7 +26,7 @@ jobs:
       actions: read # Required for Claude to read CI results on PRs
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           fetch-depth: 1
 
--- a/.github/workflows/lint-tests.yaml
+++ b/.github/workflows/lint-tests.yaml
@@ -19,9 +19,9 @@ jobs:
     timeout-minutes: 10
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - name: Setup Python
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
         with:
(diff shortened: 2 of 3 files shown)


pheus:
The failing test looks unrelated to this PR itself:
CI is running against a newer NetBox `main` change that now expects `prechange_data` to be recorded on updates, and our custom object update path does not populate that yet. It looks like #474 addresses this, so rebasing after that lands should resolve the failure.

bctiemann (approved):

Issue #404, opened 2026-02-11
Full branching support


bctiemann:
### Plugin Version

v0.4.6

### Proposed functionality

Full support for NetBox Branching involves significant work both on the Custom Objects and NetBox Branching sides.

Requirements:

**Schema replication for dynamic tables**

All DB schemas used for replicating branches would need to be populated with tables matching the Custom Objects `table7model` tables which contain the Custom Objects upon creation of a branch. Schemas within branches can diverge from the main branch schema at that point, but then when the branch is merged, the changes must be tracked and replayed against the main branch (see below).

**Branch-aware model generation**

When Custom Object Types are created or changed, the branch in which they were changed must be tracked within the COT objects themselves and available as part of the model (`_meta`?) for introspection.

**Merge strategy for schema changes**

When branches are merged, there must be a resolvable history for all COT changes within a branch that will then be applied to the main branch. Schema changes that have been applied to the branch would need to be replayed in order against the main branch.

If a COT has been changed in both a work branch and the main branch, merging would need to be performed in a resolvable way or not at all. Irreconcilable changes in different branches would need to flag the user that merging is not possible. Changes that are compatible, ideally, should be able to be performed upon merge (and the merge strategy should be able to determine whether the changes are mergeable).

### Use case

Users will need to be able to use NetBox Changes and Custom Objects at the same time, and will expect that Custom Objects should behave the same as core NetBox objects within branches.

Though full support for both plugins entails an expectation of full functionality, there should nonetheless be heightened expectation of potentially irreconcilable schema changes if Custom Object Types are changed in two parallel branches.

arthanson:
Fixed via https://github.com/netboxlabs/netbox-custom-objects/pull/476

PR #476, opened 2026-04-20
404 Add branching support

arthanson:
### Fixes: #404 

Adds native NetBox Branching support to the netbox-custom-objects plugin. Custom object types, fields, and instances can now be created, modified, and deleted inside a branch and then merged, synced, or reverted like any other model - the previous exempt_models workaround in PLUGINS_CONFIG and the branch-warning UI banner are no longer required.

**Note:** A separate PR will be added to add in a check framework to check for minimum versions of NetBox and branching if branching is installed.  This isn't added here as it would break any tests running until those versions are released.

## Key Changes

**Schema-aware DDL routing**
- New `_get_schema_connection()` helper routes schema-editor operations (add/remove/alter column, create/drop tables) to the active branch's PostgreSQL schema instead of always targeting main.

**Branch-aware model cache**
- `CustomObjectType._model_cache` is now keyed by `(cot_id, branch_id)` so main and each branch get their own generated class. Only main's class is registered in Django's app registry, so `content_type.model_class()` consistently resolves to a class matching main's schema even when called inside an active-branch context.

**Deserialization hooks for branch merge/revert**
- `deserialize_object` classmethods on `CustomObject`, `CustomObjectType`, and `CustomObjectTypeField` so `netbox-branching`'s `ObjectChange.apply()` runs the full save lifecycle (model generation, schema editor, signals) instead of bypassing it via `save_base(raw=True)`.

**Squash-merge ordering fix**
- ContextVar-keyed deferred CO field data: when a squash merge applies a CO CREATE before the field's CREATE, values are stashed and applied via raw UPDATE after each column is added.
- Pre/post merge/sync/revert signals reset the ContextVar at every operation boundary so leftover state from a failed prior op cannot leak forward.

**M2M change visibility**
- Custom `add`/`remove`/`clear`/`set` on the M2M descriptor now emit `m2m_changed`, and the through model is flagged `auto_created` so Django's serializer includes M2M values in `ObjectChange.postchange_data`. Without these, a merge or sync replays zero through-table rows for CustomObject M2M fields.

**Conflict resolution**
- `_schema_alter_field` detects divergent renames (same field PK renamed differently in main and branch), looks up the live column name in the target schema, and renames it to converge on the merge target.
- `_rename_objectchange_field_key` updates JSON keys in `core_objectchange` and `netbox_branching_changediff` so historical audit data stays consistent post-rename.

**Polymorphic fields**
- Polymorphic OBJECT/MULTIOBJECT now serialize a POLY_M2M_SIDECAR_KEY sidecar (field name + PK) so squash's dependency graph can order field CREATE before CO CREATE; their content_type references use db_constraint=False (plain PositiveBigIntegerField on the through) to avoid cross-connection django_content_type lock contention; PolymorphicManyToManyManager fires m2m_changed so change-logging captures values, and deserialize_object replays the sidecar via raw INSERT into a per-(cot, branch) through model.

**Branching extension hooks**
- `PluginConfig.serializer_resolver` resolves dynamically-generated CO serializers (which have no importable path under the conventional `{app_label}.api.serializers.{Model}Serializer`).
-  branching.objectchange_field_migrator (registered via register_objectchange_field_migrator) translates stale field-name keys in stored ObjectChange data so replays survive field renames. Shares CustomObject.resolve_field_aliases with deserialize_object for CREATE replay.
- `branching.supports_branching_resolver` (registered via `register_branching_resolver`) marks dynamically-generated CO M2M through-models as branchable so their queries are routed to the active branch's connection.
- `branching.add_custom_object_dependencies` (connected to `squash_dependency_graph_built` signal) extends squash's dependency graph with CO-specific edges (local M2M target lists, CustomObjectTypeField PKs from _field_objects, and polymorphic POLY_M2M_SIDECAR_KEY entries) so squash applies field CREATEs before CO CREATEs and orders self-referential M2M targets correctly.

## Other
- Datetime filters now make naive values timezone-aware before comparison.
- `cache_timestamp` is excluded from ObjectChange serialization.
- `snapshot()` is called before saves that bump `cache_timestamp` so change logging records a correct pre-state.
- `pre_delete` handler reconnect moved into a try/finally flow so it's always restored.
- The previous `is_in_branch` API/UI gates and branch-bypass warning banners are removed.

## Tests

New `tests/test_branching.py` (~1,500 lines) covering:
- Simple and all-field-types merge + revert
- Object create/modify/delete in branch
- Field rename, unique toggle, non-schema attribute changes
- Sync (main→branch) with/without concurrent edits
- Concurrent field renames (conflict resolution)
- Sequential A→B→C rename chains with CO updates at each step

Merge/revert and rename-chain tests run under both iterative and squash strategies via parameterized base classes (`BaseBranchingTests`, `BaseConcurrentEditMergeTests`, `SequentialRenameTestCase` / `SequentialRenameSquashTestCase`).

Test infrastructure additions in `tests/base.py`: stale-model purge, ContentType recreation, and ObjectChange cleanup between `TransactionTestCase` runs.


arthanson:
@claude /review

arthanson:
All claude concerns addressed.

bctiemann (changes requested):
Overall this PR is excellent work — the branching architecture is sound and the test coverage is thorough. One bug found during the `_get_schema_connection()` coverage audit that needs a fix before merge.

## Bug: `create_polymorphic_m2m_table` existence check probes the wrong schema

**File:** `netbox_custom_objects/field_types.py`

```python
def create_polymorphic_m2m_table(self, field_instance, model, schema_editor):
    ...
    with connection.cursor() as cursor:              # ← global connection (main schema)
        existing_tables = connection.introspection.table_names(cursor)
        if table_name not in existing_tables:
            schema_editor.create_model(through)      # ← branch-aware schema_editor
```

The caller (`CustomObjectTypeField.save()`) correctly opens `schema_editor` via `schema_conn.schema_editor()` where `schema_conn = _get_schema_connection()`, so `create_model` targets the branch schema. But the idempotency guard uses the module-level `connection` import, which always probes the main schema. This diverges in two cases:

1. **Through table previously merged into main, not yet present in the branch** — check finds it in main → skips `create_model` → through table is missing in the branch schema.
2. **Through table exists in the branch but not in main** (e.g. mid-merge replay on a branch-only COT) — check finds it absent in main → spurious `CREATE TABLE` against a table that already exists → DB error.

Every other DDL path in the PR is correctly branch-aware (the non-polymorphic `create_m2m_table` re-binds its local `connection` to the passed `schema_conn`; all `_schema_add/remove/alter_field` helpers take `schema_conn` explicitly). This one was missed because it receives the schema_editor rather than a raw connection.

**Fix:** derive the connection from `schema_editor.connection` (`BaseDatabaseSchemaEditor` always sets `self.connection` in `__init__`):

```python
conn = schema_editor.connection
with conn.cursor() as cursor:
    existing_tables = conn.introspection.table_names(cursor)
    if table_name not in existing_tables:
        schema_editor.create_model(through)
```

This is the same pattern used by `remove_polymorphic_object_columns` and the non-polymorphic `create_m2m_table` — it just needs to be applied here too.

Issue #391, opened 2026-02-03
Automatically handle Custom Object Type schema upgrades created through changes to core mixins


bctiemann:
### Plugin Version

v0.4.6

### Proposed functionality

**Important consideration:** If changes are made to core mixins (the base classes that Custom Object models inherit from), there is currently **no mechanism to generate migrations** for the dynamically created Custom Object tables.

**The Problem:**

Custom Object models inherit from mixins like `NetBoxModel`, `ChangeLoggingMixin`, etc. When these core mixins are updated (e.g., new field added, field type changed), the changes need to propagate to:

1. The dynamic Custom Object tables in the database
2. Any related indexes, constraints, or triggers

Since Custom Objects don't have traditional migration files (they use runtime schema operations), there's no standard way to:

* Detect that a mixin has changed
* Generate the appropriate `ALTER TABLE` statements
* Apply these changes across all existing Custom Object tables

**Impact:**

| **Scenario** | **Current Behavior** | **Desired Behavior** |
| -- | -- | -- |
| New field added to mixin | Not applied to existing COT tables | Auto-detect and apply |
| Field type changed in mixin | Mismatch between model and table | Migration generated and applied |
| Index added to mixin | Missing on COT tables | Created on all COT tables |
| NetBox upgrade with mixin changes | Potential runtime errors | Seamless upgrade |

**Possible Solutions:**

1. **Mixin version tracking** - Track which mixin version each COT was created with; compare on startup
2. **Schema introspection** - Compare actual table schema to expected model schema; generate diffs
3. **Explicit upgrade command** - `manage.py upgrade_custom_objects` that applies pending mixin changes
4. **Hook into NetBox migrations** - Detect relevant core migrations and apply to COT tables

### Use case

If changes are made to core mixins (the base classes that Custom Object models inherit from), there is currently no mechanism to generate migrations for the dynamically created Custom Object tables.

### External dependencies

*No response*

PR #480, opened 2026-04-23
Closes: #391 (Phase 1) - Snapshot mixin base columns into schema_document

bctiemann:
## Summary

- Adds `CustomObjectType._collect_base_columns()` -- static method that returns the concrete DB columns contributed by the `CustomObject` mixin hierarchy (e.g. `id`, `created`, `last_updated`), excluding user-defined field names.
- Adds `CustomObjectType._store_base_column_snapshot()` -- writes those columns to `schema_document["base_columns"]` immediately after a COT table is created, preserving any other existing keys in `schema_document` (e.g. the schema-exporter output from #386).
- Hooks `_store_base_column_snapshot()` into `create_model()` so every new COT records its base column state at birth.
- Adds migration `0009_backfill_base_columns` that uses DB introspection to populate `schema_document["base_columns"]` for all existing COTs, subtracting user-defined field names to isolate the base columns.

## Context

This is Phase 1 of the fix for #391 ("Automatically handle Custom Object Type schema upgrades created through changes to core mixins"). The snapshot recorded here will be consumed by the Phase 2 `post_migrate` auto-heal handler, which will diff the stored snapshot against the current `CustomObject` model definition and apply any new columns to existing COT tables when NetBox is upgraded.

See the strategy discussion on #391 for the full design.

## Migration note

`392-validate-semver-fields` (PR #455) has a `0009_alter_customobjecttype_version` migration that will conflict with our `0009_backfill_base_columns`. That PR will need its migration renumbered to `0010` before it merges into the updated `386-schema-export`.

## Test plan

- [ ] `CollectBaseColumnsTestCase` -- `_collect_base_columns()` returns `id`/`created`/`last_updated`, excludes user-defined field names, and produces entries with `name`/`field_class`/`null` keys
- [ ] `StoreBaseColumnSnapshotTestCase` -- snapshot is written to `schema_document["base_columns"]`, existing document keys are preserved, in-memory attribute is updated
- [ ] `CreateModelSnapshotTestCase` -- a freshly created COT has `base_columns` in `schema_document` with the expected mandatory columns
- [ ] `BackfillBaseColumnsTestCase` -- migration function populates existing COTs, excludes user fields, is idempotent, skips already-backfilled rows, handles multiple COTs, and produces entries with required keys

Generated with [Claude Code](https://claude.com/claude-code)

PR #481, opened 2026-04-23
Closes: #391 (Phase 2) - post_migrate auto-heal for mixin column drift

bctiemann:
## Summary

Implements automatic detection and repair of mixin column drift for Custom Object Type tables (Phase 2 of #391). When NetBox is upgraded and a mixin (e.g. `ChangeLoggingMixin`) gains a new concrete column, existing COT tables would previously be missing that column silently. This PR closes that gap.

**Depends on PR #480** (Phase 1 -- base column snapshot).

### New module: `mixin_migration.py`

- `_expected_base_fields(cot)` -- resolves the full set of base columns from the live `CustomObject` model, excluding user-defined fields
- `_can_auto_add(field)` -- determines whether a missing column can be safely added (`ADD COLUMN`) without breaking existing rows (nullable or has a default)
- `heal_cot(cot, verbosity, dry_run)` -- detects drift for a single COT, adds safe columns via `schema_editor.add_field()`, warns on unsafe ones (NOT NULL/no default, type changes, removals), never auto-drops, refreshes `schema_document["base_columns"]` after successful additions
- `heal_all_cots(verbosity, dry_run)` -- iterates all COTs, returns `{total, healed, warnings}` summary

### Signal wiring (`__init__.py`)

`_heal_mixin_columns()` is connected to `post_migrate` in `ready()`. It fires automatically after every `manage.py migrate` run. A module-level `_heal_ran` flag ensures the work happens only once per process even though `post_migrate` fires once per installed app. Excluded during `makemigrations` and `collectstatic`.

### Management command: `upgrade_custom_objects`

Explicit escape hatch for inspection and manual repair:
```
manage.py upgrade_custom_objects               # check and fix all COTs
manage.py upgrade_custom_objects --dry-run     # preview without DB changes
manage.py upgrade_custom_objects --cot mytype  # single COT by name or ID
```

## Safety rules

| Scenario | Action |
|---|---|
| New nullable column | Auto-ADD |
| New column with default | Auto-ADD |
| New NOT NULL column, no default | Warn only |
| Column type changed | Warn only |
| Column removed from mixin | Warn only, never DROP |

## Test plan

- [ ] `CanAutoAddTestCase` -- nullable, defaulted, both, neither
- [ ] `ExpectedBaseFieldsTestCase` -- returns id/created/last_updated, excludes user fields, returns real Django Field instances
- [ ] `HealCotTestCase` -- no-drift returns empty; nullable column is added to DB; non-nullable warns; snapshot updated after addition; removed column warns but not dropped; dry_run reports without touching DB; dry_run does not update snapshot
- [ ] `HealAllCotsTestCase` -- total matches COT count, healed=0 when no drift, correct summary keys
- [ ] `UpgradeCustomObjectsCommandTestCase` -- runs without error, --dry-run flag, --cot by name, --cot by ID, unknown COT raises CommandError


bctiemann:
@claude /review

bctiemann:
All review items addressed.

arthanson (changes requested):
need to merge feature and fix merge conflicts

PR #482, opened 2026-04-24
Closes: #26 - Add a combined "Custom Objects" related tab

### LINKED ISSUE #26: Auto generation of tabs in other objects that lists related Custom Objects
### Plugin Version

commit 722bbd624c8af515dc584b719e2864cbfcc14dd0 (HEAD -> public-preview, origin/public-preview) Author: btiemann <btiemann@netboxlabs.com> Date:   Thu Jun 26 15:31:01 2025 -0400

### Proposed functionality

**Custom Object Type** should have a property that will specify objects that have been added as fields and the existence of data added to the custom objects to automatically add the custom objects to a tab on the related object type which lists all the custom objects that are related to that related object.

**Context/Background**

Many objects will display the details of other objects as a tab, some native examples of this are Devices with interface or contact tab’s which list the interfaces/ip addresses or contacts assigned to that device.

**Impact**

Users will have more control over how their custom objects are interacted with, allowing them to create intuitive workflows, saving users time.

### Use case

* A concrete example would be a custom object for firewall rules. The Custom Object Type Firewall Rules would have the following fields 

  * Device (they system the rules are added to)
  * Source IP
  * Destination IP
  * Port
  * Protocol

* The Custom Object Type would have a property Create Related Tab on Objects which would contain  IP Address and Device.

* A tab would be automatically generated for both the IP Address and Device objects which would list each of the individual Custom Objects entered that are linked in the that object. 

Image

### External dependencies

_No response_

Kani999:
## What

Adds a single combined **Custom Objects** tab to every object detail page that is referenced by a Custom Object. The tab lists every Custom Object linking to the object being viewed — across all Custom Object Types and all Object / Multi-object fields (polymorphic and non-polymorphic) — with a count badge, live quicksearch, Type/Tag filters, sortable columns, HTMX pagination, and per-row actions. It works on built-in NetBox models and on custom-object (CO→CO) detail pages alike.

It **supersedes** the old left-column "Custom Objects linking to this object" panel, which is removed.

## Why

Two motivations:

1. **One discoverable surface.** Previously a linking relationship surfaced as a static left-column panel with no search, filtering, sorting, or actions. The tab matches the native NetBox child-object list (Device → Interfaces) look and behaviour.
2. **Permission correctness.** The old panel rendered linked rows ignoring per-Custom-Object-Type view permissions — a pre-existing info leak. The tab filters by the viewing user's permissions throughout: the linked rows, the Multi-object **Value** column (polymorphic targets included), and the count badge — and the tab hides entirely when the user may view none of the linked objects.

## How it works (no coordination machinery)

- `registry.register_tabs()` registers the tab view on **every public model** (`ObjectType.objects.public()` — exactly the set a COT Object/Multi-object field can target) at plugin startup, and injects one COT-agnostic URL for custom-object host pages. Display is gated purely by a live per-request, permission-aware badge + `hide_if_empty`, so a newly-referenced model lights up on the **next request with no restart** — no middleware, signals, or shared cache.
- `views/combined.py` centralises the four reference shapes in `reference_q()` (the single source of truth), with an `.exists()` fast path that keeps the badge cheap on the common detail pages that reference nothing.
- The badge reads the viewing user from NetBox's `current_request` and restricts the per-field counts, so it reflects the rows the user can actually open (not a higher, unfiltered total).
- Styling mirrors `ObjectChildrenView`: controls row above the table, `object-list` table with `th.orderable` headers (`?sort=` / `-sort`), and a compact, keyboard-accessible per-row dropdown (a real `<button>` toggle — pencil Edit + Changelog + Delete), permission-gated.

## What is intentionally **not** here

- **Native multi-select bulk actions.** Rows are heterogeneous Custom Object Type models (and one object can repeat under several linking fields), so NetBox's single-endpoint bulk edit/delete cannot apply. Per-row actions only.
- **SQL-level filtering/sorting/pagination.** The same heterogeneity means the rows cannot be combined into a single SQL query — the tab filters, sorts, and paginates in memory, so rendering cost grows with how heavily an object is referenced. Documented in the README's Related Objects Tab section.
- Configure-Table omits the saved-table-config "Save" split (that needs a single ObjectType) — matching native, which only shows it when `config_params` exist.

## Tests

`netbox_custom_objects/tests/test_related_tabs.py` (31 tests): `reference_q` shapes + empty-Q-means-skip (including the polymorphic multi-object through-table path), per-row **and** badge permission filtering, `_public_host_model_classes` inclusion/exclusion, idempotent registration, and unit tests for the value / search / sort display helpers.

## Scope note

This branch is a clean recomposition of the earlier typed + combined + hot-reload exploration into the **combined tab only**. The typed-per-COT tabs and the hot-reload machinery are deliberately dropped as out of scope for this PR; that exploration is preserved on `feature/related-object-tabs-v2`.


damsitt:
Instead of a static list in PLUGINS_CONFIG, a per-COT "Show as dedicated tab" checkbox in the admin UI would be cleaner. High-priority COTs get their own tab; everything else falls back to a consolidated "Custom Objects" tab.

Kani999:
@damsitt thanks for the suggestion — done. Replaced `typed_tab_slugs` with a per-COT **Show as dedicated tab** BooleanField on `CustomObjectType`. The toggle is reachable from the COT edit form, bulk-edit, CSV import, the REST API, and shows up as a column on the list view. When ticked, the COT renders its own typed tab on related-object detail pages; when unticked, its objects fall back into the consolidated "Custom Objects" tab. Restart-on-change for now (registration runs once in `ready()`), with a TODO in `tab_views.py` for live re-registration as a follow-up.

12 commits, `977feee` … `fd4155b`. The `typed_tab_slugs` plugin-config setting is gone (no deprecation since this PR isn't merged yet — clean break).

bctiemann:
@Kani999 We're working on finalizing the v0.5.0 release of netbox-custom-objects this week. It already has a huge number of major features on-train, but it would be nice to get this one in there too. However, as it was not a stakeholder promise it isn't the end of the world if it has to be deferred to a v0.6.0.

What is your feeling on the readiness? Is this week realistic?

Kani999:
## Smoke Test Results

  | # | Scenario | Where to look | Expected | ✅/❌ | Notes |
  |---|----------|---------------|----------|-------|-------|
  | 1 | **Dedicated tab – Device** (`smoke-dedicated-device`, dedicated=on) | Any of the first 3 devices | Own "Smoke Dedicated Device" tab visible | ✅ | |
  | 2 | **Combined tab – Device** (`smoke-combined-device`, dedicated=off) | Same devices as above | Object appears only in "Custom Objects" combined tab, no dedicated tab | ✅ | |
  | 3 | **Dedicated tab – Site** (`smoke-dedicated-site`, dedicated=on) | Any of the first 3 sites | Own "Smoke Dedicated Site" tab visible | ✅ | |
  | 4 | **Combined tab – Site** (`smoke-combined-site`, dedicated=off) | Same sites as above | Object appears only in "Custom Objects" combined tab, no dedicated tab | ✅ | |
  | 5 | **Dual-model COT** (`smoke-dual-model`, dedicated=on, references Device + Site) | Any linked device AND any linked site | Dedicated tab appears on **both** the Device and the Site detail pages | ✅ | |
  | 6 | **All field types** (`smoke-all-fields`, dedicated=on) | Any of the first 3 devices | Tab shows all 13 field types rendered correctly — no stray em-dashes where values exist | ✅ | |
  | 7 | **Empty dedicated tab** (`smoke-empty-dedicated`, dedicated=on, 0 objects) | Any device detail page | Tab is not visible | ✅ | |
  | 8 | **Live toggle ON** (`smoke-toggle`, start dedicated=off) | Edit the COT → toggle "Show as dedicated tab" ON → open a linked device | Dedicated tab appears **without restarting** | ✅ | |
  | 9 | **Live toggle OFF** (continue from #8) | Toggle back OFF → reload the device | Dedicated tab disappears without restart | ✅ | |
  | 10 | **Slug rename duplicate** (rename any dedicated COT's slug) | Device/Site detail page after save | ⚠️  Both old and new tabs appear simultaneously (known limitation — clears on restart) | ✅ | |
  | 11 | **Permission gating** | Log in as user without view permission on a smoke COT | Tab is not visible for that COT | ❓ | [comment](https://github.com/netboxlabs/netbox-custom-objects/pull/482#issuecomment-4385769479) |
  | 12 | **Create new COT at runtime** | Create a brand-new COT with dedicated=on via UI | Tab appears on referenced object pages without restarting | ✅ | |

## Import Test Data
[smoke_data.json](https://github.com/user-attachments/files/27428572/smoke_data.json)
[import_smoke_data.py](https://github.com/user-attachments/files/27428573/import_smoke_data.py)

Kani999:
Hi @mcolemann — from my side it feels ready. I've just run through the full smoke test matrix above and all scenarios pass except permission gating (which I haven't had a chance to verify yet, but the underlying logic is standard NetBox — it should just work).

One **known limitation** worth calling out before merge: if you rename a COT's slug while the server is running, the old dedicated tab continues to appear alongside the new one until the process is restarted. The tab registry picks up the new slug immediately (live toggle works correctly), but the old URL entry stays in the router until the next startup. It's documented in the smoke test table (row 10) and in a `Known Limitations` comment in `tab_views.py`. It's an edge case — renaming slugs isn't a common operation — but I wanted to flag it explicitly.

If you or the team can spin up the test data (script + JSON are attached above) and walk through a few scenarios, that would be the fastest path to confirming it's v0.5.0-ready. Happy to address any issues that come up during your review.

Kani999:
### Permission gating — observed behaviour (smoke test row 11)

Tested with a user who has **view: Device** only (no permissions on any custom object type).

**Base panel** (the "Custom Objects" linked-objects panel in the left-side panels area):
Still visible and shows all 5 linked objects. This panel is rendered by the base plugin template extension and doesn't gate on per-COT permissions — it lists everything. That's probably worth a follow-up: ideally it would hide rows the user can't access, but it's existing behaviour, not a regression from this PR.

**"Custom Objects" combined tab** (the tab injected at the top of the device detail page):
The tab badge shows **1** (correct — there is 1 non-dedicated object linked to this device), but after clicking, **0 rows** are rendered. The tab body does respect permissions correctly; it's only the badge count that leaks the existence of an object the user can't see.

**Dedicated tabs:**
Not visible at all for this user — correct.

**Summary:**
- Badge count on the combined tab is a minor info leak (count visible, data not)
- Base panel ignores per-COT permissions entirely (pre-existing)
- Dedicated tabs are correctly hidden
- Tab body correctly renders 0 rows when user lacks permission

Neither issue is introduced by this PR, but worth noting before v0.5.0 ships.

bctiemann:
@Kani999 Thanks for pushing this forward. I think, in the interest of avoiding too much churn and destabilization, I'd like to defer this to a v0.6.0 release. (But note that doesn't mean it will be a long time before that release; it's just the next one to be cut from the `feature` branch. Which, note, is what this PR will need to target, not `main`.)

The main thing I'm worried about is [polymorphic object/multiobject fields](https://github.com/netboxlabs/netbox-custom-objects/pull/442), which are just about to land in `feature` and go out with v0.5.0. There is a lot of movement in the code there that will very likely impact this PR/feature, and I'd like polymorphism to settle first to give us a chance to ensure related object tabs are baked well and support polymorphism properly. I don't want to rush this.

Kani999:
Understood, thanks for the context — agreed it's better to land related-object tabs on top of finalized polymorphic field support than to chase a moving target.

I've already retargeted this PR from `main` to `feature` and rebased onto the current `feature` head, so it's MERGEABLE now and the diff reflects only the tab work. I'll keep an eye on #442 and rebase again once polymorphism lands, so this is ready to revisit whenever v0.6.0 is being cut.

Kani999:
Heads up — this is **not ready to merge** yet.

I need to:

- Rework the code to align with the **polymorphic** feature
- Re-run the test suite against the updated implementation
- Add/adjust tests as needed

I'll push the revised commits and updated test results once that's done.

jeremystretch:
@Kani999 I've converted this PR to a draft per your note above. Just mark it as "ready for review" when the time comes. Thanks!

Kani999:
This branch has been completely reworked on top of upstream `main` (v0.5.1 + #524) with full polymorphic Object/Multi-object field support. The PR description above has been updated to reflect the new implementation — please review against the current description rather than the original one.

Kani999:
@bctiemann 

### Quick thought / question on `nbco:tab_registry_version` (Redis-based hot-reload)

Before this lands I'd like to get your gut feel on the Redis-counter hot-reload mechanism, because I'm honestly a bit uncomfortable with the complexity it adds and I'd rather hear "yeah, that's fine" or "no, please simplify" from you now than after merge.

**Why this mechanism exists at all.** Unlike core NetBox tabs, the tabs this PR introduces depend on database state (`CustomObjectType` rows + `show_dedicated_tab`), which users can change at runtime. NetBox builds the tab registry and URL conf once in `ready()`, so without some kind of cross-process signal, a `PATCH show_dedicated_tab=true` on one worker would not become visible on other workers until a NetBox restart. That's a pretty rough UX for a feature whose whole point is runtime-defined object types.

**What I did.** A small monotonic counter in Redis (`nbco:tab_registry_version`) + a thin middleware that GETs it once per request and rebuilds the local registry if it's behind. Cost on the steady-state hot path is ~1 ms (one Redis GET).

**What bugs me.** It's still a piece of distributed state that has to stay in sync with the DB, every request pays a Redis GET, and the `ViewTab.visible()` + startup seeding are defence-in-depth specifically because there are more moving parts than I'd like.

**What I tried.** I built a side branch that strips the Redis machinery and refreshes the registry only in the process that handled the mutation — diff [here](https://github.com/netboxlabs/netbox-custom-objects/compare/main...Kani999:netbox-custom-objects:experiment/local-only-tab-refresh), single commit [`3366144`](https://github.com/netboxlabs/netbox-custom-objects/commit/3366144efd3aec47330e9b32d50dc57ec1d0c892). I deployed it on a server running **granian with multiple workers** and confirmed the failure mode empirically: after toggling `show_dedicated_tab=true`, reloading the related detail page shows the tab on some reloads and not on others — the tab is visible only when the request happens to hit the worker that handled the PATCH, and invisible on every other worker. So without the Redis counter + middleware I can't get hot-reload to propagate to all workers; a NetBox restart is the only way to converge them.

**What I'd love your input on.**
1. Is there an existing primitive in NetBox/Django land for "tell all workers to invalidate something" that I missed? (I looked but didn't find one.)
2. If not, are you OK with the Redis-counter approach as it stands, or would you rather accept the restart-required tradeoff and drop the whole mechanism?
3. Or — third option — is there a different angle entirely (lazy registry rebuild on miss, periodic poll, a TTL on the registry, …) that you'd prefer?

Happy to go any direction. I just don't want to silently merge in a non-trivial coordination primitive without you having had a chance to weigh in.

bctiemann:
@Kani999 My first reaction is that dealing with cache invalidation across multiple workers has been one of the biggest persistent issues in Custom Objects ever since the cache (whose purpose is to prevent infinite recursion during `get_model` invocations of COTs with object/multiobject fields) was introduced. One of the thing we tried hard to avoid having to introduce was a shared cache backend, i.e. in Redis, and thus far we've been successful at that.

What we've had to do in recent changes (in the 0.5.0 cycle) is introduce caching techniques with versioning/timestamps that force different workers to invalidate their caches. I don't have the details at top of mind, but if you're using Claude to introspect the cache machinery, I'd prompt it with the challenge to use a similar technique to solve this problem — see whether the dynamic model cache can be manipulated when tab display needs to be updated, using the same existing techniques for ensuring cache freshness across workers that are used when COTFs are updated.

Kani999:
@bctiemann — done, and thanks, that was the right instinct.

I dropped the Redis counter entirely and rebuilt hot-reload on the **`cache_timestamp` versioning the dynamic-model cache already uses**. Each worker keeps a `(MAX(cache_timestamp), COUNT(*))` snapshot over `CustomObjectType`; a thin middleware compares it per request and re-registers tabs only when it drifts. `MAX` catches creates/updates (every COT + COTField save bumps `cache_timestamp`, including the `m2m_changed` path for polymorphic targets); `COUNT` catches deletes. No shared cache backend — the same DB column that gates the model cache gates the tab registry, so the invariant lives in one place.

Kani999:
Stepping back, though: even on `cache_timestamp`, the hot-reload subsystem is **heavy** — middleware on every request, signal handlers, registry purge/re-register, and the part I trust least, `_inject_host_typed_tab_urls()` rewriting host apps' frozen `urlpatterns` in place. It works, but it's a lot of moving parts for a tabs feature, and this PR's size reflects that.

I think we can **delete the entire mechanism** by splitting the feature in two:

**Step 1 — combined "Custom Objects" tab only.** One tab per host model; badge/contents/filters all read from the DB per request. No middleware / signals / URL-injection / cache token. Branch: [`feature...feature/related-object-tabs-combined`](https://github.com/netboxlabs/netbox-custom-objects/compare/feature...Kani999:netbox-custom-objects:feature/related-object-tabs-combined) — a clean ~+1.1k-line pure addition, zero model/migration changes. It's purely additive: it does **not** remove or change the existing "Custom Objects linking to this object" panel (`template_content.py`) — both surface the same relationships, so if we go this way we should decide whether the tab supersedes that panel or they coexist.

Liveness, by host kind:

- **References between custom object types (CO→CO) are always live — no restart.** Custom-object detail pages are rendered by the plugin's own template, so the tab nav-link is computed live from the DB per render, and its URL is a single COT-agnostic route (`…/<cot-slug>/<pk>/custom-objects/`, slug as a path param) injected once at startup that reverses for *any* slug — including COTs created later. (Smoke-tested: a brand-new COT referencing another shows the tab immediately, badge updates live.)
- **Built-in NetBox hosts (Device, Site, …):** live for everything *except* the first-ever reference to a NetBox model type nothing referenced at startup — that model's per-model tab URL is frozen into the URLconf at boot, so it needs a restart to appear. (No `manage.py` command avoids this: a separate process can't mutate live workers' URLconf; only shared polling — what we're deleting — or a restart crosses the process boundary.)
- **Variant B** removes even that built-in caveat: register the tab on *all* models at startup and gate display by the live DB badge → fully live everywhere, still zero machinery, at the cost of a cheap "is this referenced?" check on every detail page.

Everyday usage (new COTs against existing targets, new objects, edits) is live under every variant.

**Step 2 — per-COT typed tabs (separate top-level tabs), as a focused follow-up.** These give each opted-in COT its own dedicated tab on the host page (native columns, per-field filters, bulk actions). They're the part that genuinely *needs* runtime URL registration — each typed tab requires its own reversible URL name, which is exactly what drove the hot-reload / URL-injection machinery in this PR. I'd land them on top of the combined tab once it's in, and decide then between liveness options for them specifically.

Honestly I lean toward shipping **Step 1 (combined tab) first** — it's the least noisy, lands the core value, and lets typed tabs come as a focused follow-up rather than carrying the full machinery in one merge. What's your call on the built-in liveness (accept the restart caveat vs. Variant B), and on combined-first vs keeping it all in one PR?

bctiemann:
I do like the sound of Step 1 (with the Variant B -- the tradeoff of a "is this referenced?" check (if the plugin is even present?) versus having to worry about the "first time referenced after startup" business seems fine to me).

Step 2 may or may not even be necessary or advisable as I understand it -- there can be an arbitrary number of different COTs linking to a core object, which means an arbitrary number of tabs, right? I had thought the shape this was going to take was just a single "Custom Objects" tab that aggregated everything the same way the "Custom Objects linking to this object" card does now. Either way it does seem like a value-add that I'd be happy to split off as a later improvement.

Kani999:
Rescoped 🧹

I've force-pushed this branch to a **clean combined-only history** (5 commits).

The earlier exploration carried three things — per-COT **typed tabs**, the **combined tab**, and **hot-reload** machinery — and the combined tab alone delivers the user-facing value without the coordination complexity (no middleware/signals/shared cache; everything is live per-request).

**What changed vs. the previous push:**

- Dropped the typed-per-COT tabs and the hot-reload/registry-rebuild machinery.
- Kept + polished the combined "Custom Objects" tab; restyled it to match the native ObjectChildrenView (Device → Interfaces) look.
- Removed the old "Custom Objects linking to this object" panel (superseded).

The original typed-tabs work is preserved at `feature/related-object-tabs-v2` if we want to revisit it as a follow-up.

History (on top of `feature`):

1. add combined "Custom Objects" tab
2. supersede the old linking panel
3. tests
4. README
5. (unrelated) fix literal `{toggle_text}`/`{attrs.title}` placeholders in the COT fields actions column

Kani999:
Pushed a follow-up that removes two sources of redundant queries in the combined tab, both scoped to the tab's own code paths:

- **Multi-object Value column N+1.** Non-polymorphic multi-object fields were resolved one row at a time (`manager.all()` → one through-table query + one target query per row). They're now prefetched once per `(model, field)` group on the page via the custom M2M manager's `get_prefetch_querysets`, with permission filtering done in Python (as the polymorphic path already does).
- **Linked-field discovery.** The body render and the ViewTab badge both walked `_iter_linked_fields`, which calls `get_model()` per linked type — so each page regenerated those models twice. It's now memoized on the request object (with a fresh-build fallback when there's no request context). The cached triples are user-independent, so the per-row `.restrict()` in each caller still enforces permissions.

Added a regression test asserting multi-object value resolution stays constant in query count as rows grow.

Out of scope here: the plugin's navigation menu regenerates every dynamic model on each page render (`CustomObjectTypeMenuItems.__iter__` → `get_model()` per type), which remains the dominant per-page query cost. It's best addressed separately — e.g. request-scoped `get_model()` memoization shared across the menu, tab, and badge.

Kani999:
@bctiemann I think this is ready for review now. The combined "Custom Objects" tab is feature-complete — count badge, quicksearch, type/tag filters, sortable columns, HTMX pagination, per-row actions, and per-user permission filtering throughout.

Let me know if you'd like anything changed.

bctiemann:
@Kani999 could you resolve the merge conflicts please?

bctiemann:
Nice PR — the architecture is clean, the permission fix is important, and the test coverage is thorough. A few things worth addressing before merge:

**1. `_call_super_ready_once()` contract**

The comment in `ready()` explains that `super().ready()` must run before `register_tabs()` so the view registry is populated before `urls.py` freezes it. But `_call_super_ready_once()` is an internal wrapper whose behaviour isn't visible in this diff. If that wrapper ever skips or defers the call (e.g. for hot-reload idempotence), tabs would be registered before the registry is ready and `reverse()` calls would fail. Could you add a brief note or docstring confirming that `_call_super_ready_once()` always delegates to `super().ready()` synchronously?

**2. Silent failure in `ready()` can mask bugs**

```python
try:
    register_tabs()
except Exception:
    logger.exception("related_tabs.register_tabs() failed; continuing without tabs")
```

Swallowing a registration failure means users see no tabs and no indication why — a broken import or a logic error at startup gets logged and silently discarded. Consider raising a Django system check warning (`checks.Warning`) alongside the log so the problem surfaces in `manage.py check` and doesn't stay invisible in production.

**3. `_MAX_MULTIOBJECT_DISPLAY = 3` as a plugin config setting**

The truncation limit for multi-value display is currently a hardcoded constant. Users with denser relationships (10+ related objects per row) will always see a trailing "…" with no way to control it. A low-effort improvement would be to expose this as a `PLUGINS_CONFIG` setting with `3` as the default, e.g. `max_multiobject_display`.

---

Also — could you add a screenshot or two of the tab in action (either in the PR description or as a follow-up comment)? It would be helpful to see what the tab looks like on a built-in model's detail page and on a CO→CO page before merging.

bctiemann (changes requested):
See above.

Kani999:
Thanks for the review — addressed all three points. Summary below, screenshots at the end.

### 1. `_call_super_ready_once()` contract

Expanded the method's docstring to state explicitly that the first call always delegates to `super().ready()` **synchronously** — never deferred, scheduled, or skipped; only repeat calls are no-ops (the duplicate-registration guard). It also notes why the ordering is load-bearing: `super().ready()` populates `registry['views']`, which `urls.py` snapshots when `register_tabs()` runs, so registering tabs first would break `reverse()` for CustomObject feature URLs. No behaviour change. 

### 2. Silent failure in `ready()`

`ready()` now records the swallowed exception, and a new Django system check (`netbox_custom_objects.W002`, in `checks.py`) re-surfaces it in `manage.py check` **alongside** the existing `logger.exception` line — so a failed tab registration is visible rather than invisible. The check stays silent when registration succeeds or is skipped. 

Example output when registration fails:

```
?: (netbox_custom_objects.W002) The combined "Custom Objects" related tab failed to register at startup and will not appear on object detail pages (RuntimeError: ...).
	HINT: Resolve the underlying error (full traceback is in the NetBox log) and restart NetBox.
```

### 3. `_MAX_MULTIOBJECT_DISPLAY` as a config setting

Now a `PLUGINS_CONFIG` setting — `max_multiobject_display` (default `3`) — read via `get_plugin_config` at request time, and documented in `docs/installation.md`. 

```python
PLUGINS_CONFIG = {
    'netbox_custom_objects': {
        'max_multiobject_display': 10,  # related objects shown per row before truncating
    },
}
```

---

### Screenshots

**Built-in model detail page** — `dcim.Site`, with the combined **Custom Objects** tab: count badge, quicksearch, Type filter, sortable columns, and per-row actions, matching the native child-object list styling.

site_detail

**Type filter in action** — the same tab filtered to a single Custom Object Type.

site_detail_cotype_select

**CO→CO detail page** — a Custom Object referenced by other Custom Objects, including a polymorphic reference (`cotab2_childpoly` → `Ref`).

co_to_co_display

bctiemann:
@claude /review

bctiemann:
Thanks @Kani999 -- I think we're very close now. Please look over the Claude-generated review above; it looks to me as though items 1 and 2 are worth fixing, but the remaining ones are documentation/comments or optional optimizations. It's up to you whether to address any of those, but with 1 and 2 done I'm OK with merging this.

Kani999:
@bctiemann Thanks! I didn't have time to look at this yet, but I hope to get to it during next week. I'll address items 1 and 2 and update the PR

Kani999:
@bctiemann Items 1 and 2 are addressed, and the branch is synced with `upstream/feature` at v0.6.0 (merge commit, no force-push). I also picked up items 4 and 6 as cheap wins. Details below — item 1 turned out to need one level more than the suggested one-liner, which is worth a read.

### Item 1 — prefetch cache read (why the fix is two-part)

**Proposed:** replace the direct read of Django's private `_prefetched_objects_cache` in `_batch_multiobject_values` with `list(getattr(obj, field.name).all())`, on the assumption that the manager serves prefetched results from cache ("this call is free — no DB hit").

**Why that alone would have been wrong:** that assumption holds for Django's built-in `ManyRelatedManager`, whose `get_queryset()` checks the prefetch cache first — but our `CustomManyToManyManager.get_queryset()` didn't. It always built a fresh through-table subquery, so the one-liner would have compiled, returned correct data, and silently issued one through-table query per row — reintroducing exactly the N+1 that `_batch_multiobject_values` exists to eliminate. Verified, not just inferred: with the one-liner applied on its own, the `CaptureQueriesContext` guard in `test_batch_multiobject_values_are_correct_and_dont_scale_per_row` fails with `3 != 7 : multiobject Value resolution must not issue a query per row` — the query count scales with row count again.

**What's in the commit:**

1. `CustomManyToManyManager.get_queryset()` now serves prefetched results when present, mirroring Django's `ManyRelatedManager`:

   ```python
   def get_queryset(self):
       # Serve prefetched results when present, like Django's ManyRelatedManager.
       try:
           return self.instance._prefetched_objects_cache[self.prefetch_cache_name]
       except (AttributeError, KeyError):
           pass
       # ... existing subquery path unchanged ...
   ```

   This is the one legitimate place to touch the private attribute — inside the manager that co-owns the cache contract, exactly where Django itself reads it. No recursion hazard: while `prefetch_related_objects()` is building the cache the entry doesn't exist yet, so the `KeyError` path falls through to the fresh queryset. Chaining (`.filter()` / `.restrict()`) on a cached result clones the queryset, dropping `_result_cache` and re-deriving the same SQL.

2. With that in place, `_batch_multiobject_values` reads through the descriptor as suggested, and the tab code no longer references `_prefetched_objects_cache` at all:

   ```python
   prefetch_related_objects(objs, field.name)
   # The manager serves the prefetched result from cache, so .all() is query-free here.
   per_obj_targets = {id(obj): list(getattr(obj, field.name).all()) for obj in objs}
   ```

Query counts are unchanged (guarded by the existing `CaptureQueriesContext` test). Side benefit beyond this PR: any caller that does `prefetch_related` on a custom M2M field and iterates `.all()` now actually benefits from the prefetch instead of silently re-querying. The plugin's own custom-object list view already gains from it — its multiobject prefetch is no longer wasted, dropping the recorded query-count baseline for `ObjectFieldViewTestCase` from 50 to 47 queries (regenerated with `UPDATE_QUERY_COUNTS=1` in `tests/query_counts.json`).

### Item 2 — polymorphic MULTIOBJECT: cap before permission filtering

Applied as proposed:

```python
except AttributeError:
    # Polymorphic targets: not a queryset (no .restrict) — cap first so
    # permission filtering handles at most ``limit`` objects (viewable
    # targets past the cap are dropped, not backfilled).
    return restrict_to_viewable(user, list(qs[:limit]))
```

The Value column now truncates to `max_multiobject_display + 1` before `restrict_to_viewable`, so permission filtering handles at most the display slice instead of the full target list — with many targets across several models, that's the difference between one restricted `pk__in` query per target model over everything versus over a handful of objects.

Two honest footnotes:

1. **What this bounds is the permission filtering, not the DB fetch.** `PolymorphicResultList` materializes the full result on any access (slices included) — one through-table read plus one batch fetch per content type — so `qs[:limit]` doesn't reduce that part. It's a pre-existing property of the polymorphic manager shared by all its callers, not something this PR introduced; a true DB-level cap (slicing through-rows before the batch fetch) would touch that shared code and feels like follow-up material rather than part of this PR.
2. **The semantics aren't perfectly identical to the non-polymorphic path.** Non-polymorphic restricts in SQL and then slices, so it backfills up to `limit` viewable objects; the polymorphic path now filters only the first `limit` objects in display order, with no backfill — worst case a cell shows fewer values although viewable targets exist past the cap. That's the trade-off the review called intended design, and the code comment states it explicitly.

### Items 4 and 6 — also picked up

**Item 4 (in-memory pagination as a known limitation):** documented rather than mitigated, per the review's "at minimum warrants a sentence." The README's *Related Objects Tab* section now states the root cause and both of its consequences in one place: rows span heterogeneous models (one per Custom Object Type), so they can't be combined into a single SQL query — which is why the tab filters, sorts, and paginates in memory, why actions are per-row rather than bulk, and why rendering costs grow with how heavily an object is referenced. The optional "showing first N" row cap was deliberately not implemented — a behaviour change right before merge that the review didn't require. Also mirrored in the PR description's "What is intentionally not here" section.

**Item 6 (`_HARDCODED_TAB_NAMES` template coupling):** the comment above the frozenset now names `templates/netbox_custom_objects/customobject.html` and says to keep the set in sync with it. This coupling got a live test during the upstream sync: v0.6.0 adds a new hand-rendered Config Context tab to that template — exactly the drift scenario the review worried about. Verified that `CustomObjectConfigContextView` is *not* registered via `register_model_view` on the dynamic models, so `plugin_extra_tabs` never sees it, no duplicate tab can appear, and the set correctly needs no fifth entry.

**Items 3, 5, 8, 9:** left as-is per "up to you" — 3 and 5 are observations without a requested change (both already have the mitigations the review acknowledged: the documented mirror-invariant comment, the `.exists()` fast path plus per-request memoization), 8 is cosmetic and only affects CO→CO pages where Contacts/Journal/Changelog are hand-rendered anyway due to the slug-based routes, and 9 would tag only the two new checks while leaving the pre-existing branching check untagged — follow-up material if you want all three tagged together.

**Item 7:** one factual correction rather than a change — the Configure Table button is functional: `CustomObjectsTabTable` exists precisely for the per-user column-preference machinery, so the modal genuinely shows/hides/reorders the six columns and persists the choice. The part that's deliberately omitted (the saved-table-config "Save" split, which needs a single ObjectType) is noted in the PR description, matching native behaviour.

### Upstream sync

Same approach as the previous sync (merge, no history rewrite). Two small conflicts: the tab list in `customobject.html` (kept both the new Config Context tab and the combined-tab lines) and the `test_api.py` import header. Full suite passes post-merge and post-fix (1150 tests, 63 skipped).

bctiemann (approved):
Looks good — approving. Jan's handling of the two substantive items was thorough and technically correct.

**Item 1 (prefetch cache):** the two-part fix is exactly right. Making `CustomManyToManyManager.get_queryset()` serve prefetched results from cache (mirroring Django's own `ManyRelatedManager`) was the necessary prerequisite before the descriptor-based one-liner in `_batch_multiobject_values` could work without reintroducing the N+1. The query-count drop from 50→47 in `ObjectFieldViewTestCase` is a concrete side benefit beyond this PR. The explanation in the comment is accurate and worth keeping.

**Item 2 (polymorphic MULTIOBJECT cap before permission filtering):** applied correctly, footnotes are honest. The semantic difference from the non-polymorphic path is an accepted and commented trade-off.

One minor docs note for a follow-up (not blocking merge): a CO that references the host object through more than one field appears once per linking field in the tab. The **Field** column makes this self-labelling in the UI, but a one-sentence note in the README's *Related Objects Tab* section alongside the existing in-memory pagination caveat would help users who encounter it without knowing the data model.

bctiemann (approved):

PR #484, opened 2026-04-30
Fixes: #471 - Selectable on_delete behavior for Object-type fields (and: Change Object-type FK constraints from CASCADE to SET NULL)

### LINKED ISSUE #471: Data loss: cascading deletion of Custom Objects when deleting Contact (Tenancy)
### Plugin Version

0.4.6

### NetBox Version

4.4.8

### Python Version

3.12.3

### Steps to Reproduce

1. Go to Custom Object Types in NetBox
2. Create a new Custom Object Type (e.g., "Business Service")
3. Add a new field:

- Name: Owner
- Type: Object
- Related object: Tenancy → Contact

4. Create a new object of this type and assign a Contact as Owner
5. Navigate to Tenancy → Contacts
6. Delete the Contact used in step 4

Result

- The Contact is deleted
- The related Custom Object (e.g., Business Service) is also deleted automatically

### Expected Behavior

Deleting a Contact should NOT delete Business Services.

Preferred behavior:

- owner field becomes NULL (ON DELETE SET NULL), or
- deletion is blocked if dependencies exist (PROTECT)

### Observed Behavior

- Contact is deleted
- All linked Business Services are deleted automatically (cascade)

bctiemann:
## Summary

Extends the fix from #484 by replacing the blanket SET NULL behavior with a user-selectable on_delete_behavior for single-Object-type fields.

- Adds on_delete_behavior CharField to CustomObjectTypeField with three choices: Set null (default), Cascade, Protect
- The DB-level FK constraint in _ensure_field_fk_constraint now uses the field's choice (SET NULL / CASCADE / RESTRICT)
- The Django field definition in ObjectFieldType.get_model_field() also uses the choice
- Changing on_delete_behavior on an existing field re-applies the FK constraint on save
- Exposed in the UI form (single-object fields only) and REST API
- Migration 0011 adds the column; all existing fields default to set_null (consistent with what #484 set in the DB)
- Tests cover all three behaviors via raw SQL DELETE to exercise the actual DB constraints

## Notes

- MultiObject fields (M2M through-table) are unaffected — CASCADE on the join-table FKs is correct for those
- PROTECT maps to PostgreSQL RESTRICT (raises IntegrityError on delete attempt, preventing deletion)
- This PR depends on #484 and should be reviewed/merged after it

## Test plan

- [ ] Run python manage.py test netbox_custom_objects.tests.test_deletion --keepdb
- [ ] Verify 8/8 tests pass including the new cascade and protect tests
- [ ] Manually verify on_delete_behavior dropdown appears for Object-type fields in the UI
- [ ] Verify multiobject fields do NOT show the on_delete_behavior dropdown

Generated with Claude Code

See also previous smaller scope of this PR:

## Summary

Fixes #471 — deleting a referenced NetBox object (e.g. a Contact) was silently deleting every Custom Object that held a reference to it via an Object-type field.

- **`field_types.py`**: `ObjectFieldType.get_db_field()` — both the `LazyForeignKey` (self-referential) and `ForeignKey` (cross-referential / external NetBox model) now use `on_delete=SET_NULL`. Fields are already `null=True, blank=True` so this is always valid.
- **`models.py`**: `_ensure_field_fk_constraint()` now issues `ON DELETE SET NULL` in the raw `ALTER TABLE` statement instead of `ON DELETE CASCADE`. Constraint renamed from `*_fk_cascade` to `*_fk`.
- **`migrations/0010_fix_object_fk_set_null.py`**: data migration that iterates all existing Object-type fields, introspects the live FK constraint via `information_schema`, and recreates any CASCADE constraint as SET NULL. Skips constraints already correct; logs a warning and continues if a table cannot be found.
- **`tests/test_deletion.py`**: updated `test_delete_co_referenced_by_another_co` and `test_delete_referenced_core_object` to assert SET NULL semantics (CO survives, FK column becomes NULL) instead of CASCADE (CO deleted).

**Note:** M2M through-table `source`/`target` FKs are intentionally left as CASCADE — those rows are join-table entries, not Custom Objects themselves, so removing either side of an M2M association should delete the join row.

## Test plan

- [x] `python manage.py test netbox_custom_objects.tests.test_deletion --keepdb` — all 5 tests pass
- [ ] Manual: create a Custom Object with an Object-type field referencing a Contact; delete the Contact; verify the Custom Object survives with the field nulled out
- [ ] CI on both `main` and `feature` NetBox refs


netbox_custom_objects/models.py
```diff
@@ -726,69 +727,89 @@ def get_model_with_serializer(self):
         self.register_custom_object_search_index(model)
         return model
 
-    def _ensure_field_fk_constraint(self, model, field_name):
+    _ON_DELETE_SQL = {
```
arthanson: can we move that to the top of the class, instead of mixed in with the functions.

netbox_custom_objects/field_types.py
```diff
@@ -486,6 +487,13 @@ def render_table_column(self, value):
         return ", ".join(value)
 
 
+_ON_DELETE_MAP = {
```
arthanson: move to top of class instead of mixed in with classes

arthanson (changes requested):
just a couple minor move of constants

arthanson (approved):

PR #485, opened 2026-05-01
Closes: #471 -  selectable on_delete behavior for Object-type fields

bctiemann:
## Summary

Extends the fix from #484 by replacing the blanket SET NULL behavior with a user-selectable on_delete_behavior for single-Object-type fields.

- Adds on_delete_behavior CharField to CustomObjectTypeField with three choices: Set null (default), Cascade, Protect
- The DB-level FK constraint in _ensure_field_fk_constraint now uses the field's choice (SET NULL / CASCADE / RESTRICT)
- The Django field definition in ObjectFieldType.get_model_field() also uses the choice
- Changing on_delete_behavior on an existing field re-applies the FK constraint on save
- Exposed in the UI form (single-object fields only) and REST API
- Migration 0011 adds the column; all existing fields default to set_null (consistent with what #484 set in the DB)
- Tests cover all three behaviors via raw SQL DELETE to exercise the actual DB constraints

## Notes

- MultiObject fields (M2M through-table) are unaffected — CASCADE on the join-table FKs is correct for those
- PROTECT maps to PostgreSQL RESTRICT (raises IntegrityError on delete attempt, preventing deletion)
- This PR depends on #484 and should be reviewed/merged after it

## Test plan

- [ ] Run python manage.py test netbox_custom_objects.tests.test_deletion --keepdb
- [ ] Verify 8/8 tests pass including the new cascade and protect tests
- [ ] Manually verify on_delete_behavior dropdown appears for Object-type fields in the UI
- [ ] Verify multiobject fields do NOT show the on_delete_behavior dropdown

Generated with Claude Code


bctiemann:
@claude /review

bctiemann:
Addressed all actionable issues from the review; also noticed that the portable schema needed to be updated to include the new deletion behavior, and at the same time hardened to detect future infrastructural changes to COT/COTF/other internal models. Now has tests to detect such changes.

bctiemann:
@arthanson Note that this is based off of another branch that initially fixes the "bug" by setting SET_NULL as the default behavior (a much smaller change). But if you approve this we can merge that into this and then merge both into feature, if we want to simplify the migration path.

arthanson (approved):

PR #495, opened 2026-05-07
Fixes #49: Support CUSTOM_VALIDATORS keyed by COT slug

### LINKED ISSUE #49: Add `CustomValidationMixin` to Custom Object model
### Plugin Version

commit 003c6879d5fe3777a1d9274ccb9dc1b2b208c4ac (HEAD -> netbox-models, origin/netbox-models)

### NetBox Version

v4.2.8-Docker-3.2.0

### Python Version

3.12

### Steps to Reproduce

- Create a Custom Object Type X
- Add a `text` field called `testtextfield`
- Add a JSON custom validator section to `configuration.py`
  - Added the `dcim.site` one to check that it was working at all - it does work for `dcim.site`
  - I believe the path should be `<app_name>.<custom_object_type_name>` but I could use confirmation
  - I tried both `netbox_custom_objects.x` and `netbox_custom_objects.X`

```
CUSTOM_VALIDATORS = {
    "dcim.site": [
        {
            "name": {
                "min_length": 5,
                "max_length": 30,
                "regex": "^[A-Z]{3}-.*"  # Must start with 3 uppercase letters
            }
        }
    ],
    "netbox_custom_objects.x": [
        {
            "testtextfield": {
                "min_length": 5,
                "max_length": 30,
            }
        }
    ]
}
```

### Expected Behavior

When I create a Custom Object of type X with the `testtextfield` smaller than 5 characters, the validator should not allow it to proceed

### Observed Behavior

The Custom Object is created

bctiemann:
Fixes: #49 
## Summary

- Overrides `clean()` in `CustomObject` to run `CUSTOM_VALIDATORS` using the COT slug as the key (e.g. `netbox_custom_objects.my-cot-slug`), alongside the standard `post_clean` signal which uses the internal `table{id}model` name that users could not discover
- Adds three imports to `models.py`: `get_config`, `run_validators`, `get_config_value_ci`
- Documents the feature in `docs/api.md` with a usage example
- Adds two tests to `CustomObjectTestCase` covering the enforced and passing cases

## Test plan

- [ ] `test_custom_validators_slug_key_enforced` — validator raises `ValidationError` when rule is violated
- [ ] `test_custom_validators_slug_key_passes_for_valid_value` — no error when value satisfies rule
- [ ] Full `test_models` suite passes

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -166,6 +166,27 @@ The response will include the created object with its assigned ID and additional
 }
 ```
 
+### Custom Validation
+
+NetBox's [`CUSTOM_VALIDATORS`](https://netboxlabs.com/docs/netbox/en/stable/configuration/data-validation/#custom_validators) setting is supported for Custom Objects. Use `netbox_custom_objects.<cot-slug>` as the key, where `<cot-slug>` is the slug of the Custom Object Type:
+
+```python
+# configuration.py
+CUSTOM_VALIDATORS = {
+    "netbox_custom_objects.server": [
+        {
+            "hostname": {"min_length": 3, "max_length": 64},
+            "cpu_cores": {"min": 1},
+        }
+    ]
+}
+```
+
+Validators are enforced on both API writes and UI form submissions. Any violation raises a 400 error (API) or a form validation error (UI).
+
+> [!NOTE]
+> The key must use the COT **slug** (e.g. `server`), not the COT name or the internal model name.
+
 ### Linked Objects
 
 Any object in NetBox (whether a core object such as Device or Site, or a Custom Object) can be pointed to by one or more Custom Objects via `object` or `multiobject` fields. You can get a list of all these "linked" Custom Objects (analogous to the "Custom Objects linking to this object" table in the NetBox UI) by querying the following endpoint:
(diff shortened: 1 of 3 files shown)


bctiemann:
@claude /review

bctiemann:
Thanks for the thorough review! Here is a summary of what was addressed and the rationale for each point.

**`@override_settings` and `get_config()` compatibility**

Not an issue. `Config.__getattr__` reads from `django.conf.settings` live on every attribute access — no caching involved. So `@override_settings` correctly patches the values seen by `get_config().CUSTOM_VALIDATORS` in tests, and the tests are exercising the real code path.

**Comment tightened**

Rewrote the `clean()` comment to remove the awkward mid-phrase line break and clarify that the slug-key validators run *in addition to* (not instead of) the `post_clean` signal pass.

**Unrelated registry change**

Left in this PR intentionally — it was required to fix the `tests (feature)` CI failure caused by `registry['models']` being removed in the NetBox feature branch head. Splitting it out would leave the PR in a broken-CI state.

**Additional tests**

Added three new tests to cover the gaps identified in the review:

- `test_custom_validators_non_text_field_enforced` — validates an integer `count` field with a `min` rule
- `test_custom_validators_no_key_configured_passes` — confirms `clean()` is a no-op when no matching key exists in `CUSTOM_VALIDATORS`
- `test_custom_validators_slug_key_case_insensitive` — confirms an uppercase key (`NETBOX_CUSTOM_OBJECTS.TEST-OBJECTS`) still matches via `get_config_value_ci`

Total validator tests: 5.

arthanson (approved):

PR #501, opened 2026-05-11
Fixes #477: Hide through-table entries from delete confirmation page

### LINKED ISSUE #477: ValueError: Cannot query "<object_name>": Must be "Table4Model" instance when deleting Custom Object via UI
### Plugin Version

0.4.9

### NetBox Version

4.5.7

### Python Version

3.12.3

### Steps to Reproduce

1. Create a Custom Object Type with a `multiobject` field (e.g. VM Dependencies → Virtualization > Virtual Machine)
2. Create any Custom Object of that type
3. Navigate to the object's detail page
4. Click the "Delete" button

Image

### Expected Behavior

The object should be deleted successfully, or a confirmation page should be displayed before deletion.

### Observed Behavior

A server error is raised immediately when clicking Delete:

```
<class 'ValueError'>
Cannot query "<object_name>": Must be "Table4Model" instance.
```

The error occurs in `CustomObjectDeleteView` when NetBox's `_get_dependent_objects()` uses Django's `Collector` to find related objects before showing the confirmation page. The Collector traverses M2M relations (e.g. `multiobject` fields) and passes the object's string representation instead of a proper model instance into the ORM query.

The object name in the error message always matches the name of the object being deleted, confirming the issue is in the delete view's dependent object resolution.

**Workaround:** Deletion works correctly via the Django shell:
```python
from netbox_custom_objects.models import *
cot = CustomObjectType.objects.get(slug="<your-slug>")
model = cot.get_model()
obj = model.objects.get(id=<id>)
obj.delete()
```

**Suggested fix** — add `_get_dependent_objects` override to `CustomObjectDeleteView` in `views.py`:
```python
def _get_dependent_objects(self, obj):
    try:
        return super()._get_dependent_objects(obj)
    except ValueError:
        return {}
```

bctiemann:
The crash in the bug report is no longer directly reproducible in Custom Objects v0.5.0. However, this PR adds a unit test to cover the relevant code path against future regressions, and also cleans up the delete confirmation dialog to no longer show the irrelevant through-table relation objects that will be deleted (and which were the cause of the bug in the first place).

## Summary

- Overrides `_get_dependent_objects` in `CustomObjectDeleteView` to filter out M2M through-table model entries (those with `app_label == APP_LABEL` and `model_name` starting with `through_`)
- These are implementation details, not business objects — showing them on the confirmation page exposes internal naming like `Through_custom_objects_3_devices` which is confusing to users
- Adds a regression test verifying through-table names are absent from the delete confirmation page response

## Test plan

- [ ] Run `test_delete_confirmation_page_with_populated_multiobject_field` — confirms the page returns HTTP 200 and contains no `through_` strings in the response body
- [ ] Manually verify: create a custom object with a populated multiobject field, navigate to its delete confirmation page, confirm no through-table labels appear

### DIFF
--- a/netbox_custom_objects/tests/test_views.py
+++ b/netbox_custom_objects/tests/test_views.py
@@ -742,7 +742,7 @@ class ObjectFieldViewTestCase(CustomObjectsTestCase, ViewTestCases.PrimaryObject
         self.assertHttpStatus(response, 200)
         # M2M through-table rows must not appear on the confirmation page —
         # they are implementation details, not user-facing business objects.
-        self.assertNotIn(b'through_', response.content.lower())
+        self.assertNotIn(b'through_', response.content)
 
 
 class ObjectSelectorViewTestCase(TestCase):
--- a/netbox_custom_objects/tests/test_views.py
+++ b/netbox_custom_objects/tests/test_views.py
@@ -649,7 +649,7 @@ class ObjectFieldViewTestCase(CustomObjectsTestCase, ViewTestCases.PrimaryObject
             # Skip if DCIM models are not available
             cls.site = None
             cls.device = None
-            cls.instance = None
+            cls.instance_1 = None
 
     def setUp(self):
         """Set up test data."""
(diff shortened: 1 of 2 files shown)


bctiemann:
@claude /review

jnovinger (approved):

PR #504, opened 2026-05-11
Fixes: #408 - Resolve cross-COT FK fields missing after server restart

### LINKED ISSUE #408: Nested custom objects are broken after 0.4.0
### Plugin Version

0.4.6

### NetBox Version

4.5.3

### Python Version

3.12

### Steps to Reproduce

Create Custom object: systeme<br>display_name: Systeme<br>display_name_plurar: Systemes

> Create field:

* Type: Text
* Name: name
* Label: Systeme
* Primary name
* required
* must be unique

Create Custom object sous_systeme<br>display_name: Sous-Systeme<br>display_name_plurar: Sous-Systemes

> Create field:

* Type: Text
* Name: name
* Label: Sous-Systeme
* Primary name
* required

> Create field:

* Type: Object
* Related Object-Type: Custom Object systeme
* Name: systeme
* Label: Systeme
* required

Create Custom object composant<br>display_name: Composant<br>display_name_plurar: Composants

> Create field:

* Type: Text
* Name: name
* Label: Composant
* Primary name field
* required

> Create field:

* Type: Object
* Related Object-Type: Custom Object sous_systeme
* Name: sous_systeme
* Label: Sous-Systeme
* required

Import the following as a systeme custom objects:

name<br>AM

Import the following as a sous_systeme custom objects:

name;systeme<br>AM-CAB;AM<br>AM-CUV;AM<br>AM-DCA;AM<br>AM-GAI;AM<br>AM-LMA;AM<br>AM-PAL;AM<br>AM-TCA;AM

Import the following as a composant custom objects:

sous_systeme;name<br>AM-CAB;BCO<br>AM-CAB;CAE<br>AM-CAB;ECL<br>AM-CAB;HPR<br>AM-CAB;INI<br>AM-CAB;INT<br>AM-CAB;LDI<br>AM-CAB;PBA<br>AM-CAB;PCP<br>AM-CAB;POR<br>AM-CAB;PSG<br>AM-CAB;SEL<br>AM-CAB;VEN<br>AM-CAB;VLU<br>AM-CUV;AMO<br>AM-CUV;APF

View the objects, all data is there.<br>Restart NetBox<br>Object sous_systeme will no longer have data in the custom object pointing to systeme.

I do not know where it broke but there is a bug fix referencing custom-objects in [NPL-637](https://linear.app/netboxlabs/issue/NPL-637/saving-a-custom-object-type-field-breaks-custom-object-relationships) for caching.

This will make the custom-objects work again after a reboot.<br>***So maybe a cache-clear at startup is all that is needed.***

### Expected Behavior

No corrupted database and data not disappearing on restart.

### Observed Behavior

Debugging the data shows that the table still contains the reference to objects and all table data seems to look good. Something gets broken/corrupted in the database.<br>Trying to re-assign the custom object configuration to Object sous_systeme will result in a traceback.<br>Only way to bring things back is to force a cache clear for the object that is problematic. (sous-\_system).

bctiemann:
## Summary

- Uses `LazyForeignKey` for all cross-COT foreign keys so alphabetical startup ordering cannot cause a FK-source model to be processed before its FK-target model exists in the app registry
- `ready()` now runs two passes: Pass 1 generates all COT models (deferred FK resolutions are caught silently), Pass 2 re-resolves all `LazyForeignKey` fields now that every model is registered; serializers are registered only after both passes complete
- `get_model()` no longer caches partial models generated with `skip_object_fields=True`, preventing a stale cache hit from permanently hiding FK fields on subsequent lookups

## Why two passes are sufficient for any chain depth

A reviewer might ask: *"Two passes handles A->B->C — but what about A->B->C->D?"*

The two passes do fundamentally different things, so chain depth is irrelevant:

- **Pass 1** generates ALL COT model classes. By the time it completes, every COT (regardless of chain depth) exists as a Python class in `apps.all_models`. Some `LazyForeignKey` fields still hold string references — those where the target wasn't yet registered when Django's internal `lazy_model_operation` fired during `type()` class creation — but every model class exists.
- **Pass 2** resolves those string references to class references. It iterates all COT models and calls `_resolve_model` on every `LazyForeignKey`. Since every target is already registered (Pass 1 put them there), every `apps.get_model(target)` lookup succeeds. Each resolution is **independent** — resolving A's FK to B does not require B's FK to C to be resolved first, because all Pass 2 does is set `remote_field.model` to an already-registered class (updating a pointer to a class that already exists).

A third pass would only be necessary if Pass 2 itself triggered further model **generation** that could fail due to ordering — but it does not. Pass 2 only updates `remote_field.model` on already-generated classes.

Correct mental model: Pass 1 solves the generation-ordering problem for **all** chain depths at once by deferring unresolvable FKs; Pass 2 resolves **all** deferred FK pointers in a single flat sweep because every target already exists.

## Test plan

- [ ] `NestedCOTStartupOrderingTestCase` (new): three-level `alpha_type->beta_type->gamma_type` chain simulates worst-case alphabetical startup; all three assertions verify FK fields survive the two-pass startup simulation
- [ ] Run `python manage.py test netbox_custom_objects.tests.test_models.NestedCOTStartupOrderingTestCase`
- [ ] Manual: create a chain of COTs with cross-COT FK fields and restart the server; verify FK fields appear on the detail/edit views of all COTs

Closes #408


bctiemann:
@claude /review

jnovinger (approved):

PR #505, opened 2026-05-11
Fixes #483: Patch stale through-model FK path_infos on COT model regeneration

### LINKED ISSUE #483: Deletion is not Possible due to active n:m Relationshings
### Plugin Version

0.4.8

### NetBox Version

4.5.3

### Python Version

3.12.12

### Steps to Reproduce

1. Create a Custom Object 
2. Add A field that is referencing to another object
3. Add an object with a value in the Field
4. Delete this object you

### Expected Behavior

Deletion is possible and Relations will be cleand up

### Observed Behavior

Image

the error appears until you have removed all n:m Relationships to other objects

bctiemann:
## Background

Issue #483 reports that deleting a Custom Object (CO) fails with an `IntegrityError` when the CO has any N:M (cross-COT multiobject) relationships. The reporter shows that the error disappears only after manually removing all M2M relationships first, which should be unnecessary — Django's deletion collector is supposed to cascade through them automatically.

## Root cause

Custom Object models are generated dynamically and cached per COT. When a COT's cached model is **invalidated and regenerated** (e.g. because a field was added or changed), Django's `apps` registry gets a new model class. However, any **cross-COT M2M through model** that was previously registered keeps its `target` FK pointing at the **old model class**.

When Django's deletion collector traverses the new model's related objects to build the cascade plan, it finds the through model's `target` FK and calls:

```python
through_model.objects.filter(target__in=[obj_to_delete])
```

Internally, `check_rel_lookup_compatibility` compares `obj_to_delete`'s class against the FK's registered target using `PathInfo.to_opts` — a `@cached_property` on `ForeignKey` that stores `PathInfo(to_opts=old_model._meta)`. Because `old_model._meta` and `new_model._meta` are different objects even though they describe the same table, the check raises:

```
ValueError: Cannot query <NewModel>: Must be <OldModel> instance.
```

Django then wraps this as an `IntegrityError` at the database level, producing the screenshot in the issue.

Simply updating `target.remote_field.model = new_model` is insufficient — the stale `path_infos` cached property still holds the old `_meta`. Both must be patched together.

## Fix

Inside `get_model()`, after the regenerated model is registered with Django's app registry, we walk all inbound non-polymorphic multiobject fields (i.e. fields on *other* COTs whose `related_object_type` points at this COT's content type) and patch their through model's `target` FK in-place:

```python
target_field.remote_field.model = model                # point to new class
target_field.related_model = model                     # update denormalized reference
target_field.__dict__.pop('path_infos', None)          # clear @cached_property
target_field.__dict__.pop('reverse_path_infos', None)  # clear @cached_property
```

Popping from `__dict__` clears the `@cached_property` without touching the descriptor, so Django rebuilds `PathInfo` lazily from the updated `remote_field.model` on next access.

## Why this fixes the reported IntegrityError

The reporter's `IntegrityError` occurs precisely in this scenario:

1. A source CO has a cross-COT M2M field pointing at a target COT
2. The target COT's model is regenerated (cache miss — common after plugin restart or field change)
3. The user tries to delete a target CO

Step 3 fails because the through model's `target` FK still references the pre-regeneration target model class. After this fix, `get_model()` always repoints the through FK to the current model class before returning it, so the deletion collector's lookup compatibility check passes.

## Test coverage

`test_delete_target_co_after_source_model_regeneration` reproduces this exact failure path:

1. Creates source and target COTs with a cross-COT M2M field between them
2. Creates source and target COs and links them via the M2M
3. **Explicitly clears the target COT's model cache** (`CustomObjectType.clear_model_cache(cot_target.id)`) to force regeneration — simulating the post-restart or post-field-change scenario from the issue
4. Calls `get_model()` on the target COT (triggering regeneration and the new patching logic)
5. Deletes the target CO and asserts it succeeds with no `ValueError` or `IntegrityError`

Without the fix, step 5 raises `ValueError: Cannot query …: Must be OldModel instance`. With the fix it passes.

The three additional tests cover adjacent deletion paths (deleting the source CO, deleting without a cache miss, and deletion through a multi-hop M2M chain) to prevent regressions.

## Test plan

- [x] `test_delete_source_co_with_cross_cot_multiobject_field` — deleting the source CO clears through rows and succeeds
- [x] `test_delete_target_co_with_cross_cot_multiobject_field` — deleting the target CO (no cache miss) succeeds
- [x] `test_delete_target_co_after_source_model_regeneration` — **the failing case**: deleting the target CO after the target COT model is regenerated
- [x] `test_delete_co_in_multi_hop_cross_cot_m2m_chain` — deletion works through a multi-hop M2M chain


bctiemann:
@claude /review

jnovinger (approved):

PR #506, opened 2026-05-11
Fixes #443: Guard validate() against non-dict data in generated serializers

### LINKED ISSUE #443: Update via API Does not work for some Objects
### Plugin Version

0.4.6

### NetBox Version

4.5.3 or 4.5.5

### Python Version

3.12.12

### Steps to Reproduce

Create 6 Custom Object Types
1. Applications with 
- Name(Text)
- Description(Text)
- Proivded to Address (Single Object IPAM-> IP Address)
- Provided to Application components (Multiple Objects CustomObjects->Application Component)
- Provided to Identity Groups (Multiple Objects CustomObjects->Application Identity Groups)
- Provided to Prefix (Single Object IPAM-> Prefix)
- Provided to Tags (Multiple Objects Extras->Tag)
2. Application Components
- Name (Text)
- Application Disks (Multiple Objects Virtualization->Virtual Disks)
- Application DNS Record (Multiple Object Netbox DNS ->Record
- Application FQDNs (Multiple Object Custom Objects -> Application FQDN)
- Application ID (Selection (just some Texts)
- Application Port Groups (Multiple Objects Custom Objects -> Application Port Groups)
- Application Ports (Mulltiple Objects Custom Objects -> Application Ports)
- Application Tags (Multiple Objects Extras-> Tags)
- IP Addresses (Multiple Objects -> IPAM IP Addresses)
3. Application FQDN
- FQDN (Text)
- Description (Text)
4. Application Identity Group
- Name (Text)
5. Application Port
- Name (text)
- Description (Text)
- TCP Port Range (Text)
- UDP Port Range (Text)
6. Application Port Groups
- Name (Text)
- Application Ports (Multiple Objects Custom Objects -> Application Port)
- Description (Text)

Add Objects via API should work on all
a PUT Objects via API will not work on Applications,  Application Components
i think not all of the Data Fields are reqired and maybe its Refrence between Application and Appliction Components and vice versa that causes the Issue
i also have observed that on a POST the Tag will not be applied

### Expected Behavior

That Update on All of the Objects are working

### Observed Behavior

Update does not work

bctiemann:
### Fixes: #443

This PR tightens the validation behavior for PATCH/PUT requests containing M2M field PKs. It does not directly address the behavior described in #443, but investigation into it revealed some unit tests gaps that are covered by this fix.

A followup comment has been posted to #443 to investigate further into the OP's observation and see whether it is reproducible.

## Summary

- The `validate` closure in `get_serializer_class()` is attached to every generated serializer class, including nested child serializers used for `many=True` cross-COT multiobject fields
- When DRF processes a PATCH body containing a list of PKs for such a field, it resolves each PK to a model instance via `to_internal_value`, then calls `validate(model_instance)` on the child serializer
- Without the guard, the existing code calls `data.update(saved)` on the model instance, raising `AttributeError: 'TableXModel' object has no attribute 'update'` and causing the PATCH/PUT to fail with a 500 error
- Fix: at the top of `validate`, check `if type(data) is not dict` and delegate to `NetBoxModelSerializer.validate` — mirroring the identical pattern already used in `TaggableModelSerializer.validate` (line 73 of `features.py`)

## Test plan

- [ ] `test_patch_updates_cross_cot_m2m_field` — PATCH with a cross-COT M2M value correctly updates the through table
- [ ] `test_patch_clears_cross_cot_m2m_field` — PATCH with an empty list clears the M2M relation
- [ ] `test_patch_scalar_field_preserves_m2m` — PATCH of a scalar field leaves existing M2M values unchanged
- [ ] `test_get_response_includes_cross_cot_m2m_field` — GET response includes the cross-COT M2M field with correct nested representation

Closes: #443


bctiemann:
@claude /review

jnovinger (approved):

jnovinger (changes requested):
Has merge conflicts after merging #510 .

jnovinger (approved):

PR #509, opened 2026-05-12
Fixes #507: Make migration 0011 idempotent and safe for long table names

### LINKED ISSUE #507: Migration 0011_non_deferrable_fk_constraints fails with DuplicateObject on partial re-run / long table names
### Plugin Version

0.5.0

### NetBox Version

4.5.7

### Python Version

3.12

### Steps to Reproduce

1. Run NetBox with netbox-custom-objects 0.4.9 installed and have at least one CustomObjectType with a sufficiently long name such that the dynamic table name plus an FK column name approaches PostgreSQL's 63-character identifier limit. In my case the table was `custom_objects_14_technical_account_accounts` (44 chars).
2. Have multiple foreign-key fields on that custom object type (e.g. tenant_id, type_id, applicant_user_id).
3. Upgrade to 0.5.0: `pip install --upgrade netboxlabs-netbox-custom-objects==0.5.0`
4. Run `python manage.py migrate`.

### Expected Behavior

Migration 0011_non_deferrable_fk_constraints completes successfully, converting all DEFERRABLE FK constraints on `custom_objects_*` tables to non-deferrable with the `_fk_cascade` naming convention.

### Observed Behavior

Migration 0011 fails partway through with:

```
psycopg.errors.DuplicateObject: constraint "custom_objects_14_technical_account_accounts_type_id_fk_cascade" for relation "custom_objects_14" already exists
```

Full traceback:

```
Applying netbox_custom_objects.0011_non_deferrable_fk_constraints...Traceback (most recent call last):
  File "/opt/netbox/venv/lib/python3.12/site-packages/netbox_custom_objects/migrations/0011_non_deferrable_fk_constraints.py", line 43, in fix_deferrable_fk_constraints
    cursor.execute(f"""
django.db.utils.ProgrammingError: constraint "custom_objects_14_technical_account_accounts_type_id_fk_cascade" for relation "custom_objects_14" already exists
```

**Analysis:**

There appear to be two related issues:

1. **PostgreSQL 63-character identifier truncation is not accounted for.** The migration constructs new constraint names as `{table_name}_{column_name}_fk_cascade`. For tables whose names are already long, the resulting identifier exceeds 63 characters and PostgreSQL silently truncates it. Two different columns on the same table can end up with the same truncated constraint name, or collide with names from a previous naming scheme.

2. **Inconsistent naming between `information_schema` and existing constraints.** On my database, `information_schema.table_constraints.table_name` reports `custom_objects_14` for the parent table, but the constraints already present on it use the longer prefix `custom_objects_14_technical_account_accounts_*`. This suggests an earlier version of the plugin (or Django's automatic FK naming) produced constraints with names derived from the model name rather than the physical table name. When the new migration logic generates names from the shorter `custom_objects_14`, those names truncate differently and collide.

3. **No safety against partial failure.** Once the migration crashes mid-table, some constraints have already been converted to the non-DEFERRABLE `_fk_cascade` form, while others remain DEFERRABLE. Re-running the migration finds the still-DEFERRABLE ones (the WHERE clause filters `is_deferrable = 'YES'`), but then collides again with the names produced in the partial run.

**State of my database after the crash:**

`custom_objects_14` ended up with a mix of constraint name schemes:
- 3 with the new `_fk_cascade` naming (partially applied, with the longer "technical_account_accounts" prefix)
- 3 with the original auto-generated hash-suffix naming

**Suggested fixes:**

- Use `DROP CONSTRAINT IF EXISTS <new_name>` before `ADD CONSTRAINT <new_name>` so a name collision doesn't abort.
- Or wrap each iteration in a savepoint so one collision doesn't kill the whole migration.
- Or generate constraint names that are guaranteed unique and within 63 chars, e.g. by hashing the (table, column) pair like Django itself does.
- Consider migrating any pre-existing `_fk_cascade` constraints with the old longer naming convention to the new scheme, or skip them.

bctiemann:
## Summary

- **Truncation collision**: `fix_deferrable_fk_constraints` built new constraint names as `{table_name}_{column_name}_fk_cascade` with no length check. Through-tables with names like `custom_objects_14_technical_account_accounts` (44 chars) plus column names easily exceeded PostgreSQL's 63-char identifier limit. Two columns sharing enough of a prefix produced identical truncated names, causing `DuplicateObject` on the second `ADD CONSTRAINT`. Fixed by introducing `_safe_constraint_name()`, which uses the same truncate-and-MD5-hash strategy already present in `field_types._safe_pg_identifier`.
- **Old DROP without IF EXISTS**: If the migration failed mid-loop and was re-run, the old DEFERRABLE constraint might already be gone, causing `DROP CONSTRAINT "..."` to error. Fixed with `IF EXISTS`.
- **No pre-drop of new name**: If a partial prior run left behind a non-DEFERRABLE `_fk_cascade` constraint (invisible to the `is_deferrable = 'YES'` query), a later iteration could attempt to `ADD CONSTRAINT` with the same name and fail. Fixed by pre-dropping the new constraint name with `IF EXISTS` before the `ADD`.

## Test coverage

Added `netbox_custom_objects/tests/test_migrations.py` with 11 tests across three scenarios, all run against a real PostgreSQL database:

| Group | What it tests |
|---|---|
| `BasicConversionTestCase` | DEFERRABLE constraints on a short-named table are converted to non-DEFERRABLE with `_fk_cascade` names |
| `LongTableNameTestCase` | 47-char through-table + two columns whose naive `_fk_cascade` names share the same 63-char truncation prefix — verifies `_safe_constraint_name` produces distinct names <= 63 chars and the migration completes without `DuplicateObject` |
| `PartialRerunTestCase` | Pre-existing non-DEFERRABLE `_fk_cascade` constraint alongside the original DEFERRABLE one — verifies the migration is idempotent, both columns end up non-DEFERRABLE, and the partially-applied column collapses to exactly one FK constraint |

All 11 tests pass locally (`Ran 11 tests in ~11s — OK`).

Closes: #507

### DIFF
--- a/netbox_custom_objects/migrations/0011_non_deferrable_fk_constraints.py
+++ b/netbox_custom_objects/migrations/0011_non_deferrable_fk_constraints.py
@@ -7,7 +7,27 @@ because it has pending trigger events". Recreate any existing DEFERRABLE FKs
 on custom_objects_* tables as non-DEFERRABLE.
 """
 
+import hashlib
+
 from django.db import migrations
+
+_PG_MAX_IDENTIFIER_LEN = 63
+
+
+def _safe_constraint_name(table_name, column_name, suffix='_fk_cascade'):
+    """
+    Return a constraint name that fits within PostgreSQL's 63-char identifier limit.
+
+    Uses the same truncate-and-hash strategy as field_types._safe_pg_identifier so
+    that long through-table names (e.g. custom_objects_14_technical_account_accounts)
+    combined with column names do not silently collide after PostgreSQL truncation.
+    """
+    full_name = f'{table_name}_{column_name}{suffix}'
+    if len(full_name) <= _PG_MAX_IDENTIFIER_LEN:
+        return full_name
+    digest = hashlib.md5(full_name.encode()).hexdigest()[:8]
+    prefix = full_name[:_PG_MAX_IDENTIFIER_LEN - 9].rstrip('_')
+    return f'{prefix}_{digest}'
 
 
 def fix_deferrable_fk_constraints(apps, schema_editor):
@@ -36,9 +56,17 @@ def fix_deferrable_fk_constraints(apps, schema_editor):
         rows = cursor.fetchall()
 
         for table_name, constraint_name, column_name, foreign_table in rows:
-            new_constraint_name = f'{table_name}_{column_name}_fk_cascade'
+            new_constraint_name = _safe_constraint_name(table_name, column_name)
+            # IF EXISTS on the old name handles partial re-runs where the old constraint
+            # was already dropped in a previous (failed) attempt.
             cursor.execute(
-                f'ALTER TABLE "{table_name}" DROP CONSTRAINT "{constraint_name}"'
+                f'ALTER TABLE "{table_name}" DROP CONSTRAINT IF EXISTS "{constraint_name}"'
+            )
+            # Pre-drop the new name too: if a previous run converted this constraint and
+            # then failed later (leaving the non-DEFERRABLE copy behind), the ADD below
+            # would collide. Dropping first makes every iteration idempotent.
+            cursor.execute(
+                f'ALTER TABLE "{table_name}" DROP CONSTRAINT IF EXISTS "{new_constraint_name}"'
             )
             cursor.execute(f"""
                 ALTER TABLE "{table_name}"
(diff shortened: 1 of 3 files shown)


bctiemann:
@claude /review

jnovinger (approved):

PR #515, opened 2026-05-13
Fixes: #396 - Require add permission for add URL, not change

### LINKED ISSUE #396: Permission Add does not work for adding Objects
### Plugin Version

0.4.6

### NetBox Version

v4.4.9

### Python Version

3.12.12

### Steps to Reproduce

Create A User give him Read And Add Permissions on a Custom object
Login as that user
try to Add some Objects, Error messages says permission denied

### Expected Behavior

That you can add and object but dont edit with the Add permission

### Observed Behavior

for adding the edit permissions is requrired

bctiemann:
## Summary

- `ObjectEditView.dispatch()` sets `_permission_action` based on whether `kwargs` is truthy. Custom object add URLs always include `custom_object_type` in kwargs, so `kwargs` is truthy even when adding — causing `change` permission to be required instead of `add`.
- Override `get_required_permission()` in `CustomObjectEditView` to check `self.kwargs.get('pk')`. By the time the permission check runs, `setup()` → `get_object()` has already popped `custom_object_type` from `self.kwargs`, leaving `pk` only for edits.
- Adds a regression test verifying that a user with only `add` permission can access the add URL (200) but is denied the edit URL (403).

## Test plan

- [ ] Run `test_add_permission_is_sufficient_to_access_add_url` in `CustomObjectViewTestCase`
- [ ] Manually assign only "Add" permission to a user in NetBox and confirm they can add custom objects without needing "Change"

Closes: #396

### DIFF
--- a/netbox_custom_objects/tests/test_views.py
+++ b/netbox_custom_objects/tests/test_views.py
@@ -371,6 +371,12 @@ class CustomObjectViewTestCase(CustomObjectsTestCase, ViewTestCases.PrimaryObjec
         edit_url = self._get_url('edit', self.instance1)
         self.assertHttpStatus(self.client.get(edit_url), 403)
 
+        # Symmetrical: change-only permission must not grant access to the add URL
+        obj_perm.actions = ['change']
+        obj_perm.save()
+        self.assertHttpStatus(self.client.get(add_url), 403)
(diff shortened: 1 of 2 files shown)


bctiemann:
@claude /review

arthanson (approved):

PR #516, opened 2026-05-14
Fixes: #380 - Set filterset on bulk edit/delete views so _all selects all

### LINKED ISSUE #380: Bulk Edit "Select all # objects matching query" only selects page size
### Plugin Version

0.4.5

### NetBox Version

4.4.5

### Python Version

Python 3.12.3

### Steps to Reproduce

Create netbox custom objects in a number greater than 50. 

Select all, then check the box to "Select all # objects matching query", edit selection. 

Image

Note selected objects is not "all objects matching query", but is limited 50. 

Image

### Expected Behavior

Editing all objects selected. In my case I expected to see 318 objects selected. 

### Observed Behavior

Selected objects is not "all objects matching query", but is limited 50. Must increase page size and manually select all to get each item you intend to edit.

bctiemann:
## Summary

- `BulkEditView.post()` and `BulkDeleteView.post()` check `self.filterset is not None` before applying the "select all matching query" (`_all`) flag. When `_all` is set without a filterset, both views fall through to `pk_list = request.POST.getlist('pk')` — only the PKs from the current page.
- `CustomObjectBulkEditView` and `CustomObjectBulkDeleteView` never set `self.filterset`, so "Select all X objects matching query" was silently limited to the current page size (default: 50), regardless of the actual query count.
- Fix: set `self.filterset = get_filterset_class(model)` in `setup()` on both views.

## Test plan

- [ ] `test_bulk_edit_select_all_respects_full_queryset` — verifies the view handles `_all` without 403/500 against a 62-object queryset
- [ ] `test_bulk_delete_select_all_respects_full_queryset` — creates 62 objects, submits `_all` with only 2 PKs in the form field, and asserts all 62 are deleted (not just the 2)
- [ ] Manual: create >50 custom objects, select all matching query, bulk edit/delete, confirm full set is processed

Closes: #380


bctiemann:
@claude /review

arthanson (approved):

arthanson (changes requested):
Looks good but has merge conflicts.

arthanson (approved):

PR #518, opened 2026-05-14
Fixes #517: Migration to rename mixed-case CustomObjectTypeField names

### LINKED ISSUE #517: Mixed-case field names create quoted PostgreSQL identifiers that break schema cloning
### Plugin Version

0.4.10

### NetBox Version

4.5.7

### Python Version

3.12

### Steps to Reproduce

Create a custom object type (e.g., ndfcfabric)
Add a multiobject field with a mixed-case name (e.g., Bordergateway)
Inspect the resulting database objects: SELECT tablename FROM pg_tables WHERE schemaname = 'public' AND tablename LIKE 'custom_objects_%';

### Expected Behavior

The plugin creates database tables, columns, indexes, and constraints using lowercase identifiers consistent with PostgreSQL/Django conventions (e.g., table custom_objects_10_bordergateway, column bordergateway), or rejects mixed-case names with a validation error.

### Observed Behavior

The plugin creates database objects with quoted mixed-case identifiers (e.g., table "custom_objects_10_Bordergateway", column "Bordergateway", index "custom_objects_10_Bordergateway_pkey"). This breaks any tooling that references these objects with unquoted identifiers, since PostgreSQL folds unquoted identifiers to lowercase. Additionally, renaming the column to lowercase via ALTER TABLE RENAME COLUMN does not propagate to the index relation's pg_attribute entries, requiring a superuser catalog update to fully resolve.

bctiemann:
## Background

Prior to v0.5.0, `CustomObjectTypeField.name` had no case restriction — operators could create fields named e.g. `Bordergateway` or `myField`. PostgreSQL stores table and column names in lowercase unless they are double-quoted; mixed-case identifiers therefore required quoting everywhere they were referenced. This worked for direct access but silently broke unquoted references such as those generated during schema cloning.

v0.5.0 added a `RegexValidator` enforcing `^[a-z0-9]+(_[a-z0-9]+)*$`, preventing new mixed-case names from being created. Databases that were populated on v0.4.x and upgraded to v0.5.0 may still carry the old mixed-case rows.

This PR closes the gap for those existing databases.

## Changes

**Migration `0014_fix_mixed_case_field_names`**

- On first run, queries all `CustomObjectTypeField` rows where `name != name.lower()`
- If none found: records as a no-op (clean installs and already-fixed databases are unaffected)
- If found: prints a report and prompts the operator interactively
  - Confirming with `y` runs the necessary DDL (`ALTER TABLE … RENAME COLUMN` / `RENAME TO`) then updates the DB rows; migration is recorded as applied
  - Answering anything else raises `RuntimeError` — the migration is **not** recorded as applied and can be re-run later
  - Non-interactive stdin (CI, scripted deploys) also raises `RuntimeError` rather than hanging or silently skipping

## Column/table handling per field type

| Field type | DB object renamed |
|---|---|
| Scalar | `{name}` column on main COT table |
| `object` (non-polymorphic) | `{name}_id` column on main COT table |
| `object` (polymorphic) | `{name}_content_type_id` + `{name}_object_id` columns |
| `multiobject` (either) | Through table `custom_objects_{cot_id}_{name}` (with `_safe_pg_identifier` applied) |

All DDL is guarded by `information_schema` existence checks so partial re-runs are safe.

## Test plan

- [ ] Run on a database with no mixed-case fields — migration applies silently as a no-op
- [ ] Run on a database with mixed-case fields — report is printed, prompt appears
  - Answer `N` — migration is not recorded; re-running shows the same prompt
  - Answer `y` — columns/tables renamed, field records updated, migration recorded as applied
- [ ] Run non-interactively (e.g. piped stdin) with mixed-case fields present — `RuntimeError` raised, migration not recorded

Closes: #517


bctiemann:
@claude /review and particularly evaluate the correctness of the SQL operations in the migration.

netbox_custom_objects/migrations/0014_fix_mixed_case_field_names.py
```diff
@@ -0,0 +1,162 @@
...
+                    cursor.execute(
+                        f'ALTER TABLE "{main_table}" RENAME COLUMN "{old_name}" TO "{new_name}"'
+                    )
+
+            CustomObjectTypeField.objects.filter(pk=field.pk).update(name=new_name)
+
+    print("Done. All mixed-case field names have been renamed to lowercase.")
+
+
+def noop(apps, schema_editor):
```
arthanson: why a custom hoop, there is a RunSQL.noop (https://docs.djangoproject.com/en/6.0/ref/migration-operations/#django.db.migrations.operations.RunSQL.noop)

netbox_custom_objects/migrations/0014_fix_mixed_case_field_names.py
```diff
@@ -0,0 +1,162 @@
...
+    print()
+
+    if not sys.stdin.isatty():
+        raise RuntimeError(
+            "Mixed-case CustomObjectTypeField names detected but stdin is not a TTY "
+            "(non-interactive run). Run this migration interactively to fix them. "
+            "The migration has NOT been recorded as applied."
+        )
+
+    answer = input("Rename all fields to lowercase now? [y/N] ").strip().lower()
```
arthanson: Having an input in a migration is fairly strange, would it work for the migration to just error as above and have this as a management command to run?  I would not expect a migration to be prompting.

arthanson (changes requested):
change the noop and a question re possibly moving the prompt out to a management command.

netbox_custom_objects/migrations/0014_fix_mixed_case_field_names.py
bctiemann: Done

netbox_custom_objects/migrations/0014_fix_mixed_case_field_names.py
bctiemann: Changed this to an inline rename, just with a clear `print` about what is being done and why so there is no operator surprise.

bctiemann:
@claude /review particularly the correctness of the SQL operations in the migration.

arthanson (approved):

Issue #519, opened 2026-05-15
constraint already exist


alfredhdahl:
### Plugin Version

0.5.0

### NetBox Version

4.5.10

### Python Version

3.12

### Steps to Reproduce

1. create a custom object with 0.4.9-plugin
2. upgrade plugin to 0.5.0



### Expected Behavior

no errors during upgrade

### Observed Behavior


migration 0011_non_deferrable_fk_constraints tries to create a constraint that already exists 


psycopg.errors.DuplicateObject: constraint "custom_objects_11_abc" for relation "custom_objects_11" already exists

jeremystretch:
Thank you for opening a bug report. Unfortunately, the information you have provided is not sufficient for someone else to attempt to reproduce the reported behavior. Remember, each bug report must include detailed steps that someone else can follow on a clean, empty NetBox installation to reproduce the exact problem you're experiencing. These instructions should include the creation of any involved objects, any configuration changes, and complete accounting of the actions being taken. Also be sure that your report does not reference data on the public NetBox demo, as that is subject to change at any time by an outside party and cannot be relied upon for bug reports.

bctiemann:
Thanks for the report. To investigate further, could you provide:

1. The full Python traceback (not just the last line) — specifically which `cursor.execute()` call in migration 0011 raises `DuplicateObject`
2. Your PostgreSQL `search_path` setting and whether you are using a non-public schema
3. The name of the field/COT involved (the constraint name `custom_objects_11_abc` suggests COT #11 with a field named `abc` — is that accurate, or was that simplified?)
4. Whether migration 0011 had previously been applied (i.e., was this a fresh 0.4.9→0.5.0 upgrade, or had 0.5.0 been partially applied before?)

In the meantime, we have identified one probable cause: migration 0011's query against `information_schema.table_constraints` is missing a `table_schema = current_schema()` filter. On a PostgreSQL installation where `search_path` includes multiple schemas, the query can return DEFERRABLE constraints from a schema other than the active one. The subsequent `DROP CONSTRAINT IF EXISTS` then operates against the wrong schema (silently no-ops), while `ADD CONSTRAINT` targets the correct table where a constraint with that name already exists — producing the `DuplicateObject` error. We will apply that fix regardless.

alfredhdahl:
sorry for not suppliying sufficient data;

snippet from my claude dialogue ( answers 1-4 below)(Claude already looked into the other issue you mentioned):


[...]

PR #520, opened 2026-05-15
Add table_schema filter to migration 0011 query

bctiemann:
## Summary

Migration 0011's `information_schema.table_constraints` query was missing `AND tc.table_schema = current_schema()`. Without this filter, the query returns DEFERRABLE constraints from **all schemas** accessible to the current PostgreSQL user, not just the current one.

On installations where the NetBox database uses a non-public schema (or where `search_path` exposes multiple schemas), this causes:

1. The query returns a DEFERRABLE constraint found in a **different** schema
2. `DROP CONSTRAINT IF EXISTS` operates on the **current** schema's table — silently no-ops because the constraint isn't there
3. `ADD CONSTRAINT` tries to add to the current schema's table where a constraint with that name **already exists** → `psycopg.errors.DuplicateObject`

The fix is a one-line addition matching the pattern already used elsewhere in the codebase (e.g. `_ensure_field_fk_constraint` in `models.py`).

## Test plan

- [ ] Upgrade from 0.4.9 to 0.5.x on a standard public-schema PostgreSQL setup — migration applies cleanly (no regression)
- [ ] Upgrade on a non-public-schema setup — migration now correctly scopes the query to the current schema and applies without `DuplicateObject`

### DIFF
--- a/netbox_custom_objects/migrations/0011_non_deferrable_fk_constraints.py
+++ b/netbox_custom_objects/migrations/0011_non_deferrable_fk_constraints.py
@@ -54,6 +54,7 @@ def fix_deferrable_fk_constraints(apps, schema_editor):
                 ON tc.constraint_name = rc.constraint_name
                 AND tc.table_schema = rc.constraint_schema
             WHERE tc.constraint_type = 'FOREIGN KEY'
+                AND tc.table_schema = current_schema()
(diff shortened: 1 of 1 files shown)


bctiemann:
Possibly fixes #519 -- follow up

arthanson (approved):

PR #524, opened 2026-05-20
Fixes: #523 - Delete NetBox CustomField references before dropping ObjectType

### LINKED ISSUE #523: Deleting a CustomObjectType raises ProtectedError + NoReverseMatch when a NetBox CustomField references the COT's ObjectType
## Summary

Deleting a `CustomObjectType` crashes with a 500 error when any NetBox `CustomField` (extras) has `related_object_type` pointing to the COT's underlying `ObjectType` (i.e., a custom field of type _Object_ or _Multi-object_ that maps to this COT).

## Steps to Reproduce

1. Create a Custom Object Type (e.g., \"Cat\").
2. On any NetBox model, create a Custom Field of type **Object** and set its _Related object type_ to your COT.
3. Navigate to the COT list and delete the COT.
4. Confirm the deletion.

## Expected Behavior

The COT and all dependent data are deleted cleanly. Any NetBox `CustomField` that referenced the COT's ObjectType is also removed (or the user is warned).

## Actual Behavior

```
django.db.models.deletion.ProtectedError: ("Cannot delete some instances of model 'ObjectType'
because they are referenced through protected foreign keys: 'CustomField.related_object_type'.",
{<CustomField: Cat>})
```

After the `ProtectedError` is raised, a second error is logged:

```
django.urls.exceptions.NoReverseMatch: Reverse for 'customobjecttype' with arguments '(None,)' not found.
```

This second error occurs because `super().delete()` is called (line ~979 in `models.py`) **before** `object_type.delete()` (line ~984). After `super().delete()`, `self.pk` is `None`. The exception propagates to NetBox's generic delete view, which tries `redirect(obj.get_absolute_url())` → `reverse(..., args=[None])`.

## Root Cause

`CustomObjectType.delete()` already cleans up `CustomObjectTypeField` records that reference the COT's `ObjectType`, but it does **not** clean up regular NetBox `CustomField` records (`extras.models.CustomField`) that may have `related_object_type` pointing to the same `ObjectType`. `CustomField.related_object_type` uses `on_delete=PROTECT`, so the subsequent `object_type.delete()` call is blocked.

## Fix

Before calling `object_type.delete()`, also delete any `CustomField` records that reference this `ObjectType`:

```python
from extras.models import CustomField
CustomField.objects.filter(related_object_type=object_type).delete()
```

This mirrors the existing logic that deletes `CustomObjectTypeField` records.

bctiemann:
## Summary

- Deleting a CustomObjectType that has a NetBox CustomField (extras) with related_object_type pointing at the COT ObjectType now succeeds cleanly
- CustomObjectType.delete() already cleaned up CustomObjectTypeField records; this PR extends it to also delete matching extras.CustomField rows before calling object_type.delete()
- Adds a regression test covering this exact scenario

## Root Cause

CustomField.related_object_type is declared with on_delete=PROTECT. When CustomObjectType.delete() called object_type.delete(), Django refused because at least one CustomField row still referenced that ObjectType. The plugin already handled its own CustomObjectTypeField foreign keys but missed the standard NetBox custom field.

A secondary symptom (NoReverseMatch with pk=None) occurred because super().delete() runs before object_type.delete() — once the COT row is gone, self.pk is None, so the error handler get_absolute_url() call failed. This is resolved implicitly since the ProtectedError no longer fires.

## Test Plan

- [x] New test test_delete_cot_with_netbox_custom_field_referencing_object_type in test_schema_operations.py creates a COT, attaches a CustomField with related_object_type set to the COT ObjectType, deletes the COT, and asserts both the COT and the CustomField are gone

Closes: #523

### DIFF
--- a/netbox_custom_objects/models.py
+++ b/netbox_custom_objects/models.py
@@ -31,6 +31,7 @@ from extras.choices import (
     CustomFieldUIEditableChoices,
     CustomFieldUIVisibleChoices,
 )
+from extras.models import CustomField
 from extras.models.customfields import SEARCH_TYPES
 from extras.utils import run_validators
 from netbox.config import get_config
@@ -976,6 +977,12 @@ class CustomObjectType(NetBoxModel):
 
         object_type = ObjectType.objects.get_for_model(model)
         ObjectChange.objects.filter(changed_object_type=object_type).delete()
+
+        # Delete any NetBox CustomField records (extras) with related_object_type pointing
+        # to this COT's ObjectType. CustomField.related_object_type uses on_delete=PROTECT,
+        # so these must be removed before object_type.delete() is called below.
+        CustomField.objects.filter(related_object_type=object_type).delete()
+
         super().delete(*args, **kwargs)
 
         # Temporarily disconnect the pre_delete handler to skip the ObjectType deletion
--- a/netbox_custom_objects/tests/test_schema_operations.py
+++ b/netbox_custom_objects/tests/test_schema_operations.py
@@ -154,6 +154,33 @@ class SchemaOperationsTestCase(TransactionCleanupMixin, CustomObjectsTestCase, T
             "Deleted COT's model must be removed from the app registry.",
         )
 
+    def test_delete_cot_with_netbox_custom_field_referencing_object_type(self):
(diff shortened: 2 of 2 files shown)


bctiemann:
@claude /review

arthanson (approved):

PR #526, opened 2026-05-21
Fixes: #491 - Extend cycle depth-first-search to traverse polymorphic field edges

### LINKED ISSUE #491: Multi-hop polymorphic cycle detection gap
## Summary

The DFS cycle check in `CustomObjectTypeField.clean()` only traverses non-polymorphic FK edges. A cycle that passes entirely through polymorphic legs goes undetected at field creation time.

## Example

Given three COTs:

- **A** has a polymorphic object field pointing at `[B]`
- **B** has a polymorphic object field pointing at `[C]`
- **C** has a polymorphic object field pointing at `[A]`

Creating C's field is accepted without error, but A → B → C → A is a reference cycle.

## Current behaviour

The `check_polymorphic_recursion` m2m_changed signal guards only the direct A→B assignment. The DFS in `clean()` explicitly skips polymorphic fields with a KNOWN LIMITATION comment in models.py.

## Fix

Extend the cycle-detection DFS to walk `field.related_object_types.filter(app_label=APP_LABEL)` for polymorphic fields, recursing into each reachable COT the same way it recurses on non-polymorphic FK fields. Extend `check_polymorphic_recursion` similarly so multi-hop cycles are caught both at field creation and when the M2M is mutated.

## Related

Introduced alongside polymorphic field support in PR #442.

bctiemann:
## Summary

- Extends `_has_circular_reference` to walk polymorphic field edges, so cycles that pass entirely through polymorphic legs are now detected at M2M assignment time (via the existing `check_polymorphic_recursion` signal)
- Removes the KNOWN LIMITATION comment block from `models.py`
- Replaces `PolymorphicCycleDetectionGapTest` (which pinned the broken behaviour) with `PolymorphicCycleDetectionTest` covering the 2-hop (A→B→A) and 3-hop (A→B→C→A) cases, plus a guard that the first non-cyclic edge is still accepted

## How it works

`_has_circular_reference` previously only followed non-polymorphic `related_object_type` FK edges. Polymorphic fields store their allowed types on the `related_object_types` M2M, so any cycle passing entirely through those was invisible to the DFS.

The fix adds a second traversal loop after the existing one: for each `is_polymorphic=True` field on the current COT, it iterates `related_object_types.filter(app_label=APP_LABEL)` and recurses into each reachable COT. The shared `related_objects_checked` set deduplicates across both loops.

Because `check_polymorphic_recursion` already delegates to `_has_circular_reference`, the signal guard inherits the fix automatically — no changes to the signal itself are needed.

Closes: #491

## Test plan

- [x] `test_first_poly_edge_is_accepted` — A→B with no back-edge is allowed
- [x] `test_two_hop_polymorphic_cycle_is_detected` — A→B→A raises `ValidationError`
- [x] `test_three_hop_polymorphic_cycle_is_detected` — A→B→C→A raises `ValidationError`
- [x] Full `test_polymorphic_fields` module (47 tests) passes

### DIFF
--- a/netbox_custom_objects/models.py
+++ b/netbox_custom_objects/models.py
@@ -1734,6 +1734,7 @@ class CustomObjectTypeField(CloningMixin, ExportTemplatesMixin, ChangeLoggedMode
                 CustomFieldTypeChoices.TYPE_OBJECT,
                 CustomFieldTypeChoices.TYPE_MULTIOBJECT,
             ],
+            is_polymorphic=False,
             related_object_type__isnull=False,
             related_object_type__app_label=APP_LABEL
         ):
--- a/netbox_custom_objects/models.py
+++ b/netbox_custom_objects/models.py
@@ -1725,22 +1725,10 @@ class CustomObjectTypeField(CloningMixin, ExportTemplatesMixin, ChangeLoggedMode
         # Add this type to visited set
         visited.add(custom_object_type.id)
 
-        # Check all *non-polymorphic* object and multiobject fields in this COT.
-        #
-        # KNOWN LIMITATION: polymorphic fields (is_polymorphic=True) store their
-        # allowed target types on the related_object_types M2M, not on the
-        # related_object_type FK.  This DFS therefore does not traverse edges
-        # introduced by polymorphic fields.  A cycle that passes entirely through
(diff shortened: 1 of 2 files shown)


bctiemann:
@claude /review

arthanson (approved):

PR #527, opened 2026-05-26
Fixes: #559 - Add slug filter to CustomObjectTypeFilterSet

### LINKED ISSUE #559: Custom Object Type API ignores the `slug` filter and returns every type
### Plugin Version

0.5.1

### NetBox Version

4.6.1

### Python Version

3.14

### Steps to Reproduce

1. Create two or more Custom Object Types with distinct slugs (e.g. "vendor-policies" and "device-roles").
2. Query the REST API filtering by one slug:
   `GET /api/plugins/custom-objects/custom-object-types/?slug=vendor-policies`
3. Inspect the `count` / `results` in the response.

### Expected Behavior

The response should contain only the Custom Object Type whose slug is `vendor-policies` (and the related lookups `?slug__ic=`, `?slug__isw=`, etc. should narrow the result set accordingly).

### Observed Behavior

The `slug` query parameter is silently ignored — the endpoint returns **every** Custom Object Type instead of filtering.

Root cause: `CustomObjectTypeFilterSet` (`netbox_custom_objects/filtersets.py:259`) does not list `slug` in `Meta.fields`, so `django-filter` never generates a filter for it and drops the unrecognised query parameter without error. Adding `slug` to the tuple lets `django-filter` auto-generate the full lookup family (`exact`, `__ic`, `__isw`, `__regex`, ...) backed by the `CustomObjectType.slug` SlugField.

A proposed fix already exists in PR #527 (closed pending this bug report): add `"slug"` to `CustomObjectTypeFilterSet.Meta.fields`. Originally surfaced during the related-object-tabs smoke test (PR #482).

Kani999:
## Summary

`GET /api/plugins/custom-objects/custom-object-types/?slug=foo` previously returned **every** Custom Object Type — `django-filter` silently dropped the unrecognised query parameter because `CustomObjectTypeFilterSet.Meta.fields` didn't include `slug`. Adding `slug` to the tuple lets `django-filter` auto-generate the full lookup family (`exact`, `__ic`, `__isw`, `__regex`, ...) backed by the `CustomObjectType.slug` SlugField.

## Verification

Verified via `manage.py shell`:

```python
from netbox_custom_objects.filtersets import CustomObjectTypeFilterSet
'slug' in CustomObjectTypeFilterSet.base_filters
# True
```

After the change, all of these now narrow the result set:

- `?slug=vendor-policies`
- `?slug__ic=policies` (icontains)
- `?slug__isw=vendor` (istartswith)

## Context

Surfaced during the related-object-tabs smoke test on 2026-05-26 (PR #482).

### DIFF
--- a/netbox_custom_objects/filtersets.py
+++ b/netbox_custom_objects/filtersets.py
@@ -261,6 +261,7 @@ class CustomObjectTypeFilterSet(NetBoxModelFilterSet):
         fields = (
             "id",
             "name",
+            "slug",
             "group_name",
         )
 


bctiemann:
Hi @Kani999 -- please don't open a PR unless there is an accepted bug or feature issue to link it to.

Please open a bug ticket first, then once it's triaged and accepted you can reopen this PR. Thanks!

bctiemann (approved):

PR #528, opened 2026-05-26
Fixes: #558 - Avoid "<type> None" in CustomObject.__str__ post-delete

### LINKED ISSUE #558: CustomObject.__str__ renders "<type> None" after deletion when the type has no primary field
### Plugin Version

0.5.1

### NetBox Version

4.6.1

### Python Version

3.14

### Steps to Reproduce

1. Create a Custom Object Type that has NO field marked primary=True (e.g. "SmokeTabTarget").
2. Create at least one custom object instance of that type.
3. Delete the instance via the UI (or call instance.delete() in a manage.py shell).
4. Observe the success toast and the changelog entry rendered after deletion.

### Expected Behavior

The delete confirmation toast and changelog entry should display the object's type label cleanly — never a literal "None".

### Observed Behavior

The rendered string contains a trailing "None", e.g.:

    Deleted SmokeTabTarget SmokeTabTarget None

Root cause: `CustomObject.__str__` (`netbox_custom_objects/models.py`, the `if not primary_field_value:` fallback) returns `f"{self.custom_object_type.display_name} {self.id}"`. Django's `Model.delete()` resets `self.id` / `self.pk` to `None` after the DB row is removed, and NetBox's delete-toast and changelog renderers call `str()` on the post-delete instance, so `{self.id}` interpolates as `None`. Only affects types with no `primary=True` field.

A proposed fix already exists in PR #528 (closed pending this bug report): branch on `self.pk is None` and return just `self.custom_object_type.display_name` in that case. Originally found during smoke testing of the related-tabs feature (PR #482).

Kani999:
## Summary

`CustomObject.__str__` includes `self.id` in its fallback output when no field on the COT is marked `primary=True`. But Django's `Model.delete()` sets `self.pk` to `None` after the database row is removed — and NetBox's delete-toast and changelog renderers call `str()` on the **post-delete** instance. The result is a literal "None" in the user-facing message:

```
Deleted SmokeTabTarget SmokeTabTarget None
```

## Fix

Branch on `self.pk` inside the no-primary-field path:

- **`self.pk is None`** (post-delete) → return just the COT's display name (`"SmokeTabTarget"`).
- **`self.pk` is set** (normal rendering) → behave as before (`"SmokeTabTarget 5"`).

8 lines, no behavioural change for any code path with a live `pk`.

## Verification

Observed while smoke-testing PR #482's related-tabs work.

## Context

Surfaced during the related-object-tabs smoke testing (PR #482); independent of that work and easy enough to land standalone.

### DIFF
--- a/netbox_custom_objects/models.py
+++ b/netbox_custom_objects/models.py
@@ -135,6 +135,8 @@ class CustomObject(
                 self, primary_field["name"]
             )
         if not primary_field_value:
+            if self.pk is None:  # post-delete: pk is cleared by Django's Model.delete()
+                return str(self.custom_object_type.display_name)
             return f"{self.custom_object_type.display_name} {self.id}"
         return str(primary_field_value) or str(self.id)
 
--- a/netbox_custom_objects/tests/test_field_types.py
+++ b/netbox_custom_objects/tests/test_field_types.py
@@ -1113,6 +1113,23 @@ class PrimaryFieldChangeTestCase(FieldTypeTestCase):
         refreshed_target = new_target_model.objects.get(pk=target_instance.pk)
         self.assertIn("T-001", str(refreshed_target))
 
+    def test_str_post_delete_no_primary_field_does_not_contain_none(self):
+        """Regression #558: str() on a post-delete instance (pk=None, no primary
+        field value) must return the COT display name, not '<type> None'."""
+        cot = self.create_custom_object_type(
+            name="DeleteStrTest", slug="delete-str-test",
+            verbose_name_plural="Delete Str Tests",
+        )
+        # No primary field — falls through to the pk fallback path in __str__.
+        model = cot.get_model()
+        instance = model.objects.create()
+
+        instance.delete()  # sets instance.pk = None
+
+        result = str(instance)
+        self.assertNotIn("None", result)
+        self.assertEqual(result, cot.display_name)
+
 
 # ---------------------------------------------------------------------------
 # Context field — ts-parent-field widget attribute on get_form_field()


bctiemann:
Hi @Kani999 , as with the other PR just now, please don't open a PR unless there is an accepted bug to link it to (this is for later bookkeeping/forensics). Please open a bug issue with repro steps and any relevant details according to the template, and then once it's triaged and accepted we can reopen this PR. Thanks!

bctiemann:
Two small suggestions:

**1. Trim the comment** (project convention is one short line max):

```python
if self.pk is None:  # post-delete: pk is cleared by Django's Model.delete()
    return str(self.custom_object_type.display_name)
```

**2. Regression test** in `PrimaryFieldChangeTestCase` in `tests/test_field_types.py`:

```python
def test_str_post_delete_no_primary_field_does_not_contain_none(self):
    """Regression #558: str() on a post-delete instance (pk=None, no primary
    field value) must return the COT display name, not '<type> None'."""
    cot = self.create_custom_object_type(
        name="DeleteStrTest", slug="delete-str-test",
        verbose_name_plural="Delete Str Tests",
    )
    # No primary field — falls through to the pk fallback path in __str__.
    model = cot.get_model()
    instance = model.objects.create()

    instance.delete()  # sets instance.pk = None

    result = str(instance)
    self.assertNotIn("None", result)
    self.assertEqual(result, cot.display_name)
```

Both verified locally — test passes with the fix and fails on the unpatched code.

bctiemann (changes requested):
See above

bctiemann (approved):

PR #530, opened 2026-05-26
Fixes: #384 - Patch stale FK refs and fix false-positive cycle detection

### LINKED ISSUE #384: Model generation problem: fields disappear when Custom Object Types reference each other (NetBox 4.5 + Unit multi-process)
### Plugin Version

0.4.6

### NetBox Version

NetBox Community v4.5.0 (Docker 3.4.2)

### Python Version

3.12

### Steps to Reproduce

**Custom Object Types schema** 
_Custom Object Type A_

Name: vpc_domain
(dynamic model: Table13Model)

Fields:
peer_1
type: object
related object type: dcim.device

_Custom Object Type B_
Name: vpc
(dynamic model: Table14Model)

Fields:
interface
type: object
related object type: dcim.interface

vpc_domain
type: object
related object type: Custom Object Type A (vpc_domain)

1) Use NGINX Unit with more than one application process, default options in netbox docker installation.
2) Create the two Custom Object Types described above.
3) Inspect generated dynamic models from different processes or after restarts

### Expected Behavior

Model generation should be the same 

### Observed Behavior

The generated model for vpc_domain (Table13Model) is inconsistent:

Sometimes:
['id', 'created', 'last_updated', 'peer_1']

Sometimes:
['id', 'created', 'last_updated']

The field peer_1 disappears even though it exists in CustomObjectTypeField in the database, no schema changes were made, the same process is used (with processes=1, a “bad” initialization can be locked in)

This causes runtime errors such as "AttributeError: 'Table13Model' object has no attribute 'peer_1'"

bctiemann:
## Summary

Two bugs that together produce "fields disappear / Cannot assign TableNModel instance" failures in multi-worker deployments:

### Bug 1: False-positive cycle detection in `_has_circular_reference`

The DFS seeds `visited` with the origin COT's ID, then treats **any** re-encounter of a visited node as a genuine cycle. When an intermediate COT (e.g. **Control**) has a self-referencing multiobject field, the DFS re-encounters Control's ID in `visited` and returns `True` — falsely blocking a valid **EC → Control** FK with "Circular reference detected".

**Fix:** `return custom_object_type.id == self.custom_object_type.id` — only a genuine cycle when the DFS traces back to the **origin** COT.

### Bug 2: Stale FK reference after model regeneration

When a COT's model is regenerated on a cache miss (e.g. `cache_timestamp` bumped after a schema change on another worker), the existing patch loop updated inbound M:N through-model FKs (`TYPE_MULTIOBJECT`) but not inbound **direct FK fields** (`TYPE_OBJECT`). Any cached model that held a `LazyForeignKey` to the regenerated COT then carried a stale class reference, causing Django's isinstance check in `ForeignKey.__set__` to raise:

```
ValueError: Cannot assign "<TableNModel: ...>": "TableMModel.ref_field"
must be a "TableNModel" instance.
```
(Both names are identical — the two objects are different Python classes from different `get_model()` calls.)

**Fix:** After the existing through-model patch loop, add an analogous loop over inbound `TYPE_OBJECT` fields. For each, find the FK field in the owner model's `local_fields` list (uses `local_fields` to avoid `_relation_tree` recursion) and patch `remote_field.model`, `related_model`, and the cached `path_infos`.

## Test plan

- [ ] `CycleDetectionFalsePositiveRegressionTest.test_fk_into_self_referencing_cot_is_not_false_positive` — a valid EC→Control FK (Control has self-referencing M:N) no longer raises ValidationError
- [ ] `CycleDetectionFalsePositiveRegressionTest.test_genuine_cycle_is_still_detected` — a genuine A→B→A cycle still raises ValidationError
- [ ] `StaleFKReferenceRegressionTest.test_b_fk_resolves_to_a_v1_before_regeneration` — sanity: B's FK initially points to A's model class
- [ ] `StaleFKReferenceRegressionTest.test_b_fk_is_patched_after_a_regeneration` — after A is regenerated, B's FK field references the new class

### DIFF
--- a/netbox_custom_objects/models.py
+++ b/netbox_custom_objects/models.py
@@ -816,6 +816,30 @@ class CustomObjectType(NetBoxModel):
                 target_field.__dict__.pop('path_infos', None)
                 target_field.__dict__.pop('reverse_path_infos', None)
 
+            # Same staleness problem exists for direct FK fields (TYPE_OBJECT):
+            # when this COT is regenerated, any cached model for another COT that
+            # holds a LazyForeignKey pointing here still references the old class.
+            # Walk inbound non-polymorphic object fields and patch them too.
+            for inbound_fk_field in CustomObjectTypeField.objects.filter(
+                related_object_type=self.object_type,
+                type=CustomFieldTypeChoices.TYPE_OBJECT,
+                is_polymorphic=False,
+            ).iterator():
+                owner_model = CustomObjectType.get_cached_model(inbound_fk_field.custom_object_type_id)
+                if owner_model is None:
+                    continue
+                # Use local_fields list — avoids _relation_tree → get_models() recursion.
+                fk_field = next(
+                    (f for f in owner_model._meta.local_fields if f.name == inbound_fk_field.name),
+                    None,
+                )
+                if fk_field is None:
+                    continue
+                fk_field.remote_field.model = model
+                fk_field.related_model = model
+                fk_field.__dict__.pop('path_infos', None)
+                fk_field.__dict__.pop('reverse_path_infos', None)
+
             # Only cache fully-generated models.  Models generated with
             # skip_object_fields=True omit FK fields to other COTs; caching them
             # would permanently hide those fields if a dependent COT triggers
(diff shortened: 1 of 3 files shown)


bctiemann:
@claude /review

arthanson (approved):

Issue #461, opened 2026-04-09
Choice field shows raw value instead of label in detail view


FrantisekAdastra:
### Plugin Version

0.4.9

### NetBox Version

4.5.7

### Python Version

3.12.3

### Steps to Reproduce

1. Go to Customization → Custom Field Choices and create a Choice Set 
   where Value and Label differ (e.g. value: "3rd_party_open_source", 
   label: "Self-hosted 3rd Party Open Source SW")
2. Create a Custom Object Type with a field of type "Selection" using 
   this Choice Set
3. Create a Custom Object instance and select one of the choices
4. Save and view the object detail page

### Expected Behavior

The detail view should display the human-readable Label 
(e.g. "Self-hosted 3rd Party Open Source SW")

### Observed Behavior

The detail view displays the raw Value from the database (e.g. "3rd_party_open_source") instead of the Label.
The edit form correctly shows the Label in the dropdown, so the issue is only in the read/detail view rendering.

<img width="2536" height="1363" alt="Image" src="https://github.com/user-attachments/assets/070b9e4e-e281-430c-9e34-2fe528aaa76f" />
<img width="2412" height="1370" alt="Image" src="https://github.com/user-attachments/assets/a9bcbb8d-31a8-4363-a6dc-8b56e1ca7991" />
<img width="2553" height="1369" alt="Image" src="https://github.com/user-attachments/assets/a5785fba-abff-4aeb-9082-302e419c24c8" />

bctiemann:
May be a duplicate of #369

PR #531, opened 2026-05-26
Fixes: #369 - Show choice label instead of raw key for Selection fields

### LINKED ISSUE #369: Choice Set (Selection) appears as Key instead of Value in some areas
### Plugin Version

0.4.5

### NetBox Version

v4.5.0

### Python Version

3.12

### Steps to Reproduce

1. Create a Custom Object Type with a Selection field (using a Choice Set)

2. Set the Selection field as the primary field (primary name field checkbox)

3. Create a choice set with entries where the key differs from the value (e.g., key: "myvalue", value: "My Value")

4. Create one or more custom objects with this Selection field

5. View the custom objects in different contexts:

   - In tables/lists
   - In the object detail view title
   - In the "Custom Objects linking to this object" panel on related objects

### Expected Behavior

The Selection field should consistently display the human-readable __value__ (e.g., "My Value") from the choice set across all UI contexts, including:

- Primary field display (object title)
- Table rows
- Link text in related objects panels
- Any string representation of the object

### Observed Behavior

The Selection field displays inconsistently:

- Sometimes shows the __value__ (e.g., "My Value")
- 

Image

- Other times shows the __key__ (e.g., "myvalue")

Image

The inconsistency appears to depend on the context where the field is being rendered. The root cause seems to be that `SelectFieldType.get_display_value()` in `field_types.py` uses the default implementation which returns the raw database value (the key) instead of using Django's `get_FOO_display()` method to retrieve the human-readable label.

bctiemann:
## Summary

- `SelectFieldType.get_display_value()` now calls Django's `get_<fieldname>_display()` so the object title, `__str__`, and serializer context return the human-readable label instead of the raw key
- `SelectFieldType.get_table_column_field()` returns a column whose `render()` translates keys via a choices-dict closure, fixing table rows
- `MultiSelectFieldType.get_display_value()` translates each key to its label by reading choices from the model field's `base_field`
- `MultiSelectFieldType.get_table_column_field()` replaces the old `render_table_column` that returned raw keys

Issues #369 and #461 are the same root cause (both in `SelectFieldType` lacking `get_display_value()`/table rendering), covered by a single fix.

## Test plan

- [ ] `SelectFieldTypeTestCase.test_select_field_display_value_returns_label` — asserts `get_display_value()` returns `"Choice 1"` not `"choice1"`
- [ ] `SelectFieldTypeTestCase.test_select_primary_field_str_uses_label` — asserts `str(instance)` returns `"Choice 2"` when a Selection field is primary
- [ ] `MultiSelectFieldTypeTestCase.test_multiselect_field_display_value_returns_labels` — asserts joined labels `"Choice 1, Choice 3"` not raw keys
- [ ] Full test suite: `python manage.py test netbox_custom_objects.tests --keepdb`

### DIFF
--- a/netbox_custom_objects/field_types.py
+++ b/netbox_custom_objects/field_types.py
@@ -556,12 +556,7 @@ class MultiSelectFieldType(FieldType):
         values = getattr(instance, field_name) or []
         if not values:
             return ''
-        choices_dict = dict(
-            next(
-                (f.base_field.choices for f in instance._meta.local_fields if f.name == field_name),
-                [],
-            )
-        )
+        choices_dict = dict(instance._meta.get_field(field_name).base_field.choices)
         return ', '.join(choices_dict.get(v, v) for v in values)
 
     def get_model_field(self, field, **kwargs):
--- a/netbox_custom_objects/field_types.py
+++ b/netbox_custom_objects/field_types.py
@@ -472,6 +472,21 @@ class JSONFieldType(FieldType):
 
 
 class SelectFieldType(FieldType):
+    def get_display_value(self, instance, field_name):
+        value = getattr(instance, field_name)
+        if value is None:
+            return ''
+        return getattr(instance, f'get_{field_name}_display')() or value
+
+    def get_table_column_field(self, field, **kwargs):
+        choices_dict = dict(field.choices)
+
+        class _SelectLabelColumn(tables.Column):
+            def render(self, value):
+                return choices_dict.get(value, value)
+
+        return _SelectLabelColumn()
+
     def get_filterform_field(self, field, **kwargs):
         return DynamicMultipleChoiceField(
             choices=field.choice_set.choices,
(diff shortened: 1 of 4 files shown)


bctiemann:
@claude /review and limit reporting only to areas of potential concern.

bctiemann:
Addressed findings 1 and 4.

arthanson (changes requested):
This doesn't fix it from what I was seeing, in the detail view it is still showing as the key, not the display value.  List view works fine, but it was that way previously.  

Image
Image

bctiemann:
## What the original fix covered — and what it missed

The original commit (`97950db`) added `get_display_value()` and `get_table_column_field()` to `SelectFieldType` and `MultiSelectFieldType`. Those two methods cover these rendering paths:

| Rendering path | Method invoked | Fixed by original commit? |
|---|---|---|
| `CustomObject.__str__()` / page title | `get_display_value()` | ✅ Yes |
| REST API serializer display value | `get_display_value()` | ✅ Yes |
| List-view table column | `get_table_column_field()` column `render()` | ✅ Yes |
| Detail-page field row | `get_field_value` template filter | ❌ No |

The detail page uses a separate code path that was never touched by either of those methods. The template tag `get_field_value` (in `templatetags/custom_object_utils.py`) calls `getattr(obj, field.name)` directly and passes the raw stored code straight to `builtins/customfield_value.html`, which renders it verbatim — bypassing `get_display_value` entirely.

Today's follow-up commit teaches `get_field_value` to translate the raw code to its human-readable label before the template sees it, completing the fix.

arthanson (changes requested):
I think we just want to display the human readable name, right now it is showing both the key and the human readable name.

Image

bctiemann:
Hmm. I'm not sure we're seeing the same thing.

My ChoiceSet looks like this:
```
first:First Choice
second:Second Choice
third:Third Choice
```

And with single- and multi-select fields `value` and `values`, I'm seeing:

Screenshot 2026-05-28 at 7 36 40 PM
Screenshot 2026-05-28 at 7 36 47 PM

That's right, isn't it? How do our setups differ?

arthanson (approved):

Issue #508, opened 2026-05-12
ValueError "Must be 'TableNModel' instance" in CustomObjectLink.left_page() due to no_cache=True breaking dynamic model identity


FrantisekAdastra:
### Plugin Version

0.5.0

### NetBox Version

4.5.7

### Python Version

3.12.3

### Steps to Reproduce

1. Create a Custom Object Type (e.g. "Platform Service").
2. Add an Object-type field on it whose related type is the same Custom Object Type (self-referential field).
3. Create at least one instance of this Custom Object Type.
4. Open the detail view of any instance: /plugins/custom-objects/<plural>/<id>/

### Expected Behavior

The detail view renders, including the "Custom Objects linking to this object" panel.

<img width="1437" height="691" alt="Image" src="https://github.com/user-attachments/assets/88a70c2c-e200-4819-898f-92389ea2e340" />

### Observed Behavior

Detail view returns HTTP 500. Traceback:

```
ValueError at /plugins/custom-objects/platform-services/7/
Cannot query "Mobile": Must be "Table7Model" instance.

Template error:
In template .../netbox_custom_objects/templates/netbox_custom_objects/customobject.html, line 137
   {% plugin_left_page object %}

File ".../netbox_custom_objects/template_content.py", line 82, in left_page
    linked_objects = model.objects.filter(**{field.name: target_obj})
File ".../django/db/models/sql/query.py", line 1351, in check_query_object_type
    raise ValueError(
```

The string `"Mobile"` in the error message is `str(target_obj)` — i.e. the very object whose detail page is being rendered. `target_obj` is in fact an instance of a class named `Table7Model`, but not the same class object as `model`.

## Root cause

In `template_content.py` line 59, `CustomObjectLink.left_page()` calls:

```python
model = field.custom_object_type.get_model(no_cache=True)
```

`no_cache=True` causes `get_model()` to generate a fresh Python class on every call. For a self-referential field, the iteration ends up comparing `target_obj` (whose class came from an earlier, cached call to `get_model()` elsewhere in the request) against a newly minted `Table7Model` class. Both classes represent the same DB table and have the same name, but they are not identity-equal Python objects.

Django's `check_query_object_type()` in `db/models/sql/query.py` uses an `isinstance()` check that consequently fails, and raises `ValueError` formatted with `str(value)` — which is why the error appears to mention a field value (`"Mobile"`) rather than the type mismatch directly.

[...]

PR #533, opened 2026-05-28
Fixes: #353 - Prevent AttributeError in __str__ when primary field missing from generated model

### LINKED ISSUE #353: Loading Built-in Model Pages Fails with AttributeError
### Plugin Version

0.4.4

### NetBox Version

4.4.8

### Python Version

3.10

### Steps to Reproduce

Unfortunately, I can't replicate the behavior when creating new custom objects, but I'm experiencing behavior very similar to #99. It seems to affect two specific custom object types that I created a while back, likely when the plugin went GA. The steps I would've followed at that time: 

- Create a custom object type (COT)
- Create a custom object type field (COTF) of type object
  - Select a model as the related object type (currently affecting IP Address, IP Range, Prefix)
- Create an instance of the COT without specifying the COTF created above
- Attempt to navigate to a specific object of the related object type model used in the COTF (i.e. a specific IP address, range, or prefix)

### Expected Behavior

The page loads normally.

### Observed Behavior

A server error is presented:
```
<class 'AttributeError'>

'Table3Model' object has no attribute 'mitel_appliance'   # <----- A non-required COTF of type object and related object type IP Address

Python version: 3.10.12
NetBox version: 4.4.8
Plugins: 
  nbxsync: 1.0.1
  netbox_custom_objects: 0.4.4
```

FWIW, on the previous combo of versions I was running (NB 4.4.7/NBCO 0.4.3) I would run into this error somewhat routinely (a few times a week or more), but refreshing the page a handful of times or so would always do the trick. Now that is no longer working. Also of note, I am always still able to access the object's API URL.

I apologize for the lack of info, but hoping someone can push me in the right direction to help nail down reproduction steps.

bctiemann:
## Summary

- Remove `no_cache=True` from `template_content.py` so the `cache_timestamp` invalidation path handles staleness correctly instead of bypassing the cache on every request (which caused a transiently bad regeneration to be written back permanently)
- Guard `CustomObject.__str__` with `try/except AttributeError` so an incomplete generated model falls back gracefully to `"<DisplayName> <id>"` instead of crashing
- Upgrade silent-skip log from `DEBUG` to `WARNING` in `_fetch_and_generate_field_attrs` when a field is dropped due to a missing ContentType, so operators have a visible signal in logs

## Test plan

- [ ] Reproduce the issue: create a COT with an object field pointing to a now-deleted ContentType; navigate to a built-in NetBox object detail page with a linked custom object — confirm no `AttributeError` and the object displays its fallback label
- [ ] Confirm edit/save of a COTF still invalidates the cache and regenerates the model correctly
- [ ] Confirm self-referential object fields still work (no "Must be TableNModel instance" ValueError from #508)
- [ ] Check logs show a `WARNING` (not just `DEBUG`) when a field is skipped due to a stale `related_object_type_id`

Closes #353

### DIFF
--- a/netbox_custom_objects/models.py
+++ b/netbox_custom_objects/models.py
@@ -131,9 +131,12 @@ class CustomObject(
         primary_field_value = None
         if primary_field:
             field_type = FIELD_TYPE_CLASS[primary_field["field"].type]()
-            primary_field_value = field_type.get_display_value(
-                self, primary_field["name"]
-            )
+            try:
+                primary_field_value = field_type.get_display_value(
+                    self, primary_field["name"]
+                )
+            except AttributeError:
+                primary_field_value = None
         if not primary_field_value:
             return f"{self.custom_object_type.display_name} {self.id}"
         return str(primary_field_value) or str(self.id)
@@ -454,7 +457,7 @@ class CustomObjectType(NetBoxModel):
                         field.name, field.pk, self.slug,
                     )
                 else:
-                    logger.debug(
+                    logger.warning(
                         "Skipping field %r (pk=%s) on COT %r: related_object_type_id=%s "
                         "references a ContentType that no longer exists.",
                         field.name, field.pk, self.slug, field.related_object_type_id,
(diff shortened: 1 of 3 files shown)


bctiemann:
@claude /review

arthanson (approved):

PR #539, opened 2026-05-28
Fixes: #538 - Prevent OOM crash when select/multiselect fields use large base choice sets

### LINKED ISSUE #538: OOM crash (exit 247) on COT detail/add pages when select/multiselect field uses a large base choice set
### Describe the bug

Loading a COT detail page or the custom object add form crashes the browser with exit code 247 (OOM) when the COT has a select or multi-select field whose choice set is backed by a large base list (e.g. the built-in IATA airport codes, ~10 000 entries).

The page spins for 60+ seconds before the browser process is killed.

### Steps to reproduce

1. Create a **Custom Field Choice Set** using the **Airports (IATA)** base list
2. Create a **Custom Object Type** with a **Select** field and/or a **Multi-select** field, both using that choice set
3. Navigate to the COT detail/list page, or try to open the "Add Custom Object" form

### Expected behaviour

The page loads normally. The select widget lazy-loads options via the API URL (Select2 AJAX mode).

### Observed behaviour

The browser spins for 60+ seconds then crashes (exit 247 / OOM killed).

### Root cause

`DynamicMultipleChoiceField` (NetBox core) is missing the `else: self.choices = []` branch that `DynamicChoiceField` has in `get_bound_field()`. When the field has no current selection the full choices list is left intact, and the `APISelect` template renders every item as a static `<option>` element via `{% include 'django/forms/widgets/select.html' %}`. For 10 000+ entries across two fields this is enough to OOM the browser.

Affected call sites in this plugin:

- `SelectFieldType.get_filterform_field` — returns `DynamicMultipleChoiceField(choices=all_10000, ...)`
- `MultiSelectFieldType.get_filterform_field` — same
- `MultiSelectFieldType.get_form_field` — same (non-CSV path)

bctiemann:
## Summary

- Add `_LazyAPIMultipleChoiceField` — a thin subclass of `DynamicMultipleChoiceField` that adds the missing `else: self.choices = []` branch to `get_bound_field()`, preventing the full choices list from being rendered as static `<option>` elements when no value is selected
- Use it in `MultiSelectFieldType.get_form_field` (non-CSV path)
- Pass `choices=[]` to both `get_filterform_field` methods (Select and MultiSelect) — filter-form fields have no server-side validation; the `APISelectMultiple` widget loads options via AJAX

## Root cause

`DynamicMultipleChoiceField` (NetBox core) is missing the `else: self.choices = []` branch that `DynamicChoiceField` has in `get_bound_field()`. When there is no current selection the full choices list is left intact, and the `APISelect` template (`{% include 'django/forms/widgets/select.html' %}`) renders every item as a static `<option>` element. For built-in choice sets such as IATA (~10 000 entries), two such fields (select + multiselect) generate ~20 000 DOM nodes, which is enough to OOM and kill the browser process (exit code 247).

`DynamicChoiceField` already handles this correctly — the fix brings `DynamicMultipleChoiceField`'s behaviour in line.

## Test plan

- [x] Existing field-type tests (62 tests) all pass
- [ ] Manual: create a COT with a select + multi-select field backed by the IATA base choice set; confirm the COT detail/list page and the custom object add form load in normal time without browser crash
- [ ] Confirm filter bar on the custom object list page works (AJAX-loads airports on demand)
- [ ] Confirm saving a new custom object with airport values selected validates and saves correctly

Closes #538

### DIFF
--- a/netbox_custom_objects/field_types.py
+++ b/netbox_custom_objects/field_types.py
@@ -163,6 +163,33 @@ def _make_lazy_cot_fk(cot, field, on_delete, **field_kwargs):
     )
 
 
+class _LazyAPIMultipleChoiceField(DynamicMultipleChoiceField):
+    """
+    DynamicMultipleChoiceField variant that mirrors the DynamicChoiceField
+    behaviour of clearing choices to [] when no value is currently selected.
+
+    DynamicMultipleChoiceField (NetBox core) has no `else` branch in
+    get_bound_field(), so when the field has no current selection the full
+    choices list is left in place and every option is rendered as an <option>
+    element by the APISelect template (which delegates to Django's standard
+    select.html).  For large built-in choice sets such as IATA (~10 000
+    entries) this generates enough DOM nodes to OOM the browser.
+
+    The APISelectMultiple widget already loads options lazily via AJAX; the
+    static choices list is only needed at POST time to validate submitted
+    values.  Clearing it on GET is safe.
+    """
+    def get_bound_field(self, form, field_name):
+        from django.forms import BoundField as _BF
+        bound_field = _BF(form, self, field_name)
+        data = bound_field.value()
+        if data is not None:
+            self.choices = [c for c in self.choices if c[0] and c[0] in data]
+        else:
+            self.choices = []
+        return bound_field
+
+
 class FieldType:
 
     def get_display_value(self, instance, field_name):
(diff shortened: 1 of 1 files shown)


bctiemann:
Note, it might be better to fix this in core, just fixing `DynamicMultipleChoiceField` directly which this will inherit.

PR #541, opened 2026-05-28
Closes: #540 - Clarify Django bootstrap for portable schema export scripts

### LINKED ISSUE #540: Docs: Portable schema exporter example is missing Django bootstrap steps
### Change Type

Correction

### Proposed Changes

The **Exporting a Schema** section of `docs/portable-schema.md` shows a Python snippet calling `export_cots()` but does not explain that Django must be fully bootstrapped before any NetBox or plugin code can be imported.

Users running the snippet as a standalone script hit one of two errors:

1. `ImproperlyConfigured: Requested setting FIELD_CHOICES, but settings are not configured.` — `DJANGO_SETTINGS_MODULE` is not set.
2. `AppRegistryNotReady: Apps aren't loaded yet.` — `DJANGO_SETTINGS_MODULE` is set but `django.setup()` is never called.

Reported in discussion #534.

Proposed fix: expand the section to show two usage paths — `manage.py nbshell` (zero setup, recommended) and a standalone `.py` with `django.setup()` called before any imports — plus an admonition explaining the `AppRegistryNotReady` error.

bctiemann:
## Summary

- Expands the **Exporting a Schema** section with two explicit usage paths: `manage.py nbshell` (zero setup, recommended) and a standalone `.py` script that calls `django.setup()` before any imports.
- Adds an admonition explaining the `AppRegistryNotReady` error so users who encounter it have an immediate explanation and fix.

Addresses user report in discussion #534.

## Test plan

- [ ] Docs render correctly via `mkdocs serve`
- [ ] Option B snippet works when run as `python export_cot.py` from `/opt/netbox/netbox/`

### DIFF
--- a/docs/portable-schema.md
+++ b/docs/portable-schema.md
@@ -237,18 +237,23 @@ Attributes that match their defaults are omitted from exported documents to keep
 
 ### Exporting a Schema
 
-Use the Python API from `netbox_custom_objects.schema.exporter`. There are two supported
-ways to run it:
-
-#### Option 1 — NetBox shell (recommended)
-
-Run `manage.py nbshell` from the NetBox project directory and paste or pipe your script.
-Django is fully initialised for you, and all models are importable immediately:
+Use the Python API from `netbox_custom_objects.schema.exporter`. Before running either
+option, switch to your NetBox installation root and activate its virtualenv:
 
 ```bash
 # Replace /opt/netbox with your NetBox installation root.
-cd /opt/netbox/netbox
-python manage.py nbshell
+export NETBOX_ROOT=/opt/netbox
+cd "$NETBOX_ROOT"
+source "$NETBOX_ROOT/venv/bin/activate"
+```
+
+#### Option 1 — NetBox shell (recommended)
+
+Run `manage.py nbshell` from the NetBox installation root. Django is fully initialised for
+you, and all models are importable immediately:
+
+```bash
+python3 netbox/manage.py nbshell
 ```
 
 Then inside the shell, export specific COTs by slug:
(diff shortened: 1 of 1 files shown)


pheus (changes requested):
This is a useful docs improvement, but I don’t think we should merge the examples in their current form.

The current commands do not make it clear that they need to run in NetBox’s virtualenv. That’s an important detail for this type of documentation: users copying these examples with the system Python may hit import errors or dependency mismatches that look like issues with the export code itself.

We should also avoid presenting `/opt/netbox` as though it is the required installation path. It is a common example, but NetBox can be installed in other locations.

Please update the examples so they:

1. Treat `/opt/netbox` as an example NetBox installation root.
2. Activate NetBox’s virtualenv, or otherwise use the virtualenv Python.
3. Make clear that `/path/to/export_cot.py` is just the user’s script path.

For example:

```bash
# Replace this with your NetBox installation root.
export NETBOX_ROOT=/opt/netbox

source "$NETBOX_ROOT/venv/bin/activate"
cd "$NETBOX_ROOT"
PYTHONPATH="$NETBOX_ROOT/netbox" python3 /path/to/export_cot.py
```

With that clarified, the overall structure of the new section looks good.

bctiemann (commented):
Thanks for the clear feedback, Martin. Updated in 88c7891:

- Introduced a `NETBOX_ROOT` variable in both Option 1 and Option 2 so `/opt/netbox` is explicitly labelled as an example, not a required path.
- Option 2 now shows `source "$NETBOX_ROOT/venv/bin/activate"` before running the script, using your suggested bash snippet.
- Clarified that `/path/to/export_cot.py` is the user's script ("Where `/path/to/export_cot.py` is your script, containing:").
- Also folded in cleboo's suggestion from the issue: an all-COTs export snippet sits alongside the filtered one in Option 1.

pheus (changes requested):
Thanks for the update!

This is much closer. I need to be a bit nitpicky here, but from experience we should make sure readers do not miss the virtualenv setup for either option. Option 1 is still missing that step.

Since Option 1 is the recommended workflow, it is also the example users are most likely to copy. Without activating NetBox’s virtualenv, this can still run against the system Python depending on the user’s environment. It also does not use the same `NETBOX_ROOT` pattern introduced in Option 2.

Please align Option 1 with the NetBox shell documentation and the pattern used in Option 2. For example:

```bash
# Replace /opt/netbox with your NetBox installation root.
export NETBOX_ROOT=/opt/netbox

cd "$NETBOX_ROOT"
source "$NETBOX_ROOT/venv/bin/activate"
python3 netbox/manage.py nbshell
```

And for the non-interactive `nbshell` example:

```bash
python3 netbox/manage.py nbshell < /path/to/export_cot.py
```

For Option 2, the new structure looks right. I’d only suggest using `python3` there as well for consistency with the NetBox docs:

```bash
# Replace /opt/netbox with your NetBox installation root.
export NETBOX_ROOT=/opt/netbox

source "$NETBOX_ROOT/venv/bin/activate"
PYTHONPATH="$NETBOX_ROOT/netbox" python3 /path/to/export_cot.py
```

Alternatively, we could add a short prerequisite section before Option 1 that covers switching into the NetBox installation root and activating the virtualenv once, then keep the individual option examples focused on the actual command.

bctiemann (commented):
Updated in e74b034. Added a single prerequisite block before both options that covers `NETBOX_ROOT`, `cd`, and `source venv/bin/activate` — so neither option can be run without seeing the setup step. Each option now shows only its own command, and all invocations use `python3`.

pheus (approved):
Thank you!

PR #545, opened 2026-06-01
Fixes #535: Deregister through model from app registry on multiobject field delete

### LINKED ISSUE #535: Stale database table reference after deleting a multiobject field causes ProgrammingError on related object deletion
### Plugin Version

0.5.1

### NetBox Version

4.6.1-Docker-5.0.1

### Python Version

3.14.4

### Steps to Reproduce

After deleting a multiobject field from a Custom Object Type, the plugin leaves a stale reference in the database. When trying to delete a NetBox object that was previously used as a relation target of that field, a ProgrammingError is raised because the plugin still queries the (now non-existent) M2M table of the deleted field.

Create a Custom Object Type (e.g. certificate_location)
Add a multiobject field to it targeting an IPAM model (e.g. ipam.service) — field name e.g. ssl_service2
Delete the field via the plugin UI
Try to delete any object of the previously targeted model (e.g. any IPAM service)

### Expected Behavior

Deleting the field cleanly removes all database artifacts. Deleting a related NetBox object succeeds without errors.

### Observed Behavior

django.db.utils.ProgrammingError: relation "custom_objects_2_ssl_service2" does not exist
LINE 1: ... "custom_objects_2_ssl_service2"."target_id" FROM "custom_ob...

GET /ipam/services/5/delete/?return_url=/ipam/services/
The plugin queries the M2M table of the deleted field when checking for related objects during deletion.

**Investigation**

netbox_custom_objects_customobjecttypefield contains no entry for ssl_service2 — the field record was correctly deleted
\dt custom_objects* shows the M2M table custom_objects_2_ssl_service2 is missing — it was either never created or already dropped
netbox_custom_objects_customobjecttypefield_related_object_93dc contains no stale entries for the deleted field
The source of the stale reference that causes the plugin to still query the missing table could not be identified

**Workaround**
```
CREATE TABLE custom_objects_2_ssl_service2 (
    id        integer GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
    source_id bigint NOT NULL,
    target_id bigint NOT NULL,
    UNIQUE (source_id, target_id),
    FOREIGN KEY (source_id)
        REFERENCES custom_objects_2(id) DEFERRABLE INITIALLY DEFERRED,
    FOREIGN KEY (target_id)
        REFERENCES ipam_service(id) DEFERRABLE INITIALLY DEFERRED
);
CREATE INDEX ON custom_objects_2_ssl_service2 (source_id);
CREATE INDEX ON custom_objects_2_ssl_service2 (target_id);
```

The table remains empty — the deletion proceeds normally.

**Additional context**

The table naming convention custom_objects_{type_id}_{field_name} uses target_id as the FK column for direct object relations, in contrast to the polymorphic multiobject fields which use content_type_id + object_id. This distinction was discovered during the investigation.

bctiemann:
## Summary

When a `multiobject` field is deleted, `CustomObjectTypeField.delete()` was correctly dropping the through table via `schema_editor.delete_model()`, but the through model class remained registered in Django's `apps.all_models`. Django's cascade-delete collector walks `apps.all_models` when computing what to clean up before deleting an object. Because it still found the through model (with its FK to the relation target), it generated a query against the already-dropped table, producing:

```
django.db.utils.ProgrammingError: relation "custom_objects_2_ssl_service2" does not exist
```

**Fix:** After dropping the through table — in both the non-polymorphic and polymorphic code paths — remove the through model from `apps.all_models` and call `apps.clear_cache()` to flush Django's internal relation caches. This mirrors the deregistration already performed in `CustomObjectType.delete()` for full COT teardown.

## Test plan

- [x] New regression test `RelatedNameTestCase.test_deleting_multiobject_field_deregisters_through_model`: creates a multiobject field, deletes it, then deletes a related object and asserts no `ProgrammingError` is raised. A secondary assertion confirms the through model is no longer in `apps.all_models`.
- [x] Full `test_models` suite (98 tests + 1 skip) passes.

**Verified pre-fix failure:** The regression test was confirmed to catch the bug by temporarily disabling the deregistration block and running the test against the pre-fix code path. Without the fix, `slb_instance.delete()` raises the exact error from the issue report:

```
ProgrammingError: relation "custom_objects_285_related_slbs" does not exist
LINE 1: ...custom_objects_285_related_slbs"."target_id" FROM "custom_ob...
```

The test is ordered so this `ProgrammingError` is the first and most visible failure — a later reviewer or CI run sees the user-facing error directly, not an opaque registry state assertion.

### DIFF
--- a/netbox_custom_objects/models.py
+++ b/netbox_custom_objects/models.py
@@ -2391,12 +2391,17 @@ class CustomObjectTypeField(CloningMixin, ExportTemplatesMixin, ChangeLoggedMode
     def delete(self, *args, **kwargs):
         field_type = FIELD_TYPE_CLASS[self.type]()
         model = self.custom_object_type.get_model()
+        _dropped_through_model = None
 
         with connection.schema_editor() as schema_editor:
             if self.is_polymorphic:
                 if self.type == CustomFieldTypeChoices.TYPE_OBJECT:
                     field_type.remove_polymorphic_object_columns(self, model, schema_editor)
                 elif self.type == CustomFieldTypeChoices.TYPE_MULTIOBJECT:
+                    try:
+                        _dropped_through_model = model._meta.apps.get_model(APP_LABEL, self.through_model_name)
+                    except LookupError:
+                        pass
                     field_type.drop_polymorphic_m2m_table(self, model, schema_editor)
             else:
                 model_field = field_type.get_model_field(self)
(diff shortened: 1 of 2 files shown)


bctiemann:
@claude /review

jnovinger (approved):

PR #542, opened 2026-05-29
Fixes #370: lazy serializer registration via module __getattr__ and skip_object_fields guard

bctiemann:
## Summary

Two interlocking bugs caused `SerializerNotFound` in workers that started before a COT was created (or whose `ready()` ran against an empty DB / hit a transient connection error):

**Bug 1 — partial serializer clobbered the full one**

`get_serializer_class(model, skip_object_fields=True)` unconditionally called `setattr()` on the module, silently replacing the registered full serializer with a partial one that drops FK fields. This partial variant is built recursively when constructing nested field descriptors for FK-to-COT fields; it was never intended to be the module-level export. Fix: only call `setattr()` when `skip_object_fields=False`.

**Bug 2 — no fallback when serializer was never registered**

`import_string("netbox_custom_objects.api.serializers.Table{N}ModelSerializer")` raised `SerializerNotFound` in any worker whose `ready()` completed without finding that COT (Scenario A/B from the issue). `AppConfig.get_model()` generated the model on demand but never registered its serializer. Adding `get_serializer_class()` to `get_model()`'s fallback path (Options 1 & 2 reported in the issue) caused a `ValueError: Cannot assign` because the partial-variant registration bug (Bug 1) would overwrite the full serializer mid-request. Fix: add a module-level `__getattr__` (PEP 562) that generates and caches the serializer the first time `import_string()` asks for it. Now that Bug 1 is fixed, the `__getattr__` path can call `get_serializer_class(model)` safely without risk of clobbering related models' serializers.

## Changes

- `netbox_custom_objects/api/serializers.py`: guard the `setattr()` with `if not skip_object_fields`; add module-level `__getattr__`
- `netbox_custom_objects/tests/test_models.py`: `LazySerializerRegistrationTestCase` — 3 regression tests

## Test plan

- [ ] `LazySerializerRegistrationTestCase.test_module_getattr_generates_serializer_on_demand` — `__getattr__` generates and caches serializer for a COT whose serializer was never pre-registered
- [ ] `LazySerializerRegistrationTestCase.test_skip_object_fields_does_not_overwrite_full_serializer` — calling `get_serializer_class(model, skip_object_fields=True)` does not replace the registered full serializer
- [ ] `LazySerializerRegistrationTestCase.test_child_serializer_does_not_clobber_parent_serializer` — building child's serializer (which creates a nested partial for parent) leaves parent's full serializer intact
[...]

Issue #370, opened 2026-01-16
Fix custom objects serialization when used with custom fields


dmulyalinvs:
### Plugin Version

0.4.3

### NetBox Version

4.4.9

### Python Version

3.11

### Steps to Reproduce

1. Create custom object
2. Create instance of custom object
3. Create multi-object custom field for interfaces that references to "Custom objects > <custom_object_name>"
4. Edit interface to add custom object to the custom field
5. Try to query interface via REST API

### Expected Behavior

RESP API request should return correct data and not throw error

### Observed Behavior

Getting this error:

```
Environment:


Request Method: GET
Request URL: https://netbox.lab/api/dcim/interfaces/?device=device1&name=et-0/0/0

Django Version: 5.2.9
Python Version: 3.12.3
Installed Applications:
['django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'django.contrib.humanize',
 'django.forms',
 'corsheaders',
 'debug_toolbar',
 'django_filters',
 'django_htmx',
 'django_tables2',
 'django_prometheus',
 'strawberry_django',
 'mptt',
 'rest_framework',
 'social_django',
 'sorl.thumbnail',
 'taggit',
 'timezone_field',
 'core',
 'account',
 'circuits',
 'dcim',
 'ipam',
 'extras',
 'tenancy',
 'users',
 'utilities',
 'virtualization',
 'vpn',
 'wireless',
 'django_rq',
 'drf_spectacular',
 'drf_spectacular_sidecar',
 'netbox_bgp.BGPConfig',
 'netbox_custom_objects.CustomObjectsPluginConfig']
Installed Middleware:
['strawberry_django.middlewares.debug_toolbar.DebugToolbarMiddleware',
 'corsheaders.middleware.CorsMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.locale.LocaleMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware',
 'django.middleware.security.SecurityMiddleware',
 'django_htmx.middleware.HtmxMiddleware',
[...]

PR #547, opened 2026-06-01
Replace __getattr__ serializer fallback with serializer_resolver PluginConfig hook

### LINKED ISSUE #546: Replace __getattr__ serializer fallback with serializer_resolver PluginConfig hook
### Proposed Changes

Replace the module-level `__getattr__` fallback in `netbox_custom_objects/api/serializers.py` (added in #370 / PR #542) with a `PluginConfig.serializer_resolver` hook — the purpose-built NetBox extension point for exactly this use case.

Concretely:
- Add `serializer_resolver = "api.serializers.serializer_resolver"` to `CustomObjectsConfig` in `__init__.py`
- Add a `serializer_resolver(model, prefix='')` function in `api/serializers.py` that returns the CO serializer for dynamically-generated CO models and `None` for everything else
- Remove the `__getattr__` module function (fully superseded)
- Update `LazySerializerRegistrationTestCase` in `tests/test_models.py` to cover `serializer_resolver` directly

### Justification

`serializer_resolver` is the designed PluginConfig extension point for resolving serializers for dynamically-generated models that have no importable path. It runs before NetBox's `import_string()` fallback, making it strictly more reliable than `__getattr__` (PEP 562), which fires only after attribute lookup fails. Using the official hook also makes the integration visible and explicit rather than relying on a Python module protocol that NetBox maintainers have no way to discover without reading the source.

This change is also required to avoid a merge conflict when `main` is eventually merged into `feature`: the `feature` branch (PR #476) already uses `serializer_resolver` and omits `__getattr__`. Aligning `main` now makes that future merge a clean fast-forward on the serializer file.

Ref: #370 (original issue that motivated the `__getattr__` approach)

bctiemann:
## Summary

- Replaces the module-level `__getattr__` (PEP 562) added in PR #542 with `PluginConfig.serializer_resolver` — the purpose-built NetBox extension point for dynamically-generated model serializers
- Adds `serializer_resolver = "api.serializers.serializer_resolver"` to `CustomObjectsPluginConfig`
- Adds `serializer_resolver(model, prefix='')` in `api/serializers.py`; removes `__getattr__`
- Updates `LazySerializerRegistrationTestCase` to exercise `serializer_resolver()` directly

The `skip_object_fields` guard on `setattr()` (the other fix from PR #542) is unchanged.

## Why

`__getattr__` fires only after normal attribute lookup fails — it is a Python module fallback that NetBox has no visibility into. `serializer_resolver` is called by `get_serializer_for_model()` *before* any import-path lookup is attempted, making it strictly more reliable: a CO serializer is always generated on demand and `SerializerNotFound` can never be raised due to a missed startup-time registration.

Using the official hook also makes the integration explicit and discoverable rather than relying on an opaque Python protocol.

## Pre-empts a merge conflict

The `feature` branch (PR #476) already uses `serializer_resolver` and omits `__getattr__`. Aligning `main` now means the eventual `main`->`feature` merge produces no conflict in `api/serializers.py`.

## Test plan

- [ ] `LazySerializerRegistrationTestCase.test_resolver_generates_serializer_on_demand`
- [ ] `LazySerializerRegistrationTestCase.test_skip_object_fields_does_not_overwrite_full_serializer`
- [ ] `LazySerializerRegistrationTestCase.test_child_serializer_does_not_clobber_parent_serializer`
- [ ] Existing test suite passes

Closes: #546
Ref: #370

### DIFF
--- a/netbox_custom_objects/__init__.py
+++ b/netbox_custom_objects/__init__.py
@@ -133,6 +133,7 @@ class CustomObjectsPluginConfig(PluginConfig):
     }
     required_settings = []
     template_extensions = "template_content.template_extensions"
+    serializer_resolver = "api.serializers.serializer_resolver"
 
     @staticmethod
     def should_skip_dynamic_model_creation():
--- a/netbox_custom_objects/api/serializers.py
+++ b/netbox_custom_objects/api/serializers.py
@@ -694,7 +694,8 @@ def get_serializer_class(model, skip_object_fields=False):
     )
 
     # Register the FULL serializer as a module attribute so NetBox's import_string()
-    # and the module-level __getattr__ fallback can find it.
+    # can find it (the serializer_resolver below generates on demand; this keeps
+    # any direct import-path lookups working too).
     # The partial variant (skip_object_fields=True) is used only as a nested field
     # descriptor inside another serializer class and must NOT be stored on the module
     # — doing so would silently replace the full serializer with an incomplete one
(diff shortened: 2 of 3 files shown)


bctiemann:
Closing — `serializer_resolver` was introduced in NetBox core after v4.6.1 and is not available in the 4.5.x/4.6.x versions that `main` supports (min_version 4.5.2, max_version 4.6.99). The `__getattr__` fallback in PR #542 remains correct for `main`. When `main` is eventually merged into `feature` (which targets NetBox 4.7+), the conflict in `api/serializers.py` should be resolved by keeping `serializer_resolver` and dropping `__getattr__`.

PR #548, opened 2026-06-01
Fixes #522 + implements #385: related_name for polymorphic fields (form fix + reverse descriptors)

### LINKED ISSUE #522: Unable to Configure a Reverse Relation Name When Configuring a Polymorphic Object COTF
### Plugin Version

0.5.1

### NetBox Version

4.5.10

### Python Version

3.12.12

### Steps to Reproduce

1. Create a custom object type of name foo
2. Attempt to create a COTF of name bar with the following attributes:
   * Type: Object (or Multiple objects)
   * Polymorphic: True
   * Related object type: any (I am doing IP Address, IP Range, and Prefix)
   * Reverse relation name: foo_bar

### Expected Behavior

I am able to create a polymorphic COTF with multiple related object types *and* a reverse relation name.

### Observed Behavior

The reverse relation name field disappears when checking the Polymorphic checkbox, as shown in the image below. I may be being a bit presumptuous assuming this is a bug,, although netboxlabs/netbox-custom-objects#385 specifies that it should be available for object and multiobject fields. if this is a known limitation, I'm happy to create a feature request.

Without Polymorphic checked: Image

With Polymorphic checked:

Image

bctiemann:
## Summary

This PR fixes a form regression and implements full reverse-relation support for polymorphic `Object` and `MultiObject` CO fields.

### Bug fix (issue #522)

The `related_name` field disappeared from the COTF form whenever the **Polymorphic** checkbox was checked on an Object or MultiObject field.

**Root cause:** `CustomObjectTypeFieldForm.__init__` removed `related_object_type` from `self.fields` (replacing it with `related_object_types`) before the block that appended `related_name` to the fieldset. That block used `"related_object_type" in self.fields` as its signal — which was always `False` for polymorphic fields — and fell through to the `else` branch that deleted `related_name` entirely.

**Fix:** Check for either `related_object_type` **or** `related_object_types` as the object-field signal. Tighten `is_single_object` to exclude polymorphic Object fields so `on_delete_behavior` is not shown for them either (polymorphic Object fields use GFK columns, not a DB-level FK).

### Regression validation

The two tests that cover the fixed paths confirm the bug was reproducible before the fix:

```
# On pre-fix forms.py (commit dae3086):
FAILED test_related_name_visible_for_polymorphic_object       ← related_name was absent
FAILED test_related_name_visible_for_polymorphic_multiobject  ← related_name was absent

# On this branch (all 6 pass):
OK test_related_name_visible_for_polymorphic_object
OK test_related_name_visible_for_polymorphic_multiobject
OK test_related_name_visible_for_non_polymorphic_object
OK test_related_name_visible_for_non_polymorphic_multiobject
OK test_related_name_absent_for_non_object_type
OK test_on_delete_behavior_visible_only_for_non_polymorphic_object
```

### Reverse relations for polymorphic fields (issue #385)

Previously, `related_name` was silently ignored for polymorphic Object and MultiObject fields — the form showed the field (after this fix), but the value was never wired up. This PR makes it fully functional.

**GFK (polymorphic Object):** `PolymorphicObjectReverseDescriptor` / `PolymorphicObjectReverseManager` — queries the CO model filtering on `{field_name}_content_type` + `{field_name}_object_id`.

**M2M (polymorphic MultiObject):** `PolymorphicMultiObjectReverseDescriptor` / `PolymorphicMultiObjectReverseManager` — queries the polymorphic through table by `content_type_id` + `object_id`, then fetches matching source CO instances.

Both descriptors are injected onto target model classes via `setattr` in `_after_model_generation` and removed via `delattr` when the field is deleted. The source model is resolved lazily (`CustomObjectType.objects.get(pk=cot_pk).get_model()`) to avoid holding stale class references across model regeneration.

Usage after this PR:

```python
# Polymorphic GFK field named "primary_site" with related_name="co_instances":
site.co_instances.all()    # QS of CO instances whose GFK points at this site
site.co_instances.count()
site.co_instances.exists()

# Polymorphic M2M field named "linked_sites" with related_name="co_links":
site.co_links.all()        # QS of CO instances whose M2M includes this site
```

**9 new tests** in `PolymorphicReverseDescriptorTest` cover: descriptor presence on target class, correct query results across multiple target types, `count`/`exists`, blank `related_name` no-op, and cleanup on field delete.

### Form fix
- [ ] Create a COT, add an Object COTF with Polymorphic=checked — verify `related_name` is visible
- [ ] Same with MultiObject + Polymorphic=checked
- [ ] Non-polymorphic Object shows both `related_name` and `on_delete_behavior`
- [ ] Non-polymorphic MultiObject shows `related_name` but not `on_delete_behavior`
- [ ] Polymorphic Object shows `related_name` but not `on_delete_behavior`

### Reverse relations
- [ ] Create a polymorphic Object COTF with `related_name="co_ref"` pointing at Site; create a CO instance pointing at a specific site; verify `site.co_ref.all()` returns that CO instance
- [ ] Same for polymorphic MultiObject COTF
- [ ] Verify `count()` and `exists()` return correct values before/after adding a CO instance
- [ ] Delete the field; verify the reverse accessor is removed from the target class


bctiemann:
@claude /review

bctiemann:
@claude /review changes since the last review

arthanson (changes requested):
Rest of the code looks good, but the related_name collision code needs work, it checks related custom object fields but not the related names on the destination model.  For example I can create a polymorphic object to Site and in related name use 'tenant', then if I go to the site it will blow up.  Worse, cleanup does delattr after an isinstance check so it removes the descriptor and the original Site.tenant is never restored.

The new collision check in clean() only compares against other CO fields, not the target model's own attributes. Add a guard rejecting related_name that collides with an existing attribute on any target class.

bctiemann:
## Follow-up review: remaining corner cases after collision guard

The three collision guards (in `clean()`, `_wire_polymorphic_reverse_descriptors`, and the `post_add` signal branch) correctly address the original feedback. The following gaps remain and are worth tracking before merge.

---

### 1. Two CO fields from different COTs silently overwrite each other at startup (medium)

`clean()` catches this at form/API time via the CO-to-CO collision query. But both `_wire` and the `post_add` guard only block **non-CO** conflicts:

```python
if existing is not None and not isinstance(existing, (PolymorphicObjectReverseDescriptor, ...)):
    # skip
```

If COT A wires `Site.co_ref` first, COT B's `_wire` sees a CO descriptor → `isinstance` passes → silently overwrites. At startup all fields are wired in sequence without `clean()`, so any historical data with two fields sharing a name+target will silently lose whichever was wired first.

Fix: extend the guard to check descriptor identity — if `existing.cot_pk` or the field identifier differ from the field being wired, treat it as a conflict and skip.

---

### 2. `_source_model()` raises unhandled `DoesNotExist` if the COT was deleted (medium)

```python
def _source_model(self):
    return CustomObjectType.objects.get(pk=self._cot_pk).get_model()
```

If the COT is deleted but the descriptor is left wired on a target class (cleanup failure or pre-fix data), any access to `site.co_field.all()` raises `DoesNotExist` with no handling. Should return an empty queryset (or a named exception subclass) so callers get predictable behaviour.

---

### 3. Target-attribute check in `clean()` is skipped for new (unsaved) fields (low / UX)

The entire block is gated on `self.pk`. A brand-new polymorphic field with `related_name='devices'` targeting `Site` passes `clean()` without error, saves, the `m2m_changed` signal fires, `_wire` silently skips the setattr and logs a warning — but the user never sees a form error. The reverse descriptor is simply never wired.

The CO-to-CO collision check has the same limitation for the same reason (no M2M rows yet), but the target-attribute check doesn't require M2M rows — it could be run if the submitted `related_object_types` are available in the form context.

---

### 4. CO-to-CO reverse descriptors missing when the target CO type is generated after the source (low-medium)

During startup, models are generated in `CustomObjectType.objects.all()` order. When COT A is generated and its polymorphic field targets COT B, `_wire` calls `ot.model_class()` for COT B's `ObjectType`. If COT B hasn't been generated yet, `model_class()` returns `None` → skipped. When COT B is eventually generated, `_after_model_generation` only wires COT B's own field descriptors — it does not re-wire COT A's descriptor onto the freshly-created COT B class.

Result: CO-to-CO polymorphic reverse descriptors are silently absent unless COT B's primary key happens to sort before COT A's. CO-to-core-model links are unaffected (Django classes are always available).

---

### 5. Polymorphic vs. non-polymorphic `related_name` cross-check gap (low)

The collision checks are siloed:
- Polymorphic fields check against other polymorphic fields via `related_object_types` (M2M)
- Non-polymorphic fields check against other non-polymorphic fields via `related_object_type_id` (FK)

A polymorphic field and a non-polymorphic field can both claim the same `related_name` on the same target type without either check firing. At wiring/model-generation time, whichever runs second overwrites the first silently.

---

### 6. `_source_model()` performs a DB query on every `all()` / `count()` / `exists()` call (low / performance)

`CustomObjectType.objects.get(pk=cot_pk)` hits the database on every manager method call. `get_model()` uses the COT model cache so the model lookup itself is cheap after the first call, but the COT row fetch is not cached. Template loops that call `site.co_field.all()` per-row in a list view will pay this per row. The manager could cache `_source_model_cls` on the instance after the first resolution.

bctiemann:
@claude /review and limit response to changes since the last review, as well as any other corner cases in configuration involving object/multiobject fields, polymorphic and non-polymorphic interactions, etc.

arthanson (changes requested):
I think you need to use `getattr(target_cls, name, None)` instead of `__dict__.get` in the four guard sites.  Use a related_name of `save` then go edit / save the related object and you will get an error:

1. put to object type Site.
2. use related name 'save'
3. add an object of this type and select a site.
4. go to that site, edit it and change description.
5. save the site
6. error occurs.

bctiemann:
Fixed those four call sites + a fifth for consistency; also added two new tests handling that inherited method condition.

arthanson (approved):

PR #552, opened 2026-06-02
Fixes #550: Allow explicit null for optional object/multiobject fields in API

### LINKED ISSUE #550: Custom Objects API rejects null for optional object/multiobject fields
### Plugin Version

0.4.10

### NetBox Version

4.4.10

### Python Version

3.12.3

### Steps to Reproduce

1. Create a Custom Object Type (e.g. test_object_type).
2. Add a field to it: name="region", type="object", required=False, related to dcim.region.
3. Create an instance of that type via the REST API, passing the optional object field explicitly as null:

   ```
   import pynetbox
   api = pynetbox.Api(url=URL, token=TOKEN)
   api.plugins.custom_objects.test_object_type.create({"region": None})
   (equivalently POST /api/plugins/custom-objects/test-object-type/ with body {"region": null})
   ```

### Expected Behavior

The instance is created with region unset, since the field is optional. This matches the UI, which accepts a null/empty object field, and matches POST with the key omitted, which succeeds (201, region: null).

### Observed Behavior

The request is rejected:

```
400 Bad Request: {'region': ['This field may not be null.']}
```

<br>Sending an explicit null for any optional object (or multiobject) field fails. The same error also blocks clearing an already-set value via PATCH {"region": null}. Omitting the key entirely is the only API path that works, and that cannot clear an existing value.

bctiemann:
## Summary

- `ObjectFieldType.get_serializer_field()` and `MultiObjectFieldType.get_serializer_field()` now pass `allow_null=not field.required` to their DRF serializer constructors
- Without this, passing `null` for an optional object or multiobject field returned HTTP 400 even though the field is marked `required=False`
- Regression tests added in `NullOptionalObjectFieldTest` covering POST and PATCH with null for both field types

## Test plan

- [ ] `python manage.py test netbox_custom_objects.tests.test_api.NullOptionalObjectFieldTest`
- [ ] Verify POST with `{"name": "x", "device": null}` returns 201 for a COT with an optional device field
- [ ] Verify PATCH with `{"device": null}` returns 200 and clears the field

### DIFF
--- a/netbox_custom_objects/api/serializers.py
+++ b/netbox_custom_objects/api/serializers.py
@@ -566,7 +566,7 @@ def get_serializer_class(model, skip_object_fields=False):
         if many_to_many:
             for field_name, value in many_to_many.items():
                 field = getattr(instance, field_name)
-                field.set(value)
+                field.set(value if value is not None else [])
 
         for field_name, value in poly_gfk.items():
             setattr(instance, field_name, value)
@@ -609,7 +609,7 @@ def get_serializer_class(model, skip_object_fields=False):
 
         for attr, value in m2m_fields:
             field = getattr(instance, attr)
-            field.set(value, clear=True)
+            field.set(value if value is not None else [], clear=True)
 
         for field_name, value in poly_m2m.items():
             mgr = getattr(instance, field_name)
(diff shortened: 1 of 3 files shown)


bctiemann:

arthanson (approved):

PR #554, opened 2026-06-03
Fixes #529: Render selection field colors and labels correctly

### LINKED ISSUE #529: Display Selection Fields correct
### Plugin Version

v0.5.1

### NetBox Version

v4.6.1

### Python Version

3.14.3

### Steps to Reproduce

1. Create Custom Choice Field Set with the new netbox version:

Image

2. Create new Field for a custom object type as Selection field and reference to created custom choice field set

3. Create instance of custom object type and select one choice of the previous created custom choice field set

### Expected Behavior

The custom choice field set values should be displayed with colour and the label:

Image

### Observed Behavior

In the Object instance table the colour is missing:

Image

In the detailed object view is the value displayed and not the Label. Colour is also missing:

Image

bctiemann:
## Summary

- Selection (`select`) and multi-selection (`multiselect`) fields backed by a `CustomFieldChoiceSet` with colors now render colored Bootstrap badges in both the list table and the object detail view.
- The detail view was broken because the template directly `{% include %}`d `builtins/customfield_value.html` without computing `color`, so the badge branch was never reached. It now uses `{% customfield_value field value %}` which resolves label and color correctly.
- The table columns previously returned a plain label string. They now emit `<span class="badge text-bg-{color}">` markup when a color is configured; fields without colors fall back to the previous plain text / comma-joined behaviour.

## Changes

- `models.py` — Added `get_choice_label()` and `get_choice_color()` to `CustomObjectTypeField`, delegating to `choice_set`, to match the `CustomField` API expected by the `customfield_value` inclusion tag.
- `field_types.py` — `SelectFieldType` and `MultiSelectFieldType` table columns capture `field.choice_set` and render colored badges when a color is present.
- `custom_object_utils.py` — `get_field_value` template filter simplified to return the raw stored value; label/color resolution is now handled downstream by `customfield_value`.
- `customobject.html` — Replaced `{% with %}{% include %}` pattern with `{% customfield_value field value %}`.
- `test_field_types.py` — Updated existing column/filter tests; added tests for badge rendering and the new `get_choice_label`/`get_choice_color` methods.

## Screenshots

**List view** — colored badges in the `VALUE` (select) and `VALUES` (multiselect) columns:

Screenshot 2026-06-03 at 3 01 37 PM

**Detail view** — colored badges on the object detail panel:

Screenshot 2026-06-03 at 3 01 46 PM

## Test plan

- [ ] Create a `CustomFieldChoiceSet` with `choice_colors` configured
- [ ] Create a Custom Object Type with both a `select` and a `multiselect` field using that choice set
- [ ] Create instances with different values; verify colored badges appear in the list table
- [ ] Open an instance detail view; verify colored badges appear for both field types
- [ ] Verify fields with no colors configured still render as plain text / comma-joined labels
- [ ] `python manage.py test netbox_custom_objects.tests.test_field_types` passes

### DIFF
--- a/netbox_custom_objects/field_types.py
+++ b/netbox_custom_objects/field_types.py
@@ -480,12 +480,19 @@ class SelectFieldType(FieldType):
 
     def get_table_column_field(self, field, **kwargs):
         choices_dict = dict(field.choices)
+        choice_set = field.choice_set
 
         class _SelectLabelColumn(tables.Column):
             def render(self, value):
                 if value is None:
                     return self.default
-                return choices_dict.get(value, value)
+                label = choices_dict.get(value, value)
+                color = choice_set.get_choice_color(value) if choice_set else None
+                if color:
+                    return mark_safe(
+                        f'<span class="badge text-bg-{escape(color)}">{escape(label)}</span>'
+                    )
+                return label
 
         return _SelectLabelColumn()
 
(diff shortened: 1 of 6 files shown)


bctiemann:
@claude /review

arthanson (approved):
For a future might be nice to have it show the color in the list view as well

PR #555, opened 2026-06-04
Fixes #371: Persist tags on custom object create and update

### LINKED ISSUE #371: Tags Not Persisted on MLAG Custom Objects
### Plugin Version

v0.4.4

### NetBox Version

4.4.6

### Python Version

3.12.3

### 1. Create MLAG Domain with tags via POST

```python
import httpx

url = "https://your-netbox/api/plugins/custom-objects/mlag-domains/"
token = "your-token"

payload = {
    "name": "Test Domain",
    "tags": [{"name": "nbx-ingestor"}, {"name": "napalm"}]
}

response = httpx.post(
    url,
    headers={"Authorization": f"Token {token}", "Content-Type": "application/json"},
    json=payload,
    verify=False
)

# Response includes tags
result = response.json()
print(f"Tags in POST response: {result.get('tags')}")
# Output: [{'id': 38, 'name': 'nbx-ingestor', ...}, {'id': 43, 'name': 'napalm', ...}]
```

### 2. Fetch the created object

```python
# Fetch the same object back
fetch_response = httpx.get(f"{url}{result['id']}/", headers={"Authorization": f"Token {token}"}, verify=False)
fetched = fetch_response.json()
print(f"Tags after fetch: {fetched.get('tags')}")
# Output: []  <-- Tags are empty!
```

### 3. Same issue with PATCH

```python
# Try to add tags via PATCH
patch_response = httpx.patch(
    f"{url}{result['id']}/",
    headers={"Authorization": f"Token {token}", "Content-Type": "application/json"},
    json={"tags": [{"name": "nbx-ingestor"}, {"name": "napalm"}]},
    verify=False
)
patch_result = patch_response.json()
print(f"Tags in PATCH response: {patch_result.get('tags')}")
# Output: [{'id': 38, 'name': 'nbx-ingestor', ...}, {'id': 43, 'name': 'napalm', ...}]

# Fetch again
fetch2 = httpx.get(f"{url}{result['id']}/", headers={"Authorization": f"Token {token}"}, verify=False)
print(f"Tags after fetch: {fetch2.json().get('tags')}")
# Output: []  <-- Still empty!
```

## Additional Issues

DELETE operations on MLAG domains are also returning 500 errors:

```python
response = httpx.delete(f"{url}{domain_id}/", headers={"Authorization": f"Token {token}"}, verify=False)
# Status: 500
```

## Impact

This bug prevents automated tools from properly tagging MLAG custom objects for lifecycle management (orphan tracking, source attribution, etc.).

## Expected Behavior

Tags should be persisted to the database and returned when fetching the object.

## Actual Behavior

- POST/PATCH responses include the tags (suggesting they were accepted)
- Fetching the object returns empty tags
- Tags are not saved to the database

bctiemann:
## Summary

Tags submitted via POST or PATCH to custom object endpoints were silently discarded. The response appeared to include the tags (giving the illusion of success), but a subsequent GET returned an empty `tags` list.

## Root Cause

`taggit`'s `TaggableManager` is not detected by DRF's `model_meta.get_field_info()` as a standard M2M relation. Without explicit handling, `tags` remained in `validated_data` and was passed directly to `Model._default_manager.create()`. Django's `Model.__init__` then called `setattr(obj, 'tags', [...])` which, since `TaggableManager` defines no `__set__`, shadowed the manager with a plain list on the instance dict. The serializer read this back for the response, making it look correct — but `TaggableManager`'s DB-backed storage was never touched.

The same issue applied to `update()`, where `tags` fell through to the `setattr` branch of the field-setting loop.

## Fix

Mirrors `TaggableModelSerializer._save_tags()`: explicitly pop `tags` from `validated_data` before model creation/update, then persist via `instance.tags.set([t.name for t in tags])` after the instance exists. `instance._tags` is also set for NetBox's change-logging machinery.

## Tests

Two regression tests added to `CustomObjectTest`:

- `test_create_with_tags_persists_to_db` — POSTs with tags, then asserts `instance.tags.names()` on a fresh DB fetch contains the submitted tag.
- `test_patch_with_tags_persists_to_db` — PATCHes with tags, then asserts the same after `refresh_from_db()`.

Both tests fail against the unfixed code and pass with the fix.

### DIFF
--- a/netbox_custom_objects/api/serializers.py
+++ b/netbox_custom_objects/api/serializers.py
@@ -543,6 +543,13 @@ def get_serializer_class(model, skip_object_fields=False):
     def create(self, validated_data):
         ModelClass = self.Meta.model
 
+        # taggit's TaggableManager is not detected by model_meta.get_field_info() as a
+        # standard M2M relation.  If left in validated_data it gets passed to
+        # Model._default_manager.create(), where Django's __init__ sets it as a plain
+        # instance attribute (shadowing the manager), giving the illusion of success but
+        # never persisting to the DB.  Pop it here and save via taggit's own API.
+        tags = validated_data.pop('tags', None)
+
         info = model_meta.get_field_info(ModelClass)
         many_to_many = {}
         for field_name, relation_info in info.relations.items():
@@ -563,6 +570,12 @@ def get_serializer_class(model, skip_object_fields=False):
 
         instance = ModelClass._default_manager.create(**validated_data)
 
+        if tags is not None:
+            instance._tags = tags
+            instance.tags.set([t.name for t in tags])
+        else:
+            instance._tags = []
+
         if many_to_many:
             for field_name, value in many_to_many.items():
                 field = getattr(instance, field_name)
@@ -581,6 +594,11 @@ def get_serializer_class(model, skip_object_fields=False):
 
     # Stock DRF update() with custom field.set() for M2M
     def update(self, instance, validated_data):
+        # Pop tags before the setattr loop — taggit's manager has no __set__, so
+        # leaving tags in validated_data would shadow the manager with a plain list.
+        tags = validated_data.pop('tags', None)
+        instance._tags = tags or []
+
         info = model_meta.get_field_info(instance)
 
         # Pop polymorphic GFK fields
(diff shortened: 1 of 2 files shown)


bctiemann:
@claude /review

bctiemann:
Added tests to cover the review-raised gaps.

pheus (approved):
Thanks!

I’m going to merge this as-is since it addresses the reported tag persistence issue for the standard `tags` field on create/update, and the tests cover the important create, patch, clear, and preserve cases.

There is one small follow-up we can track separately rather than holding up this fix:

- NetBox 4.6 also supports `add_tags` / `remove_tags`, and the generated serializer may need explicit handling for those because it overrides `create()` / `update()`.

Issue #30, opened 2025-06-27
GraphQL support for Custom Objects


mrmrcoleman:
### Plugin Version

commit 722bbd624c8af515dc584b719e2864cbfcc14dd0 (HEAD -> public-preview, origin/public-preview) Author: btiemann btiemann@netboxlabs.com Date:   Thu Jun 26 15:31:01 2025 -0400

### Proposed functionality

We expect users to want to interact with Custom Objects via GraphQL. GraphQL has become a popular alternative to our REST API for some users.

### Use case

Further refinement required

PR #565, opened 2026-06-08
#30 - add graphql support for custom objects

arthanson:
### Fixes: #30 

Custom objects are now queryable through NetBox's GraphQL API at /graphql/, alongside built-in NetBox models. Each CustomObjectType gets two root query fields: <slug> (single object by id) and <slug>_list (paginated list).

## How it works

- A Strawberry type is generated at runtime for each CustomObjectType, mirroring how the models themselves are generated.
- The plugin contributes its Query via the standard graphql_schema resource, so NetBox merges it into the global schema during ready() - after dynamic models are built.
- custom objects in graphql are namespaces to "custom_objects_" to avoid collisions with other netbox objects as all NetBox graphql objects share a common root namespace.
- Object-level view permissions are enforced (types inherit NetBox's BaseObjectType).

## Field mapping

- Scalars → natural GraphQL types (String, Int, Decimal, Boolean, Date, DateTime, JSON, [String]).
- Object / multi-object (incl. polymorphic) → resolve to the target's native NetBox GraphQL type when one is registered (fully traversable, including nested relations), falling back to a shared CustomObjectRelatedObjectType (id, object_type, display, url) only when the target has no native GraphQL type.
- Base fields: id, display, created, last_updated, tags.

## Live schema & caching

Custom object types can be created, edited, and deleted at runtime, so the GraphQL schema can't be built once at boot like NetBox's. A small view patch swaps in a per-request, per-process schema that tracks the database - runtime changes appear without a NetBox restart, across all worker processes, with no cross-process messaging.

- **Change detection:** each request computes a cheap signature - row counts + max cache_timestamp/last_updated over CustomObjectType/CustomObjectTypeField (two aggregate queries). cache_timestamp is bumped on every structural change, and counts catch adds/removes, so any worker notices a change on its next request.
- **Schema cache:** the assembled schema is cached per process and rebuilt only when the signature changes. The hot path (unchanged) is lock-free; rebuilds are single-flight (one thread rebuilds while others keep serving the current schema instead of blocking).
- **Per-type cache:** each COT's Strawberry type is memoized by (id, cache_timestamp), so a rebuild triggered by one type reuses every other type's already-built definition instead of regenerating all of them.

Monosnap GraphiQL | NetBox 2026-06-09 08-57-09


arthanson:

arthanson:
Claude review comments addressed.

bctiemann (commented):
@arthanson A few items from local review. I think the topmost one or two and some additional tests are worth implementing.

---

### Correctness issues

**1. `csrf_exempt` on `_patched_dispatch` is likely ineffective**

In `__init__.py`:

```python
@csrf_exempt
def _patched_dispatch(self, request, *args, **kwargs):
    ...
NetBoxGraphQLView.dispatch = _patched_dispatch
```

Django's CSRF middleware checks the exemption flag on the *view callable* (the closure returned by `as_view()`), not on the `dispatch` method. That closure was already created and bound into the URL patterns before `ready()` runs. Decorating the method that `dispatch` is replaced with has no effect on the middleware-level CSRF check.

If the original `NetBoxGraphQLView` (or its Strawberry base) already carries `csrf_exempt = True` at the class level, the endpoint works correctly and the decorator here is harmless but misleading. If it doesn't, the decorator doesn't fix it. Either way the comment is inaccurate and the decorator should be removed to avoid confusion.

**2. `decimal.Decimal` annotation — verify Strawberry registration**

`DecimalFieldType.graphql_annotation = decimal.Decimal` uses Python's `decimal.Decimal` (stdlib). Strawberry ships a `Decimal` scalar at `strawberry.scalars.Decimal`; it registers a codec for Python's `decimal.Decimal`, so the stdlib type *should* work as an annotation. But this isn't documented and the behaviour is Strawberry-version-dependent.

Using `strawberry.scalars.Decimal` explicitly (the same pattern used for `JSON` in `JSONFieldType`) would be unambiguous:

```python
# field_types.py
from strawberry.scalars import Decimal as StrawberryDecimal
...
class DecimalFieldType(FieldType):
    graphql_annotation = StrawberryDecimal
```

There's no test that verifies a `Decimal` field round-trips correctly through the GraphQL endpoint, so a silent serialization failure would go undetected.

**3. PR description does not match the implementation**

The description says:

> Object / multi-object (incl. polymorphic) → a shared `CustomObjectRelatedObjectType` (id, object_type, display, url).

The actual implementation resolves relationship fields to the **native NetBox GraphQL type** of their target when one is registered, and only falls back to `CustomObjectRelatedObjectType` when no native type exists. The description should be updated to match (this is actually a *better* feature than the description implies, and the docs in `graphql.md` already describe it correctly).

---

### Design / architecture notes

**4. Process-global rebuild lock serialises cross-branch rebuilds**

The comment in `live.py` correctly explains why `_rebuild_lock` must be global (not per-branch): `build_query_classes()` clears and repopulates the process-global `_type_cache`, so two concurrent rebuilds for different branches would race on it. This is the right design given the shared cache, but it means a busy multi-branch installation serialises all schema rebuilds. If the type cache were per-branch (or isolated within each rebuild call), the lock could be per-branch and concurrency would improve. Not a blocker for this PR, but worth a `# TODO` for future consideration.

**5. `_model_type_registry` is never invalidated**

Once populated, the model-to-type registry is cached for the process lifetime. This is acceptable — NetBox's static types don't change while the server is running — but if a plugin is dynamically installed or unloaded (unusual but possible), the registry would silently serve stale mappings. A brief comment documenting this assumption would make the limitation explicit.

**6. `reset_cache()` replaces the dict reference rather than clearing it**

```python
def reset_cache():
    global _schema_cache
    _schema_cache = {}
```

This is fine for tests (which call it sequentially), but a thread that captured a local reference to the old `_schema_cache` dict before the replacement would continue writing to the stale dict. `_schema_cache.clear()` clears in-place and avoids that. Minor, since `reset_cache()` is test-only, but `.clear()` is the idiomatically safer pattern.

---

### Missing tests

**7. No dedicated test for `_filter_viewable()` with non-superuser + ObjectPermissions**

The permission enforcement in relationship resolvers is implicitly covered by `GraphQLPermissionTestCase`, but there's no test that:
- Creates a custom object with a relationship field pointing at a Site
- Restricts a non-superuser's view permission on Site to a specific subset
- Queries the relationship field and asserts that only the permitted Sites are returned

This is the core of the security claim ("objects reached through a relationship field are likewise filtered to those the user may view"). Worth adding a focused test.

**8. No test for Decimal field round-trip through the endpoint**

`test_scalar_fields_query` covers text, integer, boolean, and select. A `Decimal` (and ideally `JSON`) variant would catch the annotation issue in #2 above.

---

### Minor

**9. `_query_field_name` inner function `_taken` is defined inside the per-call scope**

`_taken` is a closure over `used_names` that is re-created on every call to `_query_field_name`. Inlining the two-condition check directly avoids the overhead and makes the collision logic more readable:

```python
def _query_field_name(custom_object_type, used_names):
    slug = graphql_safe_name(...)
    base = f"{QUERY_FIELD_PREFIX}{slug}"
    name = base
    if name in used_names or f"{name}_list" in used_names:
        name = f"{base}_{custom_object_type.id}"
        ...
```

Small readability improvement, not blocking.

**10. `int` annotation on `IntegerFieldType` vs BigInt**

`IntegerFieldType.graphql_annotation = int` maps to GraphQL's 32-bit signed `Int`. NetBox's built-in integer fields use `BigInt` (64-bit) for primary keys but the standard `Int` for integer custom fields, so this is consistent. However, if a `CustomObjectTypeField` of type `integer` stores a value outside `[-2^31, 2^31)`, Strawberry will silently clamp or error. A comment noting the deliberate choice of 32-bit `Int` (matching NetBox's custom field convention) would prevent a future maintainer from "fixing" it to `BigInt` unnecessarily.

---

### Summary

The architecture is well thought out, particularly the single-flight rebuild, the per-branch schema keying, and the `_filter_viewable` batching. The main asks are:

1. Clarify / remove the ineffective `csrf_exempt` (item 1) — easy fix.
2. Confirm or fix the `Decimal` annotation (item 2) — easy fix.
3. Fix the PR description to match the implementation (item 3) — trivial.
4. Add the relationship-permission test (item 7) — important for security correctness.
5. Add a Decimal round-trip test (item 8) — catches item 2 automatically.

Items 4–6 and 9–10 are at your discretion.

arthanson:
1. Verified, but the decorator is actually necessary. The premise that the view closure is "already created before ready() runs" doesn't hold: _patch_graphql_view() runs in ready() during apps.populate(), while the GraphQL view's as_view() runs only when the root URLconf is imported (lazily, on first request/reverse()) - i.e. after the patch.

Django's as_view() does view.__dict__.update(cls.dispatch.__dict__), copying the csrf_exempt marker from whatever dispatch is at that point.

2. fixed
3. fixed

Added tests for 7, 8

bctiemann (approved):

PR #568, opened 2026-06-09
Fixes #567: Make "Custom object type" field read-only in Add Field flow

### LINKED ISSUE #567: Custom object type changes to `undefined` in the Add Field page
### Plugin Version

0.5.2

### NetBox Version

4.6.2

### Python Version

3.13.5

### Steps to Reproduce

1. Install plugin
2. Add custom object type like "test_object"
3. After creating the custom object type, click **Add Field** button
4. Click **Custom object type** dropdown list

### Expected Behavior

There are all custom object types listed

### Observed Behavior

The dropdown list changes to "undefined"

Image

bctiemann:
## Summary

The **Custom object type** dropdown on the Add Field form showed `undefined` as the selected value when clicked.

## Root Cause

TomSelect (NetBox's select widget) uses `LABEL_FIELD = 'display'` to resolve option labels from API results. `CustomObjectTypeSerializer.Meta.fields` never included `"display"`, so the Custom Object Type API endpoint returned objects without a display value. When TomSelect preloaded options on focus, it overwrote the DOM-parsed label with `undefined`, causing the displayed selection to change to the literal string "undefined".

However, there is no need for this field to be editable in the first place, and it is disabled/read-only in the Edit flow for an existing Custom Object Type Field. The real root cause is that the field should be read-only and inferred from the COT context in the Add flow too.

## Fix

Make `custom_object_type` read-only and inferred from the view context when creating a new Custom Object Type Field.

### DIFF
--- a/netbox_custom_objects/api/serializers.py
+++ b/netbox_custom_objects/api/serializers.py
@@ -326,7 +326,6 @@ class CustomObjectTypeSerializer(NetBoxModelSerializer):
         fields = [
             "id",
             "url",
-            "display",
             "name",
             "verbose_name",
             "verbose_name_plural",
--- a/netbox_custom_objects/api/serializers.py
+++ b/netbox_custom_objects/api/serializers.py
@@ -326,6 +326,7 @@ class CustomObjectTypeSerializer(NetBoxModelSerializer):
         fields = [
             "id",
             "url",
+            "display",
             "name",
             "verbose_name",
             "verbose_name_plural",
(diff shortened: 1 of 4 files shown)


bctiemann:
@claude /review

bctiemann:
The agent review looks like it is keying on the initial commit which added `display` to the serializer, which is no longer the case. Everything else noted in the review looks good.

arthanson (approved):

Issue #569, opened 2026-06-11
Add check framework to gate branching support on required netbox / branching versions


arthanson:
### Plugin Version

v0.6.0

### Proposed functionality

Add a django check framework check to verify minimum levels of netbox and branching when started (if branching is installed)

### Use case

The new branching code requires specific versions of netbox and branching to work properly, but this can't be in pyproject.toml as it only applies if branching is being used.

### External dependencies

N/A

PR #570, opened 2026-06-11
#569 - add check for minimum netbox/branching versions for branching support

arthanson:
### Fixes: #569 

Adds a Django system check that enforces minimum NetBox and netbox-branching versions only when netbox-branching is installed.

### DIFF
--- a/netbox_custom_objects/checks.py
+++ b/netbox_custom_objects/checks.py
@@ -17,7 +17,7 @@ from importlib.metadata import PackageNotFoundError, version as _pkg_version
 
 from django.apps import apps
 from django.conf import settings
-from django.core.checks import Error, register
+from django.core.checks import Error, Warning, register
 from packaging.version import InvalidVersion, Version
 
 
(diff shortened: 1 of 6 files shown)


arthanson:

arthanson:
re claude review, fixed 1 and 2.

netbox_custom_objects/checks.py
```diff
@@ -0,0 +1,75 @@
...
+from importlib.metadata import PackageNotFoundError, version as _pkg_version
+
+from django.apps import apps
+from django.conf import settings
+from django.core.checks import Error, Warning, register
+from packaging.version import InvalidVersion, Version
+
+
+# Version floors enforced only when netbox-branching is installed.
+REQUIRED_NETBOX_VERSION = '4.6.2'
```
bctiemann: Should this one pull from the min_version on the PluginConfig? Or else we'll have to keep this in sync too. Or if this is relevant only when branching is present, maybe `REQUIRED_NETBOX_VERSION_FOR_BRANCHING` or something? (If it's truly unrelated to the custom-objects `min_version` then I don't mind this.)

bctiemann (changes requested):
See inline

netbox_custom_objects/checks.py
arthanson: changed

PR #571, opened 2026-06-11
Closes #286: Add ContactsMixin to CustomObject

### LINKED ISSUE #286: Toggleable Contacts
### Plugin Version

0.4.1	

### Proposed functionality

Option to Enable Contacts for Custom Objects

### Use case

we have the idea to Manage the Applications with the ports they are providing in Netbox and it would be also very usefull to add the Contacts of the responsible Person there

### External dependencies

_No response_

bctiemann:
## Summary

Adds `ContactsMixin` to the `CustomObject` base class, giving every Custom Object Type a Contacts tab and full contact-assignment support (UI + API) — the same experience as first-class NetBox models like Device and Site.

## Implementation

Two-line change: import `ContactsMixin` from `netbox.models.features` (same module that already provides the other 10 mixins) and add it to `CustomObject`'s base class list.

`ContactsMixin` contributes only a `GenericRelation` (no DB column), so:
- No migration is required
- The mixin-drift detection in `mixin_migration.py` is unaffected
- NetBox's `register_models()` auto-wires the Contacts view for every COT via the existing `issubclass(model, ContactsMixin)` check in `features.py`

## Test plan

- [ ] Create a Custom Object Type and verify a Contacts tab appears on the object detail page
- [ ] Assign a contact to a custom object via the UI and via the API
- [ ] Confirm `upgrade_custom_objects` reports no drift (no new columns to add)


bctiemann:
@claude /review

arthanson (changes requested):
There is no Contacts tab appearing on a custom object, I think the view needs to be updated as well.

bctiemann:
Screenshot 2026-06-15 at 7 28 56 PM

Done.

PR #573, opened 2026-06-11
Fixes #572: Use apps.is_installed() to guard netbox-branching import in live.py

### LINKED ISSUE #572: Crash on startup when netbox-branching is installed but not in PLUGINS
### Plugin Version

feature branch (post-0.5.1)

### Steps to Reproduce

1. Install `netbox-branching` into the Python environment but do **not** add it to `PLUGINS` in `configuration.py`
2. Start NetBox (`python manage.py runserver`)

### Expected Behavior

NetBox starts normally. The plugin detects that netbox-branching is not active and skips the branch-eviction signal registration.

### Observed Behavior

Startup fails with:

```
RuntimeError: Model class netbox_branching.models.changes.ObjectChange doesn't declare an explicit app_label and isn't in an application in INSTALLED_APPS.
```

The crash originates in `connect_signature_invalidation()` in `netbox_custom_objects/graphql/live.py`, which guards the `from netbox_branching.models import Branch` import with `except ImportError`. However, when the package is installed but not in `INSTALLED_APPS`, Django raises `RuntimeError` (not `ImportError`) because the app registry is not populated for that package.

### Fix

Replace the `ImportError` guard with `apps.is_installed('netbox_branching')`, which correctly handles both "not installed" and "installed but not enabled" cases.

bctiemann:
## Summary

`connect_signature_invalidation()` guarded the `from netbox_branching.models import Branch` import with `except ImportError`. When netbox-branching is installed in the Python environment but absent from `PLUGINS`/`INSTALLED_APPS`, Django raises `RuntimeError` (not `ImportError`) during model class creation because the app registry is unpopulated for that package — causing a startup crash.

**Fix:** replace the `ImportError` guard with `apps.is_installed('netbox_branching')`, which correctly handles both "package not installed" and "package installed but not enabled" without attempting to import Django models for a non-registered app.

## Test plan

- [ ] Install netbox-branching into the venv but omit it from `PLUGINS` — server starts cleanly
- [ ] With netbox-branching in `PLUGINS` — branch schema eviction still works as before

### DIFF
--- a/netbox_custom_objects/graphql/live.py
+++ b/netbox_custom_objects/graphql/live.py
@@ -206,11 +206,11 @@ def connect_signature_invalidation():
             )
 
     # Evict a branch's cached schema when the branch itself is deleted.  No-op when
-    # netbox-branching is not installed (there are then no branch-keyed entries).
-    try:
-        from netbox_branching.models import Branch
-    except ImportError:
+    # netbox-branching is not installed or not in INSTALLED_APPS.
+    from django.apps import apps as django_apps
+    if not django_apps.is_installed('netbox_branching'):
         return
+    from netbox_branching.models import Branch
     post_delete.connect(
         _evict_branch_schema,
         sender=Branch,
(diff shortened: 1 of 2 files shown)


bctiemann:

bctiemann:
All three issues raised by review have been addressed.

arthanson (approved):

PR #575, opened 2026-06-12
Fixes #561: Register <field>_id filter for non-polymorphic FK Object fields

### LINKED ISSUE #561: Related Objects link uses ?<field>_id= but custom object filterset only registers ?<field>=
**Plugin Version**

`0.5.1` (`main`, commit `dae3086`)

**NetBox Version**

`v4.6.2`

**Python Version**

`3.12`

**Steps to Reproduce**

1. Create a Custom Object Type with an Object field linked to `tenancy.Tenant`.
2. Create two custom objects of that type, each pointing at a different tenant.
3. Open the detail page for one of those tenants.
4. In the Related Objects panel, click the count for the custom object type.
5. Note the list view opens with `?tenant_id=N` in the URL.

**Expected Behavior**

The list shows only the custom objects whose tenant field points at tenant `N`.

**Observed Behavior**

The list shows every custom object of that type, unfiltered, despite `?tenant_id=N` being present in the URL.

Workaround: dropping the `_id` suffix (`?tenant=N`) filters correctly.

bctiemann:
## Summary

NetBox's Related Objects panel generates links with `?<field>_id=N` (e.g. `?tenant_id=5`), but the dynamic filterset for custom objects only registered the `?<field>=N` form (which takes a model instance). Clicking through from a tenant's Related Objects panel to the custom object list would silently return all objects unfiltered, despite the correct filter being present in the URL.

## Root cause

`build_filter_for_field()` returned only `{field.name: NonPolymorphicObjectFilter(...)}` for non-polymorphic `TYPE_OBJECT` fields. `NonPolymorphicObjectFilter` uses a `ModelChoiceField` which expects a PK and resolves it to an instance — the right input for the filter form — but the `_id` URL param produces a raw integer that the instance-based filter never sees.

## Fix

A new `NonPolymorphicObjectIdFilter` accepts a raw integer PK and filters directly by `<field_name>_id=`. `build_filter_for_field()` now emits it alongside the existing instance-based filter for every non-polymorphic `TYPE_OBJECT` field.

The `_id` suffix is the standard NetBox convention (per AGENTS.md: "FK filters must declare an explicit `<field>_id` variant").

## Tests

Two regression tests added to `ObjectFieldFiltersetTestCase`:
- `test_filter_by_id_suffix_returns_matching_object`
- `test_filter_by_id_suffix_different_value`

### DIFF
--- a/netbox_custom_objects/filtersets.py
+++ b/netbox_custom_objects/filtersets.py
@@ -19,6 +19,7 @@ __all__ = (
     "CustomObjectTypeFilterSet",
     "NonPolymorphicMultiObjectFilter",
     "NonPolymorphicObjectFilter",
+    "NonPolymorphicObjectIdFilter",
     "PolymorphicMultiObjectFilter",
     "PolymorphicObjectFilter",
     "get_filterset_class",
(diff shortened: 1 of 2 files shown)


bctiemann:

pheus (approved):
Looks good to me! Thanks!

PR #585, opened 2026-06-16
Closes #254: Quick Add support for custom object object/multiobject fields

### LINKED ISSUE #254: Quick Add functionality
### Plugin Version

0.4.0

### Proposed functionality

Implement and enable the Quick Add widget for object and multiobject field types pointing to Custom Objects.

### Use case

This functionality would bring Custom Objects into consistency with all other object selection fields in NetBox.

### External dependencies

_No response_

bctiemann:
## Summary

Adds Quick Add (`+`) button support to `object` and `multiobject` fields whose target is a Custom Object Type, bringing them into consistency with all other object selection fields in NetBox.

Screenshot 2026-06-16 at 6 19 07 PM
Screenshot 2026-06-16 at 6 17 14 PM

## How it works (entirely within the plugin — no core changes needed)

**`field_types.py`** — After constructing the `DynamicModelChoiceField` / `DynamicModelMultipleChoiceField` for a COT target, sets `widget.quick_add_context` directly on the widget. The standard path through `DynamicModelChoiceMixin.get_bound_field()` calls `get_action_url(model, action='add')`, which cannot resolve custom object URLs (they require a `custom_object_type` slug parameter). Setting the context directly bypasses that call entirely.

**`views.py`** — `CustomObjectEditView.get()` and `post()` are overridden for the `_quickadd` path. The core `htmx/quick_add.html` uses `{% action_url model 'add' %}` which fails for dynamically-generated COT models. The overrides use a plugin-local template instead. Non-quick-add requests delegate to `super()` unchanged. The success path (`htmx/quick_add_created.html`) works as-is — it only needs `object.pk` and `str(object)`.

**`templates/netbox_custom_objects/htmx/co_quick_add.html`** — mirrors the core `htmx/quick_add.html` but replaces `{% action_url model 'add' %}` with `{% url 'plugins:netbox_custom_objects:customobject_add' model.custom_object_type.slug %}`.

## Test plan

- [ ] Create two COTs (`Foo` and `Bar`), add an `object` field on `Bar` pointing at `Foo`
- [ ] Open the `Bar` add form — verify the `+` button appears next to the `Foo` field
- [ ] Click `+`, fill in the quick-add form, submit — verify the new `Foo` is created and auto-selected
- [ ] Repeat with a `multiobject` field
- [ ] Verify validation errors in the quick-add modal re-render correctly (required fields missing, etc.)

### DIFF
--- a/netbox_custom_objects/field_types.py
+++ b/netbox_custom_objects/field_types.py
@@ -21,6 +21,7 @@ from django.db.models.manager import Manager
 from django.db.models.signals import m2m_changed
 from django.utils.html import escape
 from django.utils.safestring import mark_safe
+from django.urls import reverse
 from django.utils.translation import gettext_lazy as _
 from extras.choices import CustomFieldTypeChoices, CustomFieldUIEditableChoices
 from utilities.api import get_serializer_for_model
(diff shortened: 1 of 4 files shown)


bctiemann:
@claude /review

bctiemann:
Added tests and fixed the minor issues (not 4, the one deemed unnecessary).

arthanson (approved):
nice!

PR #586, opened 2026-06-17
Closes #268: Add display_expression for composite custom object names

### LINKED ISSUE #268: Implement customizable primary name field
### Plugin Version

0.4.0

### Proposed functionality

Implement a function to customize the primary name field. At the moment it's only possible to set one field as primary name field.

Implement a new field in the custom object to define the primary name field with a Jinja2 Expression, so it's possible to build every primary name field which you want. You can use the value of the fields with Jinja2 Expressions.

Example:

Custom Object: SFP Module

Fields:

- Serial
- Manufacturer
- Name

Primary Name Field: "{{Name}} - {{Manufacturer}}"

 **Object 1:**

- Serial: 12345
- Manufacturer: Cisco
- Name: SFP-10G-LR

_**Final Primay Name: SFP-10G-LR - Cisco**_

 **Object 2:**

- Serial: 122445
- Manufacturer: Arista
- Name: SFP-10G-LR

_**Final Primay Name: SFP-10G-LR - Arista**_

### Use case

The primary name field than can be customized in more detail.

### External dependencies

_No response_

bctiemann:
## Summary

Adds a `display_expression` Jinja2 template field to `CustomObjectType` that lets administrators compose the object display name from multiple field values, replacing the single-field `primary` marker approach when set.

**Example:** a COT with `make`, `model`, and `serial` fields can use `{{ make }} - {{ model }}` as its display expression so objects display as `Cisco - ASR1001` instead of just the primary field value.

## How it works

- `display_expression` is a `CharField(max_length=500, blank=True)` on `CustomObjectType`
- `CustomObject.__str__()` calls a new `_render_display_expression()` helper that renders the expression via a Jinja2 `SandboxedEnvironment` with all field values as context
- Any rendering error (syntax error, etc.) silently falls through to the existing single-primary-field fallback
- Fields referenced in the expression that don't exist on the object resolve to an empty string (via `jinja2.Undefined`)
- If the rendered result is empty, the fallback is also used

## Form layout

The form now has a **Display** fieldset grouping `verbose_name`, `verbose_name_plural`, `display_expression`, and `group_name` — all fields that control how the type appears to users — positioned between the Name field and the URL path/slug field.

Screenshot 2026-06-17 at 11 19 00 AM

## Surface area

- Model field + migration 0015
- `_render_display_expression()` helper (separately testable/mockable)
- Form: Display fieldset
- Serializer: `display_expression` in `CustomObjectTypeSerializer`
- COT detail template: shows expression as `<code>` when set

## Test plan

- [ ] Create a COT with `make` and `model` text fields; set `display_expression = {{ make }} - {{ model }}`; verify objects display as `Make - Model`
- [ ] Verify that clearing the expression falls back to the primary-field display name
- [ ] Verify a broken expression (e.g. `{% bad %}`) falls back silently


bctiemann:
@claude /review

netbox_custom_objects/migrations/0015_customobjecttype_display_expression.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 6.0.6 on 2026-06-17 14:54
+
```
jnovinger:
```suggestion
```

netbox_custom_objects/models.py
```diff
@@ -78,6 +82,12 @@
 logger = logging.getLogger(__name__)
 
 
+@functools.lru_cache(maxsize=256)
+def _compile_display_template(expression: str):
+    """Compile a Jinja2 display expression once and cache by expression string."""
+    return _JinjaSandbox(undefined=_JinjaUndefined).from_string(expression)
```
jnovinger: Should we perhaps be using some of the Jinja related tools in core (e.g. `utilities/jinja2.py::render_jinja2(template_code, context)`), rather than building our own here?

I realize they're not necessarily public API, but they are more hardened and battle tested.

netbox_custom_objects/models.py
```diff
@@ -82,12 +79,6 @@
 logger = logging.getLogger(__name__)
 
 
-@functools.lru_cache(maxsize=256)
-def _compile_display_template(expression: str):
-    """Compile a Jinja2 display expression once and cache by expression string."""
-    return _JinjaSandbox(undefined=_JinjaUndefined).from_string(expression)
-
```
jnovinger: Apologies, just realized that using `render_jinja2()` effectively neuters the ability to use a LRU cache here.

Has this been a problem in practice? Perhaps we could propose a cache to the part of the helper in core that is responsible for initializing the template.

netbox_custom_objects/models.py
```diff
@@ -825,7 +814,7 @@ def _render_display_expression(self):
                     ctx[field_name] = '' if value is None else value
                 except Exception:  # noqa: BLE001
                     ctx[field_name] = ''
-            rendered = _compile_display_template(expression).render(**ctx).strip()
+            rendered = render_jinja2(expression, ctx).strip()
```
jnovinger: One side effect of the `render_jinja2` switch worth flagging: operator-configured `JINJA2_FILTERS` from `settings.JINJA2_FILTERS` are now available in display expressions at runtime, since `render_jinja2` registers them on the environment.

The form validator in `clean_display_expression` uses a plain `SandboxedEnvironment().parse()` with no filters registered, so a display expression referencing a custom filter will pass validation but fail silently at render time and fall back to the primary field. Before this commit, the hand-rolled environment didn't pick up `JINJA2_FILTERS` at all, so this is a new footgun.

Apologies for suggesting the switch to `render_jinja2` without thinking through this fully.

bctiemann: Done.

jnovinger (approved):

PR #590, opened 2026-06-22
Fixes #587: Use actual FK target column in migration 0011

### LINKED ISSUE #587: Custom Objects migration 0011_non_deferrable_fk_constraints fails on multi-table-inheritance COTs
### Plugin Version

0.5.2

### NetBox Version

4.5.10

### Python Version

3.12

### Steps to Reproduce

1. On a pre-0.5.x install, create a Custom Object Type whose backing table uses Django multi-table inheritance, so its primary key column is customobject_ptr_id rather than id.
2. Add one or more many-to-many fields to that COT. Each creates an M2M through table with a source_id FK pointing back at the COT table's PK. Django creates these FKs DEFERRABLE by default.
3. Run the upgrade that applies netbox_custom_objects migrations through 0011 (NetBox 4.5.8 → 4.5.10). manage.py migrate.

### Expected Behavior

Migration 0011 converts each DEFERRABLE FK on custom_objects_* tables to non-DEFERRABLE referencing its actual target column, and the migration chain (0011→0014) completes. The upgrade succeeds.

### Observed Behavior

0011 aborts with:
psycopg.errors.UndefinedColumn: column "id" referenced in foreign key constraint does not exist
django.db.utils.ProgrammingError: column "id" referenced in foreign key constraint does not exist

Applying netbox_custom_objects.0011_non_deferrable_fk_constraints...
File ".../migrations/0011_non_deferrable_fk_constraints.py", line 76, in fix_deferrable_fk_constraints
The migration rolls back atomically (DB left at 0010), manage.py migrate exits non-zero, and upgrade cannot complete.

bctiemann:
## Summary

Migration `0011_non_deferrable_fk_constraints` reconstructs DEFERRABLE FK constraints as non-DEFERRABLE. The `REFERENCES` clause hardcoded `"id"` as the target column, but COTs created under multi-table inheritance use `customobject_ptr_id` as their primary key. PostgreSQL rejects the constraint because the column `id` literally does not exist on those tables.

The fix: the query already joins `information_schema.constraint_column_usage`, which records the actual referenced column name. Adding `ccu.column_name AS foreign_column_name` to the SELECT and using it in the `REFERENCES` clause makes the migration correct for both standard (`id`) and MTI (`customobject_ptr_id`) COTs.

## Local verification

Verified with a direct DB script that:
1. Creates an MTI-style `custom_objects_test_*` table (PK=`customobject_ptr_id`) and a through table with a DEFERRABLE FK pointing at it
2. Runs the **original** migration logic → **confirmed** `ProgrammingError: column "id" referenced in foreign key constraint does not exist`
3. Restores the DEFERRABLE constraint, runs the **fixed** migration logic → **succeeds**, constraint rebuilt as non-DEFERRABLE referencing `customobject_ptr_id`

## Test plan

- [x] Bug reproduced and fix verified locally via DB-level script
- [ ] Confirm on a real pre-0.5.x install with an MTI COT that has M2M fields

### DIFF
--- a/netbox_custom_objects/migrations/0011_non_deferrable_fk_constraints.py
+++ b/netbox_custom_objects/migrations/0011_non_deferrable_fk_constraints.py
@@ -42,7 +42,8 @@ def fix_deferrable_fk_constraints(apps, schema_editor):
                 tc.table_name,
                 tc.constraint_name,
                 kcu.column_name,
-                ccu.table_name AS foreign_table_name
+                ccu.table_name AS foreign_table_name,
+                ccu.column_name AS foreign_column_name
             FROM information_schema.table_constraints AS tc
             JOIN information_schema.key_column_usage AS kcu
                 ON tc.constraint_name = kcu.constraint_name
@@ -60,7 +61,7 @@ def fix_deferrable_fk_constraints(apps, schema_editor):
         """)
         rows = cursor.fetchall()
 
-        for table_name, constraint_name, column_name, foreign_table in rows:
+        for table_name, constraint_name, column_name, foreign_table, foreign_column in rows:
             new_constraint_name = _safe_constraint_name(table_name, column_name)
             # IF EXISTS on the old name handles partial re-runs where the old constraint
             # was already dropped in a previous (failed) attempt.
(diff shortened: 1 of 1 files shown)


bctiemann:

arthanson (approved):

PR #574, opened 2026-06-12
Closes #36: Add changelog_enabled toggle to CustomObjectType

bctiemann:
## Summary

Adds a `changelog_enabled` boolean field (default `True`) to `CustomObjectType`. When disabled, changes to objects of that type are silently skipped by NetBox's change-logging signal — no `ObjectChange` rows are written, and the Changelog tab is hidden from the custom object detail page.

## How it works

The implementation is confined to two points:

1. **Model field** — `CustomObjectType.changelog_enabled` (migration 0015, default `True`).

2. **Generated model override** — `get_model()` injects a `to_objectchange()` method that returns `None` when the flag is off. NetBox's `handle_changed_object` signal guards on `objectchange and objectchange.has_changes` before saving, so `None` is the documented way to suppress a change record — no signal disconnection, no monkey-patching of the signal itself.

## Surface area

- Form: new **Options** fieldset with the `changelog_enabled` toggle
- Serializer: field included in `CustomObjectTypeSerializer`
- Table: field available as an optional column
- COT detail template: status shown with a checkmark
- CO detail template: Changelog tab is hidden when `changelog_enabled=False`; direct URL access to the changelog view returns 404

## Notes

- Branching compatibility: if `changelog_enabled=False`, objects of that type also cannot participate in branching (branching requires change capture). This is consistent with the issue's statement that "this user would be fine with these changes not being branchable."
- The per-COT toggle was chosen over a global flag because the use case is inherently per-type (some types warrant an audit trail, others don't).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #591, opened 2026-06-23
Closes #583: Add branching CI matrix job and test_branching.py

### LINKED ISSUE #583: Add CI matrix job with netbox-branching enabled so branching tests are not permanently skipped
### Plugin Version

0.5.1 (feature branch)

### Proposed functionality

Add a third CI matrix entry to `lint-tests.yaml` that installs netbox-branching and runs the full test suite with branching enabled. This would close the current gap where all `@unittest.skipUnless(HAS_BRANCHING, ...)` tests are silently skipped in every CI run.

The proposed change to `.github/workflows/lint-tests.yaml`:

- Add a `branching` entry alongside the existing `main` and `feature` matrix entries
- Install netbox-branching from its repository in this matrix job
- Use `testing/configuration_branching.py` (wraps `DATABASES` in `DynamicSchemaDict` and adds `BranchAwareRouter`) instead of `testing/configuration.py`
- Mark the job `continue-on-error: true` initially so branching CI failures are visible but non-blocking while the pipeline stabilises

### Use case

`test_branching.py` contains an extensive suite of `TransactionTestCase`-based tests covering branch sync, merge, revert, and cross-COT lifecycle scenarios. These tests currently only run when a developer proactively executes them locally with a branching-enabled configuration. A regression in branching support — even one introduced by a change that doesn't obviously touch branching-related code — goes undetected until noticed manually.

### External dependencies

Requires netbox-branching to be installable in the CI environment. If netbox-branching is not yet on PyPI, the workflow would install it from its Git repository.

bctiemann:
## Summary

- Adds `testing/configuration_branching.py`: wraps `DATABASES` in `DynamicSchemaDict` and registers `BranchAwareRouter`, satisfying netbox-branching's startup requirements for the test environment.
- Adds `netbox_custom_objects/tests/test_branching.py` (carried over from the `branching-check` branch): 2800-line `TransactionTestCase`-based suite covering branch sync, merge, revert, and cross-COT lifecycle scenarios. All tests gate on `HAS_BRANCHING` so the standard `main`/`feature` CI jobs are unaffected.
- Extends `.github/workflows/lint-tests.yaml` with a third matrix entry (`netbox-ref: main, with-branching: true`) that installs `netboxlabs-netbox-branching>=1.0.4,<2.0.0` and runs `test_branching.py` against `configuration_branching.py`. The job intentionally runs only `test_branching.py` (not the full suite) — the existing `main` and `feature` jobs cover the full `netbox_custom_objects.tests` suite, and the branching job's purpose is specifically to validate branching integration.
- Fixes two pre-existing test failures: bigint assertion in `test_field_type_change_text_to_integer_merge`, and squash dep-graph null `model_class` crash in the self-referential M2M test.
- Fixes iterative merge FK violation for self-referential MULTIOBJECT fields by creating the through table's target FK as `DEFERRABLE INITIALLY DEFERRED` and immediately re-deferring it by name (overriding `SET CONSTRAINTS ALL IMMEDIATE` called earlier in the same transaction by `_schema_add_field`).

## Test plan

- [x] Confirm the new `tests (main, branching)` CI job runs and branching tests pass
- [x] Confirm the existing `tests (main)` and `tests (feature)` jobs are unaffected


bctiemann:
Blocked by https://github.com/netboxlabs/netbox-custom-objects/pull/574

bctiemann:

bctiemann:
Thanks for the review. Here's how each point was addressed:

**1 & 2 (CI timeout / `continue-on-error`):** Both were intentional post-stabilization changes, not accidental drops. The branching suite completes in ~14 minutes in CI so 20m is sufficient, and `continue-on-error: true` was removed because the suite is now stable (FK deferral fix + lock-timeout skip). PR description updated to clarify this.

**3 (branching job runs only `test_branching.py`):** Left as-is by design. The `main` and `feature` jobs already run the full `netbox_custom_objects.tests` suite; the branching job's purpose is specifically to validate branching integration. Running the full suite in the branching job would roughly double execution time with negligible additional coverage — any regressions in `test_api`, `test_views`, etc. are caught by the existing non-branching jobs.

**4 (`pg_terminate_backend` too broad):** Fixed in ad50806. Added `AND backend_type = 'client backend'` to match the filter already used in the polling query below it, preventing background workers (autovacuum, WAL writer, etc.) from being terminated. Comment updated to note the local-dev implication (all client connections to the test DB are terminated during teardown).

**5 (`test_sequential_renames_both_sides_sync` skip):** Accepted as-is. The test covers sync conflict-resolution under load; converting a flaky lock-timeout ERROR to a SKIP is the right tradeoff — a real regression in the sync path would surface as an assertion failure before reaching the `sync()` call.

**6 (`BranchingTestBase` MRO footgun):** Accepted as-is. All existing subclasses follow the pattern correctly, and the `_TestBase` sentinel already makes the requirement visually obvious at each class definition.

**7 (version pin upper bound):** Accepted as-is. `<2.0.0` is standard semver practice; if a 2.x release happens the pin will need updating, but that's a routine maintenance task.

.github/workflows/lint-tests.yaml
```diff
@@ -80,13 +88,31 @@ jobs:
...
+            -c "ALTER SYSTEM SET checkpoint_timeout = '30min'" \
+            -c "ALTER SYSTEM SET max_wal_size = '2GB'" \
+            -c "ALTER SYSTEM SET synchronous_commit = off" \
+            -c "ALTER SYSTEM SET lock_timeout = '60s'" \
+            -c "SELECT pg_reload_conf()"
       - name: Run tests
         working-directory: netbox
         run: |
-          python netbox/manage.py test netbox_custom_objects.tests --keepdb
+          python netbox/manage.py test ${{ matrix.with-branching == true && 'netbox_custom_objects.tests.test_branching' || 'netbox_custom_objects.tests' }} --keepdb --verbosity=2
```
pheus:
```suggestion
      - name: Run tests
        if: ${{ matrix.with-branching != true }}
        working-directory: netbox
        run: |
          python netbox/manage.py test netbox_custom_objects.tests --verbosity=2

      - name: Run tests with branching
        if: ${{ matrix.with-branching == true }}
        working-directory: netbox
        run: |
          python netbox/manage.py test netbox_custom_objects.tests.test_branching --verbosity=2
```

Could we split the branching and non-branching test runs into separate steps?

The conditional expression in the command works, but separate steps make the CI output easier to scan and make it clearer which test suite is being executed for each matrix entry.

netbox_custom_objects/field_types.py
```diff
@@ -1680,6 +1687,38 @@ def create_m2m_table(self, instance, model, field_name, schema_conn=None):
...
+                    #   1. Add the constraint as DEFERRABLE INITIALLY DEFERRED
+                    #   2. Immediately re-defer it by name so it is DEFERRED
+                    #      for the rest of the merge transaction
+                    # This lets iterative branch merges (time-ordered) insert
+                    # through rows before the referenced target CO exists; the
+                    # FK check is deferred to transaction commit, by which
+                    # point all CO CREATEs have been applied.
+                    to_table = to_model._meta.db_table
+                    to_pk = to_model._meta.pk.column
+                    fk_conname = (table_name[:53] + '_target_fk').lower()
```
pheus: **nit:** Could we make the generated FK constraint name collision-resistant?

`fk_conname = (table_name[:53] + '_target_fk').lower()` is probably fine for normal table names, but two long through-table names with the same first 53 characters would generate the same constraint name. Since we later call `SET CONSTRAINTS <name> DEFERRED`, using a short hash/digest suffix would make this more robust and easier to reason about.

For example, something like `<truncated_table>_<digest>_target_fk` would preserve readability while avoiding collisions.

.github/workflows/lint-tests.yaml
```diff
@@ -80,13 +88,31 @@ jobs:
...
-
           python -m pip install --upgrade pip
           pip install -r requirements.txt -U
+      - name: Install dependencies & configure plugin (with branching)
+        if: ${{ matrix.with-branching == true }}
+        working-directory: netbox
+        run: |
+          ln -s $(pwd)/../netbox-custom-objects/testing/configuration_branching.py netbox/netbox/configuration.py
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt -U
```
pheus:
```suggestion
          pip install -r requirements.txt
```
Could we avoid upgrading NetBox's pinned requirements here?
We generally want CI to test against the dependency versions defined by NetBox's `requirements.txt`, rather than asking pip to upgrade packages to newer allowed versions. This keeps the test environment closer to the NetBox checkout being tested.

.github/workflows/lint-tests.yaml
```diff
@@ -80,13 +88,31 @@ jobs:
           pip install .
           pip install .[test]
       - name: Install dependencies & configure plugin
+        if: ${{ !matrix.with-branching }}
         working-directory: netbox
         run: |
           ln -s $(pwd)/../netbox-custom-objects/testing/configuration.py netbox/netbox/configuration.py
-
           python -m pip install --upgrade pip
           pip install -r requirements.txt -U
```
pheus:
```suggestion
          pip install -r requirements.txt
```
Same here

pheus (changes requested):
Thanks for working on this! The overall direction looks good to me.

I left a couple of smaller inline suggestions, mostly around keeping CI aligned with NetBox’s pinned dependency set and making the branching/non-branching test steps easier to read.

One question I had is about the upgrade path for existing direct M2M through tables. From my reading, the new deferrable FK is added only when the through table is newly created (`table_name not in tables`). Since CI starts from a fresh database, it exercises that path, but it may not cover installations that already have direct M2M fields and therefore existing through tables with the previous immediate FK constraint.

Is that upgrade case intentionally out of scope for this PR, or should we add an idempotent repair path / follow-up issue for existing installations?

bctiemann (commented):
Thanks for the thorough review, Martin — all three inline suggestions addressed in the latest commit.

On your question about the upgrade path for existing through tables: you're right that the deferrable FK is only added when the through table is *newly created* (). Installations that already have a through table with the old non-deferrable FK constraint don't get repaired by this PR.

My read is that this is intentionally limited in scope for now — the immediate goal was making branch merges work for *new* M2M fields created after this change. Retrofitting existing through tables would require detecting the presence/type of the existing FK constraint and issuing an  in an idempotent way, which is meaningful additional complexity and risk.

I'll open a follow-up issue to track the repair path for existing installations so it doesn't get lost.

bctiemann (commented):
Thanks for the thorough review, Martin — all three inline suggestions addressed in the latest commit.

On your question about the upgrade path for existing through tables: you are right that the deferrable FK is only added when the through table is newly created (`table_name not in tables`). Installations that already have a through table with the old non-deferrable FK constraint do not get repaired by this PR.

This is intentionally limited in scope for now — the immediate goal was making branch merges work for new M2M fields created after this change. Retrofitting existing through tables would require detecting the presence and type of the existing FK constraint and issuing an idempotent ALTER TABLE ... DROP CONSTRAINT / ADD CONSTRAINT, which is meaningful additional complexity and risk.

I will open a follow-up issue to track the repair path for existing installations so it does not get lost.

arthanson (approved):

pheus (approved):

PR #596, opened 2026-06-24
Fixes #595: SET CONSTRAINTS before poly column removal on revert

### LINKED ISSUE #595: Reverting a merged branch fails for COTs with a polymorphic object field
### Plugin Version

v0.6.0

### NetBox Version

v4.5

### Python Version

v3.13

### Steps to Reproduce

1. Create branch b1, provision it, and activate it (activate_branch(b1)).
2. Inside b1, create a COT, e.g. repro.
3. Add fields to the COT:
    - name — text, primary.
    - ref — object, polymorphic, pointing to dcim.site. (Gives a polymorphic column on the COT's table for revert to drop. Target type is irrelevant.)
  - sites — multiobject, pointing to dcim.site. (Creates the through table whose source_id → custom_objects_<id> FK is DEFERRABLE INITIALLY DEFERRED — the thing that queues pending trigger events. Target type is irrelevant.)
4. Create one instance of the COT with ref = a Site and sites = [a Site] (the multiobject must be populated so the row delete hits the deferred FK).
5. Merge b1 → main. ✅ Succeeds (forward order adds columns before inserting rows).
6. Revert the merge. ❌ Fails:
  ObjectInUse: cannot ALTER TABLE "custom_objects_<id>" because it has pending trigger events
6. (surfaced downstream as current transaction is aborted, commands ignored…)

script to repro:
```
"""Minimal repro: polymorphic-object field + netbox-branching revert bug.

Reverting a merged branch that created a COT with a polymorphic OBJECT field
fails:
  ObjectInUse: cannot ALTER TABLE "custom_objects_<id>" because it has pending
  trigger events   (masked downstream as 'current transaction is aborted')

Cause: remove_polymorphic_object_columns() drops the poly columns with ALTER
TABLE but omits 'SET CONSTRAINTS ALL IMMEDIATE' first (the scalar-field path in
models._schema_remove_field does it). On revert the row delete queues deferred
FK trigger events, so the column drop is rejected.

Run: manage.py polyrevertrepro
"""
import time
import traceback
import uuid

from django.core.management.base import BaseCommand, CommandError
from django.test import RequestFactory
from django.urls import reverse


class Command(BaseCommand):
    help = "Minimal repro of the polymorphic-object field branching revert bug."

    def handle(self, *a, **o):
        from netbox.context_managers import event_tracking
        from netbox_branching.choices import BranchStatusChoices
        from netbox_branching.models import Branch
        from netbox_branching.utilities import activate_branch
        from django.contrib.auth import get_user_model
        from core.models import ObjectType
        from dcim.models import Site
        from extras.choices import CustomFieldTypeChoices as F
        from netbox_custom_objects.models import CustomObjectType, CustomObjectTypeField

        user, _ = get_user_model().objects.get_or_create(
            username='polyrepro_user', defaults={'is_superuser': True, 'is_active': True})
        site_ot = ObjectType.objects.get(app_label='dcim', model='site')

        def req():
            r = RequestFactory().get(reverse('home')); r.id = uuid.uuid4(); r.user = user
            return r

        Branch.objects.filter(name='polyrepro').delete()
        CustomObjectType.objects.filter(slug__startswith='polyrepro-').delete()
        with event_tracking(req()):
            site, _ = Site.objects.get_or_create(slug='polyrepro-site', defaults={'name': 'PolyRepro Site'})

        # 1. create + activate a branch
        branch = Branch(name='polyrepro'); branch.save(provision=False); branch.provision(user=user)
        deadline = time.time() + 180
        while time.time() < deadline:
            branch.refresh_from_db()
            if branch.status == BranchStatusChoices.READY:
                break
            time.sleep(0.5)
        if branch.status != BranchStatusChoices.READY:
            raise CommandError(f'branch not ready: {branch.status}')
        self.stdout.write(f'branch ready: {branch.schema_name}')

        # 2. inside the branch: COT with a polymorphic OBJECT field + a MULTIOBJECT
        #    field (the multiobject through's deferred FK is what queues the
        #    pending trigger events when the row is deleted on revert).
        with activate_branch(branch), event_tracking(req()):
            cot = CustomObjectType.objects.create(name='polyrepro', slug='polyrepro-a')
            CustomObjectTypeField.objects.create(custom_object_type=cot, name='name', type=F.TYPE_TEXT, primary=True)
            ref = CustomObjectTypeField.objects.create(
                custom_object_type=cot, name='ref', type=F.TYPE_OBJECT, is_polymorphic=True)
            ref.related_object_types.set([site_ot])
            CustomObjectTypeField.objects.create(
                custom_object_type=cot, name='sites', type=F.TYPE_MULTIOBJECT, related_object_type=site_ot)

            model = cot.get_model()
            row = model.objects.create(name='row1', ref=site)
            row.sites.set([site])
        self.stdout.write('created COT (poly object + multiobject) + 1 row in branch')

        # 3. merge (succeeds)
        branch.merge(user=user, commit=True)
        self.stdout.write('merged OK')

        # 4. revert (fails with pending-trigger-events)
        try:
            branch.revert(user=user, commit=True)
            self.stdout.write('revert OK (bug fixed?)')
        except Exception as exc:
            self.stdout.write(f'revert FAILED: {type(exc).__name__}: {str(exc).splitlines()[0]}')
            traceback.print_exc(file=self.stdout)

        try:
            branch.refresh_from_db(); branch.delete()
        except Exception:
            pass
        CustomObjectType.objects.filter(slug__startswith='polyrepro-').delete()
        Site.objects.filter(slug='polyrepro-site').delete()
```

### Expected Behavior

Revert should succeed

### Observed Behavior

Revert fails with error.

bctiemann:
## Summary

- `remove_polymorphic_object_columns()` dropped the `_content_type` and `_object_id` columns with `ALTER TABLE` but did not flush deferred FK trigger events first.
- On branch revert, the preceding row deletion queues deferred FK trigger events from the `DEFERRABLE INITIALLY DEFERRED` through-table FK (created by any `MULTIOBJECT` field on the same COT). PostgreSQL then rejects the `ALTER TABLE` with `ObjectInUse: cannot ALTER TABLE "custom_objects_<id>" because it has pending trigger events`.
- Fix: issue `SET CONSTRAINTS ALL IMMEDIATE` before the first `ALTER TABLE`, mirroring what `_schema_remove_field()` already does for scalar fields.

## Test plan

- [ ] Run the repro script from issue #595 — revert should complete without error.
- [ ] Existing poly field tests pass.
- [ ] CI green.

### DIFF
--- a/netbox_custom_objects/field_types.py
+++ b/netbox_custom_objects/field_types.py
@@ -1000,6 +1000,10 @@ class ObjectFieldType(FieldType):
         ct_field_name = f"{field_instance.name}_content_type"
         oid_field_name = f"{field_instance.name}_object_id"
 
+        # Flush deferred FK trigger events before ALTER TABLE; PostgreSQL rejects
+        # column removal with "pending trigger events" when a row deletion (from
+        # the revert path) has queued events on a DEFERRABLE FK column.
+        schema_editor.execute('SET CONSTRAINTS ALL IMMEDIATE')
         try:
             oid_field = model._meta.get_field(oid_field_name)
             schema_editor.remove_field(model, oid_field)
--- a/netbox_custom_objects/field_types.py
+++ b/netbox_custom_objects/field_types.py
@@ -1000,10 +1000,6 @@ class ObjectFieldType(FieldType):
         ct_field_name = f"{field_instance.name}_content_type"
         oid_field_name = f"{field_instance.name}_object_id"
 
-        # Flush deferred FK trigger events before ALTER TABLE; PostgreSQL rejects
-        # column removal with "pending trigger events" when a row deletion (from
-        # the revert path) has queued events on a DEFERRABLE FK column.
-        schema_editor.execute('SET CONSTRAINTS ALL IMMEDIATE')
         try:
             oid_field = model._meta.get_field(oid_field_name)
             schema_editor.remove_field(model, oid_field)
--- a/netbox_custom_objects/field_types.py
+++ b/netbox_custom_objects/field_types.py
@@ -1000,6 +1000,12 @@ class ObjectFieldType(FieldType):
         ct_field_name = f"{field_instance.name}_content_type"
         oid_field_name = f"{field_instance.name}_object_id"
 
+        # Flush deferred FK trigger events before ALTER TABLE; PostgreSQL rejects
+        # column removal with "pending trigger events" when a row deletion (from
+        # the revert path) has queued events on a DEFERRABLE FK column.
+        # Also called by CustomObjectTypeField.delete() for the full removal block,
+        # but kept here so this method is self-contained when called directly.
+        schema_editor.execute('SET CONSTRAINTS ALL IMMEDIATE')
         try:
             oid_field = model._meta.get_field(oid_field_name)
             schema_editor.remove_field(model, oid_field)
--- a/netbox_custom_objects/models.py
+++ b/netbox_custom_objects/models.py
@@ -3636,6 +3636,11 @@ class CustomObjectTypeField(CloningMixin, ExportTemplatesMixin, ChangeLoggedMode
             _unwire_polymorphic_reverse_descriptors(self)
 
         with schema_conn.schema_editor() as schema_editor:
+            # Flush deferred FK trigger events before any ALTER TABLE or DROP TABLE.
+            # PostgreSQL rejects DDL with "pending trigger events" when a row
+            # deletion (e.g. from the branching revert path) has queued events on
+            # a DEFERRABLE FK column.  Guards all removal paths below.
+            schema_editor.execute('SET CONSTRAINTS ALL IMMEDIATE')
             if self.type == CustomObjectFieldTypeChoices.TYPE_COORDINATES:
                 # Drop both backing columns (latitude/longitude).
                 for column_name, model_field in field_type.get_model_field(self).items():
--- a/netbox_custom_objects/tests/test_polymorphic_fields.py
+++ b/netbox_custom_objects/tests/test_polymorphic_fields.py
@@ -1148,6 +1148,65 @@ class ReferencedObjectDeletionTest(
             through_model_name.lower(), django_apps.all_models.get(APP_LABEL, {})
         )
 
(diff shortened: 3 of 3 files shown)


bctiemann:
@claude /review and check only for correctness.

arthanson (changes requested):
Should have a test for this scenario, should be easy to adapt from script in issue.

arthanson (changes requested):
Any ALTER TABLE ... DROP COLUMN on a COT's table during branch revert fails with "pending trigger events" when the COT has a populated MULTIOBJECT field - PR #596 only guards the polymorphic-object path, leaving scalar-field removal (and the poly-multiobject drop) still broken.

suggested fix: Hoist SET CONSTRAINTS ALL IMMEDIATE to the top of the with connection.schema_editor() as schema_editor: block in CustomObjectTypeField.delete() so it guards the poly-object, poly-multiobject, and scalar paths uniformly - rather than only inside remove_polymorphic_object_columns().

bctiemann:
Good catch, Arthur — implemented in 8f3cac5, with one adjustment in bb52b68 after CI revealed a subtlety.

The initial implementation hoisted `SET CONSTRAINTS ALL IMMEDIATE` exclusively to the top of the `with schema_conn.schema_editor()` block in `CustomObjectTypeField.delete()`, removing it from `remove_polymorphic_object_columns()`. That covers your three identified paths (poly-object, poly-multiobject, scalar), but CI failed because the regression test calls `remove_polymorphic_object_columns()` directly — bypassing `delete()` entirely — and the guard was no longer in the method.

The fix keeps `SET CONSTRAINTS ALL IMMEDIATE` in **both** places:

- **`remove_polymorphic_object_columns()`** — self-contained guard for any direct caller (including the regression test, and any future caller that doesn't go through `delete()`).
- **`CustomObjectTypeField.delete()`** — covers the poly-multiobject, scalar, and coordinates paths that `remove_polymorphic_object_columns()` doesn't touch, which is exactly what your review asked for.

`SET CONSTRAINTS ALL IMMEDIATE` is idempotent, so the redundant double-call when `delete()` invokes `remove_polymorphic_object_columns()` is harmless.

arthanson (approved):

PR #597, opened 2026-06-24
Closes #376: Add first-class ownership to Custom Object instances

### LINKED ISSUE #376: Ownership of Custom Object Instances
### Plugin Version

0.4.5

### Proposed functionality

Netbox 4.5 introduced the concept of ownership on most models. This allows Netbox users document responsibility for a given object, whether that’s a device, a rack, an IP prefix, a VLAN, or many other resources.

Netbox Custom Object Types support ownership, but instances of Custom Object types do not.

This proposal extends the existing implementation so that Netbox users can own instances created from Custom Object types natively without linking to an object or multi-object field.

### Use case

Additional steps are required to assign owners to instances of Custom Objects, but it feels like this should be a first‑class capability, available as seamlessly as tags, descriptions, or changelog entries.

I should be able to click on an owner and see instances of Custom Object types the owner has been assigned in line with native Netbox objects.

### External dependencies

_No response_

bctiemann:
## Summary

- Adds `OwnerMixin` to `CustomObject` so every CO instance has an `owner = ForeignKey('users.Owner', null=True)` column
- Existing CO tables automatically gain the `owner_id` column via the existing `heal_cot` / `post_migrate` mechanism (field is nullable, so no manual migration required)
- Owner/owner-group fields appear in the CO edit form via `OwnerFormMixin`
- Owner and owner-group filters added to the CO list view filterset and filter form
- `owner` field exposed in the CO REST API as a nested `OwnerSerializer` (read: full object; write: PK)
- Owner link displayed in the CO detail page subtitle (matching the generic object template pattern)

## Test plan

- [ ] Create a new COT and a CO instance; set an owner via the edit form and confirm it appears in the subtitle and persists
- [ ] Confirm existing CO tables gain `owner_id` after running `manage.py upgrade_custom_objects` (or on next server start)
- [ ] Filter the CO list view by owner and owner group
- [ ] Via API: PATCH a CO with `{"owner": <id>}` and confirm the response includes the nested owner object
- [ ] Verify a CO with no owner renders cleanly (no broken subtitle)

Closes: #376


bctiemann:

netbox_custom_objects/tests/test_api.py
```diff
@@ -1601,3 +1606,158 @@ def test_get_response_includes_cross_cot_m2m_field(self):
...
+
+        response = self.client.get(
+            self._list_url(),
+            {'owner_id': owner_a.pk},
+            **self.header,
+        )
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        ids = [r['id'] for r in response.data['results']]
+        self.assertIn(obj_a.pk, ids)
+        self.assertNotIn(obj_b.pk, ids)
```
jnovinger, line 1742: I generally try to avoid in/not in assertions in tests as they tend to be weak signals of correctness. But, maybe that's not appropriate here since this is checking presence in a list, not a string.

jnovinger (approved):

jnovinger (approved):

PR #600, opened 2026-06-29
#532 - move integer BigIntegerField

### LINKED ISSUE #532: Field type for 64-bit integer
### Plugin Version

v0.5.1

### Proposed functionality

Hi,
I‘m looking for a field type which supports 64-bit integers. 

### Use case

Like to add integers, that are exceeding 32-bit length. 

### External dependencies

I don’t think so.

arthanson:
### Fixes: #532 

Switched IntegerField to BigIntegerField for int fields and supporting migration.  I didn't see a reason to keep the IntegerField, NetBox uses BitIntegerField for pretty much everything now.


arthanson:

arthanson:
Claude issues resolved.

bctiemann (changes requested):
LGTM, but I feel like this should be into `feature` for a couple of reasons -- 1) it's technically an expected-behavior change, and 2) it has a migration which will clash with the 0015 that just got merged into `feature` and will need to be renumbered at some point anyway.

PR #601, opened 2026-06-29
#551 Implement Location Field

### LINKED ISSUE #551: Special Custom Object Field Type - coordinates - Latitude and Longitude
### Plugin Version

0.5.1

### Proposed functionality

Implement a special object field type - coordinates, with the same semantic and validation as Latitude / Longitude fields on Netbox Site Objects and Netbox Device Objects.

Right now the custom objects plugins allows the utilization of 'Decimal' field type to represent coordinates - usually defining two Decimal fields - latitude and longitude, but there are drawbacks. First does not enforces that both lat and long fields are both null or not null. Also there are limits on the number of decimal digits. Tipically in sites / devices we use 6 decimal digits.

The most important feature of the lat/long fields in the devices and site forms is the 'MAP' button, that opens a google maps vision. If we create lat/long in a custom object using the decimal field type, we miss the 'map' button.

The idea is that a 'coordinate field type' was stored in database as two virtual columns, and rendered in forms as two fields, with the map button next to these fields. The lat/long field names may be user defined, but must be created in pairs.

### Use case

Use custom objects to document OSP objects - in particular splice boxes geocordinates and map visualization, that does not belongs to any particular site and are deployed in city streets ducts / trenchs.

### External dependencies

N/A

arthanson:
### Fixes: #551 

Adds a new coordinates field type to Custom Objects for storing a geographic latitude/longitude pair, mirroring NetBox core's native Site/Device coordinate handling - plain decimal columns, no PostGIS dependency.

## How it works

A single coordinates field named location expands into two real backing columns:

- location_latitude — decimal, range −90 to 90, 6 decimal places
- location_longitude — decimal, range −180 to 180, 6 decimal places

## Behavior

- Both-or-neither validation — latitude and longitude must both be set or both be empty; half-populated pairs are rejected in forms, bulk edit, and via the REST API.
- REST API — exposed as two flat fields (<name>_latitude / <name>_longitude), matching NetBox core's serializer convention, rather than a nested object.
- UI — rendered as two grouped inputs on create/edit forms; detail views show the pair with a Map button that opens the location using NetBox's MAPS_URL config (Google Maps by default). The build_map_url() helper uses core's build_coords_url on NetBox ≥ 4.6.2 and replicates it on earlier versions.
- Filtering — registers a numeric filter on each backing column.
- Cloning — clones both backing columns.
- Schema ops — add/rename/delete of a coordinates field manages both columns.
- Unsupported options — unique and default are rejected for coordinates fields (no single value to constrain).

Monosnap Cot1 | NetBox REST API 2026-06-29 14-18-10
Monosnap Cot1 1 | NetBox 2026-06-29 14-18-40
Monosnap Editing cot1 Cot1 1 | NetBox 2026-06-29 14-18-24


arthanson:

arthanson:
Claude review - fixed 1 and 2.

bctiemann (approved):

Issue #98, opened 2025-07-09
Ability to use config_context with custom_objects


erichester76:
### Plugin Version

0.10.0

### Proposed functionality

The ability to add config_context to custom objects for building configurations based on these new objects.

### Use case

Would be very useful with ansible (future functionality) so that these new objects can be used to build configurations based on object types.

### External dependencies

Although not dependent would be most useful with ansible.

PR #602, opened 2026-06-30
#98 - Add optional ConfigContext to Custom Objects

arthanson:
### Fixes: #98

Adds a config_context_enabled checkbox to Custom Object Types. When enabled, instances of that type gain NetBox's local_context_data JSON field — a per-object data blob usable for downstream config generation (e.g. Ansible, per the issue).

- Defaults to off.
- Can only be set at creation - disabled when editing an existing type, since the backing column is created once when the type's table is built (managed = False thereafter).

## Inherit config context

Config-context-enabled custom objects can now inherit the ConfigContexts that apply to NetBox objects they reference - the same site/tenant/role/etc. contexts a Device there would get — merged under their own local_context_data.

**How:** by field-naming convention, no configuration. If a type has a single object field named site, tenant, role, platform, location, device_type, or cluster pointing at the matching model, get_config_context() builds a proxy from those values and reuses NetBox's ConfigContextQuerySet.get_for_object() — so all of NetBox's matching, hierarchy (region/site-group/tenant-group/cluster derivation), weight ordering, and tag matching apply. local_context_data is merged last and always wins. A type with none of these fields stays local-only.

### Notes:
- Field names are unique per type, so there's no dimension ambiguity by construction.
- A field must match both name and target model; a mistyped/mispointed field is ignored.
- The Config Context tab gains the standard 3-panel view (Rendered / Local / Source Contexts); the source list is RBAC-filtered like NetBox's own view.


arthanson:

arthanson:
claude feedback:

 1 - Not valid.

2 - *_id concern not valid; added tenant + role dimension tests for coverage.

3 - Valid; docs note added then removed per your request. Enforcement stays at clean().

4 - already addressed in docs.

bctiemann (approved):

PR #622, opened 2026-07-14
#620 - avoid loading entire table into memory when opening bulk import/edit/delete pages

### LINKED ISSUE #620: On large tables, clicking the import button causes a memory spike (loading the whole table in memory?)
### Plugin Version

0.6.0

### NetBox Version

4.6.4

### Python Version

3.14.4

### Steps to Reproduce

1. navigate to Custom Objects
2. select one with millions of records
3. click import
4. observe the memory spike on the computer/container

### Expected Behavior

Just get the import page as in regular objects

### Observed Behavior

- The browser just spins and fails eventually.
- the memory on the server spikes and crashes

arthanson:
### Fixes: #620 

do `if self.queryset is not None:` instead of `if self.queryset` to avoid loading entire table into memory.

### DIFF
--- a/netbox_custom_objects/tests/test_views.py
+++ b/netbox_custom_objects/tests/test_views.py
@@ -3,12 +3,13 @@ Tests for all UI views.
 """
 from django.contrib.contenttypes.models import ContentType
 from django.db import connection
-from django.test import TestCase
+from django.test import RequestFactory, TestCase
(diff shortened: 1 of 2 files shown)


arthanson:

arthanson:
pre claude - added missing tests.

bctiemann (approved):

PR #624, opened 2026-07-15
Closes #549: Surface Custom Objects in Jinja config templates

### LINKED ISSUE #549: Surface Custom Objects in Jinja2 device configuration templates
### Plugin Version

0.5.1

### Proposed functionality

Custom Objects should be accessible inside NetBox device configuration templates, which are rendered with Jinja2.

Two access patterns should be supported:

**Attribute-style namespace** (context injection):
```jinja2
{% for iface in custom_objects.OSPFInterface.filter(device=device) %}
  interface {{ iface.name }}
    ip ospf area {{ iface.area }}
{% endfor %}
```

**Filter syntax**:
```jinja2
{% for iface in 'OSPFInterface' | custom_objects %}
  interface {{ iface.name }}
{% endfor %}
```

Both patterns resolve the Custom Object Type by **name** at render time, so templates remain valid even if internal IDs change.

The implementation will use the `jinja2_filters` plugin hook and `get_jinja2_context()` method introduced in NetBox core PR #22363 (targeting NetBox 4.7). On NetBox versions that do not yet include those hooks, the plugin degrades gracefully: an informational message is logged at startup and the `custom_objects` identifier is simply unavailable in templates — no crash, no silent wrong output.

### Use case

Device configuration templates today can reference NetBox's built-in models (devices, interfaces, IP addresses, etc.) but have no way to reach plugin-managed data. Custom Objects are commonly used to model routing protocol parameters, circuit details, or site-specific attributes that belong in rendered device configs — for example OSPF interface costs, BGP peer groups, or MPLS label ranges. Without this feature, operators must duplicate that data in custom fields or work around it with external scripts.

### External dependencies

Requires NetBox 4.7+ (with netbox-community/netbox#22363 merged) for full functionality. The plugin will continue to install and operate normally on NetBox 4.5/4.6; config template integration is simply inactive on those versions and documented as such.

bctiemann:
## Summary

Custom Objects can now be referenced directly from NetBox device config/export templates, via two equivalent access patterns, both resolving the Custom Object Type by name at render time:

```jinja2
{% for iface in custom_objects.ospf_interface.filter(device=device) %}
  interface {{ iface.name }}
    ip ospf area {{ iface.area }}
{% endfor %}
```

```jinja2
{% for iface in 'ospf_interface' | custom_objects %}
  interface {{ iface.name }}
{% endfor %}
```

## Requirements and graceful degradation

This is built on the `jinja_filters` plugin resource and `PluginConfig.get_jinja_context()` hook added to NetBox core by netbox-community/netbox#22363 (spelled `jinja2_*` at merge time, later renamed to drop the "2" by netbox-community/netbox#22436). **Both hooks require NetBox 4.7+.**

On NetBox < 4.7, these hooks simply don't exist, so core never calls into this plugin's `jinja_filters`/`get_jinja_context()` — no error, no crash, `custom_objects` is just unavailable in templates. A single INFO message is logged at plugin startup so operators have an actionable explanation (checked via `registry['plugins']['jinja_filters']` after `super().ready()`, which reflects whether core's hook actually registered anything).

No `max_version` bump: not currently a blocker (the `feature` branch's currently-reported NetBox version is still 4.6.x), and this project handles version-ceiling bumps as their own dedicated PRs rather than bundling them into feature work.

## Design notes

- Both access patterns share the same by-name resolution logic. The attribute-style form (`custom_objects.<name>`) returns the model's default manager (`.filter()`, `.all()`, etc.); the filter form (`'<name>' | custom_objects`) returns a queryset of all instances.
- An unknown type name is handled identically by both forms: a warning is logged (once per name, per process — not once per lookup, so a typo'd template rendered against many objects doesn't spam the log), and the reference resolves to an `EmptyCustomObjectsQuerySet` stand-in rather than raising. That stand-in covers the read-only QuerySet surface templates realistically chain onto it — `.filter()`, `.exclude()`, `.all()`, `.order_by()`, `.values()`, `.values_list()`, `.select_related()`, `.prefetch_related()`, `.distinct()`, `.annotate()`, slicing, `.get()` (still raises, matching real `QuerySet.get()` semantics) — so templates keep rendering no rows regardless of how the result is chained.
- `CustomObjectsNamespace` caches resolved names per-instance, so a template referencing the same type multiple times in one render issues one lookup, not one per reference. A fresh instance is created per render (via `get_jinja_context()`), so this doesn't go stale across renders.
- Continues paused work from branch `jinja2-config-template-support` (blocked on the core PR landing and FR feedback), updated for the subsequent hook rename.

## Tests

New `netbox_custom_objects/tests/test_jinja_integration.py` (25 tests):
- Direct unit tests for `EmptyCustomObjectsQuerySet`, the filter function, and `CustomObjectsNamespace` (always run, independent of NetBox version) — including chaining tolerance, per-render caching, and warn-once-per-process behavior
- `get_jinja_context()` tests (always run)
- End-to-end tests exercising the real `render_jinja2()` / `ConfigTemplate.render()` pipeline, which skip cleanly with a clear reason when NetBox doesn't expose the hooks, including unknown-name behavior for both access patterns

Verified locally against both:
- NetBox `main` (no hooks): 21/25 pass, 4 end-to-end tests skip as expected
- NetBox `feature` (hooks present): 25/25 pass, confirming the full integration actually works end-to-end (not just the degradation path)

`ruff check` passes.

## Docs

Added a "Jinja Config Templates" section to `docs/index.md`, alongside the existing Config Context section.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -190,6 +190,11 @@ Notes:
 
 - The attribute-style form returns the model's manager (`.filter(...)`, `.all()`, etc.); the filter form returns a queryset of all instances of that type.
 - An unknown type name is handled the same way by both forms: a warning is logged, and the reference resolves to an empty, chainable stand-in — further calls like `.filter(...)` or `.all()` continue to render no rows rather than raising. Check the type's internal name (shown on its detail page) if a template renders no data.
+- A type's internal name may begin with a digit (e.g. `123foo`), which isn't valid Jinja dot-notation. Use bracket notation with the attribute-style form instead:
+
+```jinja2
+{% for obj in custom_objects['123foo'].filter(device=device) %}
+```
 
 ### Deletions
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -189,7 +189,7 @@ interface {{ iface.name }}
 Notes:
 
 - The attribute-style form returns the model's manager (`.filter(...)`, `.all()`, etc.); the filter form returns a queryset of all instances of that type.
-- An unknown type name is handled differently by design: the attribute form raises `AttributeError` (which Jinja treats as `Undefined`), while the filter form logs a warning and returns an empty list. Both fail quietly rather than crashing the render — check the type's internal name (shown on its detail page) if a template renders no data.
+- An unknown type name is handled the same way by both forms: a warning is logged, and the reference resolves to an empty, chainable stand-in — further calls like `.filter(...)` or `.all()` continue to render no rows rather than raising. Check the type's internal name (shown on its detail page) if a template renders no data.
 
 ### Deletions
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -163,7 +163,7 @@ Notes:
 ### Jinja Config Templates
 
 !!! note
-    Requires NetBox 4.7 or later. On earlier NetBox versions, `custom_objects` is simply unavailable in config templates — a single informational message is logged when the plugin starts, and nothing else changes.
+    Requires NetBox 4.7 or later. On earlier NetBox versions, `custom_objects` is simply unavailable in config templates and nothing else changes — no error, no crash. A `DEBUG`-level log message noting this is emitted at plugin startup; enable debug logging if you need to confirm why `custom_objects` isn't resolving.
 
 Custom Objects can be referenced directly from NetBox [config templates](https://netboxlabs.com/docs/netbox/models/extras/configtemplate/), so device configuration can pull in data modelled with Custom Object Types (e.g. OSPF interface parameters, BGP peer groups, MPLS label ranges) alongside built-in NetBox models.
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -160,6 +160,37 @@ Notes:
 - A field must be an object field with both the **name and target model** above; a mis-named or mis-pointed field is simply ignored.
 - If a type defines **none** of these fields, aggregation is skipped entirely and its rendered context is just its Local Context Data. This is a deliberate difference from Devices/VMs: **global (unassigned) ConfigContexts are not applied** to such a type — enabling config context support alone never silently pulls in every global context. Add at least one convention field (e.g. `site`) to opt the type into source aggregation; global contexts then apply too (as they do for any object with a dimension).
 
+### Jinja Config Templates
+
+!!! note
+    Requires NetBox 4.7 or later. On earlier NetBox versions, `custom_objects` is simply unavailable in config templates — a single informational message is logged when the plugin starts, and nothing else changes.
+
+Custom Objects can be referenced directly from NetBox [config templates](https://netboxlabs.com/docs/netbox/models/extras/configtemplate/), so device configuration can pull in data modelled with Custom Object Types (e.g. OSPF interface parameters, BGP peer groups, MPLS label ranges) alongside built-in NetBox models.
+
+Two equivalent access patterns are available, both resolving the Custom Object Type by its **internal name** at render time (so templates keep working even if the type's slug or internal table ID changes):
+
+Attribute-style, via the `custom_objects` context variable:
+
+```jinja2
+{% for iface in custom_objects.ospf_interface.filter(device=device) %}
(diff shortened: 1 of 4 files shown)


bctiemann:

bctiemann:
Addressed all findings. Also changed the `INFO` level startup log message to `DEBUG` to avoid spamming users who are not on NetBox v4.7 yet.

netbox_custom_objects/jinja_env.py
```diff
@@ -0,0 +1,201 @@
...
+    of all its instances.
+
+    Example::
+
+        {% for iface in 'ospf_interface' | custom_objects %}
+    """
+    cot = _resolve_custom_object_type(type_name)
+    if cot is None:
+        return EmptyCustomObjectsQuerySet()
+    return cot.get_model().objects.all()
```
pheus: This filter can be evaluated by Jinja while the template is being compiled because the input is usually a string literal. That can cause database queries at compile time and, in some cases, run the lookup again during rendering.

Could we mark the filter as context-dependent to prevent constant folding?

```python
from jinja2 import pass_context


@pass_context
def custom_objects_filter(_context, type_name):
    ...
```

Please also add a regression test that compiles and renders the documented loop syntax and verifies the resolver is called only once.

docs/index.md
```diff
@@ -160,6 +160,37 @@ Notes:
...
 
+### Jinja Config Templates
+
+!!! note
+    Requires NetBox 4.7 or later. On earlier NetBox versions, `custom_objects` is simply unavailable in config templates and nothing else changes — no error, no crash. A `DEBUG`-level log message noting this is emitted at plugin startup; enable debug logging if you need to confirm why `custom_objects` isn't resolving.
+
+Custom Objects can be referenced directly from NetBox [config templates](https://netboxlabs.com/docs/netbox/models/extras/configtemplate/), so device configuration can pull in data modelled with Custom Object Types (e.g. OSPF interface parameters, BGP peer groups, MPLS label ranges) alongside built-in NetBox models.
+
+Two equivalent access patterns are available, both resolving the Custom Object Type by its **internal name** at render time (so templates keep working even if the type's slug or internal table ID changes):
```
pheus, line 170: `CustomObjectType.name` allows names beginning with a digit, but those names cannot be accessed using the documented dot syntax. For example, `custom_objects.123foo` is not valid Jinja syntax.

Bracket syntax works for these cases:

```jinja2
{% for obj in custom_objects['123foo'].filter(...) %}
```

Could we document this form and add a test covering a leading-digit type name?

netbox_custom_objects/jinja_env.py
```diff
@@ -0,0 +1,201 @@
...
+    ``<name> | custom_objects`` (``.filter()``, ``.exclude()``, ``.all()``,
+    ``.order_by()``, ``.values()``, ``.values_list()``, ``.select_related()``,
+    ``.prefetch_related()``, ``.distinct()``, ``.annotate()``, slicing,
+    iteration, ``len()``, ``.count()``, etc.), always yielding no results.
+    Unlike a real QuerySet, it accepts arbitrary kwargs without validating
+    them against a model, since there is no model to validate against.
+
+    This lets a template written against a Custom Object Type that was
+    renamed or deleted keep rendering (with no data) instead of raising, for
+    either access pattern, regardless of how the result is chained.
```
pheus: “Regardless of how the result is chained” feels a little too broad for a partial `QuerySet` implementation. I’d suggest narrowing this to say that the supported common queryset methods can be chained safely.

pheus (changes requested):
Thanks for the work on this. The overall approach looks good, and the added test coverage is helpful.

I’ve left two comments that I think we should address before merging:

- Prevent the `custom_objects` filter from performing database queries during Jinja template compilation, and add a regression test for it.
- Document and test bracket notation for custom object type names that begin with a digit, since those cannot be accessed using dot notation.

I’m marking this as changes requested for now. Once those are addressed, this should be in good shape.

pheus (approved):
Thanks for the follow-ups!

PR #630, opened 2026-07-20
Fix bulk import failing on required Hidden/Read-only fields

### LINKED ISSUE #626: Bulk Import Ignores Hidden Editable Fields and Returns Misleading Validation Errors
### Plugin Version

0.6.0

### NetBox Version

NetBox Community v4.6.4-Docker-5.0.1

### Python Version

3.14.4

### Steps to Reproduce

1. Create a Custom Object Type.
2. Create a field: `identifier`
3. Configure: `Required = Yes`, `Editable = Hidden`
4. Attempt to import: `{ "identifier": "12345" }`
5. Run Bulk Import.

### Expected Behavior

One of the following should happen:

**Option A (preferred)**
Bulk Import should be allowed to write to hidden fields.
Hidden fields affect manual editing forms but should not prevent imports.

**Option B**
If hidden fields are not writable through import, return an explicit error such as:
`Field 'identifier' is not importable because Editable = Hidden.`

### Observed Behavior

Import fails with:
`identifier: This field is required.`
even though the value is present in the payload.

The field is displayed in the Bulk Import field list, suggesting it can be imported.

bctiemann:
## Summary

- `CustomObjectBulkImportView.get_model_form()` included every Custom Object Type Field in the import form and disabled the ones with `ui_editable != YES`, mirroring the regular edit form's behavior — but Django's `disabled=True` form fields ignore submitted data entirely. A `Required` + `Hidden` field therefore always failed bulk import with `This field is required`, even though the imported row supplied a value (the `for_csv_import=True` kwarg passed at the call site isn't a real parameter of `get_annotated_form_field()`, so it silently did nothing).
- Fix: omit non-editable (Hidden/Read-only) fields from the import form entirely, matching core NetBox's own CSV import precedent (`NetBoxModelImportForm._get_custom_fields`). Since Custom Object fields are real model columns (unlike core's JSON-stored custom fields), this also requires listing them in `Meta.exclude` — `fields="__all__"` alone still auto-generates a field for any model column left out of the form's declared attrs.

## Test plan

- [x] Added a regression test (`test_bulk_import_omits_hidden_required_field_from_form`) asserting the hidden field is absent from the import form and that a row omitting it validates successfully.
- [x] Verified via git-stash-based regression testing: the new test fails against the pre-fix code with the exact reported error shape, and passes after the fix.
- [x] Ran the full `test_views.py` module; only 4 pre-existing, unrelated failures remain (confirmed identical against unmodified `main` — a `netbox_branching` import-guard gap).
- [x] `ruff check` passes.

Closes #626

### DIFF
--- a/netbox_custom_objects/tests/test_views.py
+++ b/netbox_custom_objects/tests/test_views.py
@@ -453,6 +453,49 @@ class CustomObjectViewTestCase(
         """Regression #620: CustomObjectBulkDeleteView.get_queryset()."""
         self._assert_get_queryset_does_not_full_scan(views.CustomObjectBulkDeleteView)
 
+    def test_bulk_import_omits_hidden_required_field_from_form(self):
+        """Regression #626: a Required + Hidden field must not break bulk import.
+
+        Before the fix, CustomObjectBulkImportView.get_model_form() included every
+        field in the import form and disabled the ones with ui_editable != YES
+        (mirroring the regular edit form). A disabled Django form field ignores
+        submitted data, so a required-but-hidden field always failed validation
+        with "This field is required" even though the imported row supplied a
+        value. The fix omits such fields from the import form entirely, matching
+        core NetBox's own CSV import precedent (NetBoxModelImportForm._get_custom_fields).
+
+        Uses its own dedicated Custom Object Type rather than the class-level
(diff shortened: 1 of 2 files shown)


bctiemann:
@claude /review and keep to areas of potential concern.

arthanson (approved):

PR #631, opened 2026-07-20
Fixes: #625 - schema_id (and related bookkeeping fields) nulled by unrelated field edits

### LINKED ISSUE #625: Editing a field causes schema_id to be set to null
### Plugin Version

0.6.0

### NetBox Version

4.6.5

### Python Version

3.12

### Steps to Reproduce

1. Create a custom object type with `Internal name = "test", URL path/slug = "test"`
2. Add a field with `Name = "test"`
3. Edit the field and set `Label = "Test"`

### Expected Behavior

The custom object field retains its `schema_id` value.

### Observed Behavior

The custom object field has a null `schema_id` value.

Before edit:
```pgsql
netbox=# SELECT id, name, custom_object_type_id, schema_id FROM netbox_custom_objects_customobjecttypefield;
 id | name | custom_object_type_id | schema_id 
----+------+-----------------------+-----------
  1 | test |                     1 |         1
(1 row)
```

After edit:
```pgsql
netbox=# SELECT id, name, custom_object_type_id, schema_id FROM netbox_custom_objects_customobjecttypefield;
 id | name | custom_object_type_id | schema_id 
----+------+-----------------------+-----------
  1 | test |                     1 |  
(1 row)
```

bctiemann:
## Summary

- `CustomObjectTypeFieldForm.Meta.fields = '__all__'` pulled `schema_id` in as a real, writable form field, but it's never rendered in any of the form's fieldsets. The browser therefore never submits a value for it; Django reads the missing key as `None`, and `ModelForm._post_clean()` overwrites the existing `schema_id` with that `None` on every save — even when the user only changed an unrelated attribute like `label`.
- Fix: set `editable=False` on `schema_id`, matching its existing `read_only_fields` treatment in the REST API serializer (`CustomObjectTypeFieldSerializer`) and NetBox core's own convention for internal, system-managed model fields. Django's `ModelForm` machinery automatically drops non-editable fields from `Meta.fields = '__all__'`, so it's simply absent from the form and never touched.
- While investigating, found the identical bug pattern also affects `deprecated`, `deprecated_since`, and `scheduled_removal` on the same form — same root cause (never rendered in a fieldset, silently reset by an unrelated edit). These are excluded via `Meta.exclude` on the form instead of `editable=False` on the model, since they **are** meant to be writable through the REST API (set via the portable-schema import executor) and aren't declared as explicit serializer fields there — `editable=False` would have made DRF silently force them read-only in the API too, a real regression.

## Test plan

- [x] Added a regression test (`test_editing_a_field_preserves_schema_id_and_deprecation_bookkeeping`) that creates a field with pre-set `schema_id`/`deprecated`/`deprecated_since`/`scheduled_removal`, edits only its `label` via the form, and asserts all four survive unchanged.
- [x] Verified via git-stash-based regression testing: the new test fails against the pre-fix code with the exact reported symptom (`schema_id` becomes `None`), and passes after the fix.
- [x] Confirmed the REST API's write access to `deprecated`/`deprecated_since`/`scheduled_removal` is unaffected (they're not `editable=False` at the model level, only excluded from this one form).
- [x] Generated migration (`0018_alter_customobjecttypefield_schema_id`) for the `editable=False` change — a no-op at the DB level, but Django tracks it in migration state.
- [x] Ran the broader `test_forms`/`test_models`/`test_api` suites (223 tests); only 2 pre-existing, unrelated failures remain (`netbox_branching` import-guard gap, confirmed identical against unmodified `main` in earlier work this session).
- [x] `ruff check` passes.

Closes #625

### DIFF
--- a/netbox_custom_objects/forms.py
+++ b/netbox_custom_objects/forms.py
@@ -248,6 +248,12 @@ class CustomObjectTypeFieldForm(CustomFieldForm):
     class Meta:
         model = CustomObjectTypeField
         fields = '__all__'
+        # deprecated/deprecated_since/scheduled_removal are set via the portable-schema
+        # import mechanism (or directly via the REST API), never through this form --
+        # excluded here so an unrelated edit doesn't silently reset them to their
+        # defaults. schema_id needs no entry: it's excluded automatically via the
+        # model field's editable=False.
+        exclude = ('deprecated', 'deprecated_since', 'scheduled_removal')
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
--- a/netbox_custom_objects/migrations/0018_alter_customobjecttypefield_schema_id.py
+++ b/netbox_custom_objects/migrations/0018_alter_customobjecttypefield_schema_id.py
@@ -1,0 +1,16 @@
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('netbox_custom_objects', '0017_customobjecttype_display_expression'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='customobjecttypefield',
+            name='schema_id',
+            field=models.PositiveIntegerField(blank=True, editable=False, null=True),
+        ),
+    ]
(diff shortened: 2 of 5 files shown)


bctiemann:
@claude /review and limit response to areas of concern.

jnovinger (approved):

PR #632, opened 2026-07-21
Fixes #628: `?id=` filter silently ignored on Custom Object REST API

### LINKED ISSUE #628: Query custom object by id doesn't work
### Plugin Version

0.6.0

### NetBox Version

4.5.3

### Python Version

3.12.3

### Steps to Reproduce

1. Create custom object type (no fields needed)
2. Create 2 instances
3. Try GET /api/plugins/custom-objects/my-object/?id=1

### Expected Behavior

Response page with single object with "id": 1

### Observed Behavior

Response page with all objects

bctiemann:
## Summary

- `get_filterset_class()` builds the dynamic FilterSet for a Custom Object Type with `Meta.fields = []` (to disable django-filter's auto-generation) and adds every filter explicitly by iterating `CustomObjectTypeField` records. That loop never includes the base `id` primary key column, since it isn't a user-defined field — so no filter was ever registered for `id`, and `GET .../?id=<pk>` silently returned every row instead of just the matching one.
- Fix: add `"id"` to `Meta.fields`, matching core NetBox's own FilterSets (which explicitly list `'id'` first in `Meta.fields`). django-filter auto-generates a `MultiValueNumberFilter` for it via `BaseFilterSet.FILTER_DEFAULTS`' `AutoField` mapping — the same mechanism core relies on. No other field handled by `build_filter_for_field` is affected, since that loop only ever touches `CustomObjectTypeField`-backed columns, never `id`.

## Test plan

- [x] Added `test_filter_by_id_returns_only_matching_object` directly to `ScalarFieldFiltersetTestCase` (the shared base class in `test_filtersets.py`), so it runs automatically across all 11 of its field-type subclasses (text, longtext, integer, decimal, boolean, date, datetime, URL, JSON, select, multiselect) rather than needing a dedicated fixture.
- [x] Added `test_filter_by_id` to `OwnerAPITest` in `test_api.py`, which already creates a Custom Object Type with zero fields in `setUp()` — the exact reproduction scenario from the reported bug — and already has the `_list_url()`/`_add_perm()` helpers needed.
- [x] Verified via git-stash-based regression testing: both tests fail against the pre-fix code with the exact reported symptom (all rows returned instead of one), and pass after the fix.
- [x] Ran the full `test_filtersets`/`test_api` suites (214 tests); all pass, no regressions.
- [x] Confirmed no migration is needed (`makemigrations --check` reports no changes) — this is a pure Python FilterSet change, no model modification.
- [x] `ruff check` passes.

Closes #628

### DIFF
--- a/netbox_custom_objects/filtersets.py
+++ b/netbox_custom_objects/filtersets.py
@@ -405,14 +405,15 @@ def get_filterset_class(model):
     """
     Create and return a filterset class for the given custom object model.
     """
-    # fields=[] disables auto-generation; all filters are added explicitly below
-    # via build_filter_for_field so there are no shadowed duplicates.
+    # id is the only base column filterable via Meta.fields auto-generation (matching
+    # core FilterSets); every other filter is added explicitly below via
+    # build_filter_for_field, so there are no shadowed duplicates.
     meta = type(
         "Meta",
         (),
         {
             "model": model,
-            "fields": [],
+            "fields": ["id"],
         },
     )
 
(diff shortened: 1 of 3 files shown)


bctiemann:
@claude /review

pheus (approved):
Thanks. This looks good to me!

PR #634, opened 2026-07-21
Closes: #621 - Add "Set null" to bulk edit for all real, nullable fields

### LINKED ISSUE #621: Bulk Edit - Set Null
### Plugin Version

0.5.2

### Proposed functionality

When Bulk Editing Custom Objects, add the "Set Null" option for all relevant fields. Currently this option only manifests under certain conditions. 

### Use case

This would allow users to blank out a field for many Custom objects at once, similarly to how you can do this for native types like devices. 

### External dependencies

_No response_

bctiemann:
## Summary

- `CustomObjectBulkEditView.get_form()` never set `nullable_fields` on the dynamically-generated bulk-edit form, so it fell back to the empty default inherited from `NetBoxModelBulkEditForm`. The "Set null" checkbox only ever appeared when a Custom Object Type also happened to have native NetBox `extras.CustomField`s attached to its content type (an unrelated, incidental mechanism) — explaining the "only manifests under certain conditions" symptom in the report. Every genuine `CustomObjectTypeField` (text, integer, boolean, date, select, non-polymorphic object/multiobject, coordinates, etc.) never got Set Null support at all.
- Fix: build `nullable_fields` explicitly from the fields actually added to the form — all scalar types and non-polymorphic object/multiobject (each maps to a single real, nullable model column or M2M relation, the same shape as core's own `nullable_fields` entries, e.g. `Site.asns`) plus coordinates' two real latitude/longitude sub-columns.
- Polymorphic object/multiobject fields are deliberately excluded: their form sub-fields (e.g. `"<name>__ct"`) don't correspond 1:1 to a real model field name, so core's generic nullify lookup (which resolves the field via `_meta.get_field()`) can't safely resolve them. Also removed the polymorphic-multiobject template's hardcoded `bulk_nullable=True`, which rendered an always-on "Set null" checkbox that silently did nothing when checked — now consistent with polymorphic single-object, which never rendered one.
- Fixed a template comment bug caught in review: Django's `{# ... #}` inline comment tag does not match across newlines, so a multi-line comment added in an earlier commit was rendering as literal visible text on the bulk-edit page. Collapsed to one line (verified empirically against Django's template engine).
- Incidental fix: renamed the bulk-edit/bulk-delete buttons from "Bulk Edit"/"Bulk Delete" to "Edit Selected"/"Delete Selected", matching core NetBox's own button labels (e.g. Sites' bulk action bar).
- **Addressed automated review feedback:**
  - **Required fields are now excluded from `nullable_fields`.** Every custom object column is `null=True` at the DB level regardless of the field's own `required` flag, so nulling a required field previously succeeded silently — the type's own required constraint was never enforced anywhere. Guarded both the scalar/object/multiobject path and the coordinates path with `if not field.required`.
  - **Coordinates fields now get one shared "Set null" checkbox instead of two independent ones** for latitude/longitude. Previously a user could null just one half of what's logically a single field; the existing `bulk_clean()` validation caught the resulting inconsistent state (safe, but confusing). The two sub-fields are now excluded from `nullable_fields` entirely; a dedicated checkbox posts the field's logical name through the same `_nullify` mechanism core already parses, and `post_save_operations()` clears both columns together when present.

This is a non-breaking, purely additive UI/form change (no migration, no API contract change), so it targets `main` directly.

Screenshot 2026-07-21 at 6 48 21 PM

## Test plan

- [x] Added `test_bulk_edit_form_nullable_fields_includes_scalar_fields`, asserting the generated form's `nullable_fields` includes real scalar fields (and, per review, that the required `name` field is excluded).
- [x] Added `test_bulk_edit_set_null_clears_field`, an end-to-end test posting to the bulk-edit view with `_nullify` and confirming the field is actually cleared across selected objects.
- [x] Added `test_bulk_edit_set_null_clears_object_and_multiobject_fields`, validating the PR's core claim that non-polymorphic object/multiobject fields correctly support Set Null (matching core's `Site.asns` precedent).
- [x] Added `test_bulk_edit_coordinates_not_individually_nullable` and `test_bulk_edit_set_null_clears_coordinates_atomically`, covering the new atomic coordinates behavior.
- [x] Verified via git-stash-based regression testing at every step: each new/extended test assertion fails against the pre-fix code with the exact expected symptom, and passes after the corresponding fix.
- [x] Ran the broader `test_views`/`test_polymorphic_fields`/`test_field_types` suites (357 tests); only 5 pre-existing, unrelated failures remain (`netbox_branching` import-guard gap, confirmed identical against unmodified `main` in earlier work this session).
- [x] Explicitly reran the existing polymorphic bulk-edit tests to confirm the template changes don't affect their rendering/application paths.
- [x] Verified the template comment fix empirically against Django's template engine directly (reproduced the multi-line leak, confirmed the single-line version renders cleanly).
- [x] Verified live against a running local instance with FK/M2M Custom Object fields (Airport, Bar, Device, etc.) — confirmed no "Set null" checkbox exists pre-fix, and that an apparent "--" dropdown option some testers may see is a real, coincidentally blank-named record in test data, not a null-clearing mechanism.
- [x] `ruff check` passes.

Closes #621

### DIFF
--- a/netbox_custom_objects/templates/netbox_custom_objects/inc/bulk_edit_fields.html
+++ b/netbox_custom_objects/templates/netbox_custom_objects/inc/bulk_edit_fields.html
@@ -23,7 +23,9 @@
         <h3 class="col-9 offset-3 mb-3 h4">{{ group_info.1 }}</h3>
       </div>
       {% for sub_name in group_info.0 %}
-        {% render_field form|getfield:sub_name bulk_nullable=True %}
+        {# Not nullable: these sub-fields aren't in form.nullable_fields (see get_form()),
+           so a "Set null" control here would silently do nothing when checked. #}
+        {% render_field form|getfield:sub_name %}
       {% endfor %}
     {% endwith %}
   {% elif field.name in form.custom_object_type_rendered_names %}
--- a/netbox_custom_objects/templates/netbox_custom_objects/inc/bulk_edit_fields.html
+++ b/netbox_custom_objects/templates/netbox_custom_objects/inc/bulk_edit_fields.html
@@ -23,8 +23,7 @@
         <h3 class="col-9 offset-3 mb-3 h4">{{ group_info.1 }}</h3>
       </div>
       {% for sub_name in group_info.0 %}
-        {# Not nullable: these sub-fields aren't in form.nullable_fields (see get_form()),
-           so a "Set null" control here would silently do nothing when checked. #}
+        {# Not in form.nullable_fields (see get_form()) -- no bulk_nullable control. #}
         {% render_field form|getfield:sub_name %}
       {% endfor %}
     {% endwith %}
--- a/netbox_custom_objects/templates/netbox_custom_objects/inc/bulk_edit_fields.html
+++ b/netbox_custom_objects/templates/netbox_custom_objects/inc/bulk_edit_fields.html
@@ -27,8 +27,26 @@
         {% render_field form|getfield:sub_name %}
       {% endfor %}
     {% endwith %}
+  {% elif field.name in form.custom_object_type_coordinates_groups %}
+    {# Coordinates group: heading + lat/long sub-fields + one shared Set Null control #}
+    {% with group_info=form.custom_object_type_coordinates_groups|dict_get:field.name %}
+      <div class="row">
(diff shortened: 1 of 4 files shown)


bctiemann:
@claude /review and limit response only to areas of potential concern. Ensure correctness of UI flow and consistency with core NetBox usage.

arthanson (approved):

PR #641, opened 2026-08-02
Closes #496: Add optional link title to URL fields

### LINKED ISSUE #496: Add a name to the URL field
### Plugin Version

0.4.10

### Proposed functionality

When you create a URL field for an object, the URL is displayed on the instance's page.

It would be good to add a name to that field to specify a value that replaces the displayed URL.

### Use case

This way, the page wouldn't look so cluttered with random URLs, and we'd have a more organized page.

### External dependencies

_No response_

bctiemann:
## Summary

- A url-type CustomObjectTypeField now expands into two real DB columns: the URL itself, and an optional <name>_title used as the visible link text on an object's detail page instead of the raw URL (falling back to the URL itself when no title is set). List/table views are unaffected -- still plain-text URL.
- Mirrors CoordinatesFieldType's existing two-column pattern, but unlike coordinates the primary URL column keeps behaving like any other single-value column: unique, default, and regex validation all still apply to it exactly as before.
- This required making three previously coordinates-only DDL/validation code paths in models.py dict-aware (the generic single-column schema helpers used by every field type, the unique-conversion probe in clean(), and the backing-column-collision guard), since URL is the first multi-column type that also needs to flow through the generic single-column path via its unique/default support.
- No new migration or upgrade script is needed for existing installations: the plugin's existing post_migrate schema-heal pass already covers any nullable non-mixin column whose attribute name doesn't match a user field's own name, which the new title column satisfies the same way coordinates' latitude/longitude columns already do. Documented this explicitly in mixin_migration.py.

Screenshot 2026-08-06 at 6 12 51 AM

Screenshot 2026-08-06 at 6 13 01 AM

## Upgrading

Existing `url` fields gain the new `<name>_title` column automatically — no manual migration is needed. On the main schema this happens the next time `manage.py migrate` runs (or immediately via `manage.py upgrade_custom_objects`, which also supports `--dry-run`). On a NetBox Branching branch it happens the next time that branch itself is migrated (its "Migrate branch" action, available whenever the branch's migration state lags behind main).

This will need to be captured in a release note for the next minor release in which this feature ships.

## Test plan

- [x] New/extended tests across test_field_types.py (URLFieldTypeTestCase: model generation, title-optional, unique still enforced, type-conversion rejection both directions, backing-column collision both directions), test_schema_operations.py (rename and delete both drop/rename the title column), test_api.py (serializer exposes both columns flat, create round-trip with and without a title), test_views.py (add form renders both inputs, create with and without a title).
- [x] ruff check clean across the whole package.
- [x] Ran the full plugin test suite (1109 tests) against a NetBox 4.6.6 checkout. All new/changed tests pass. The only pre-existing failures (20, unrelated to this change) trace to an environment gap in that test setup -- netbox_branching is importable but not enabled in PLUGINS, which breaks any test touching CustomObjectTypeField/CustomObjectType rename or delete, confirmed by the identical failure occurring on the unmodified, pre-existing coordinates rename test -- plus one unrelated csv_update_data scaffolding gap in generic view tests.

### DIFF
--- a/docs/field-attributes.md
+++ b/docs/field-attributes.md
@@ -116,13 +116,14 @@ A `url` field stores the URL value plus an optional human-readable **link title*
 Adding one `url` field named `website` creates two backing columns:
 
 - `website` — the URL itself
-- `website_title` — optional display text shown in place of the raw URL on the
-  object's detail page
+- `website_title` — optional display text shown in place of the raw URL
 
 Behaviour:
 
-- **Detail page only.** The title is used as the visible link text on the object
-  detail page. List/table views continue to show the raw URL as plain text.
+- **Detail page and list view.** The title, when set, is used as the visible link
+  text both on the object's detail page and in its column in list/table views —
+  the two backing columns are never shown as separate table columns. Falls back to
+  the raw URL as link text when no title is set.
 - **Optional independently.** The title has no relationship to the URL value —
   setting one without the other is allowed and harmless.
 - **REST API.** The pair is exposed as two flat fields, `<name>` and `<name>_title`.
@@ -131,3 +132,8 @@ Behaviour:
 - **CSV import.** Bulk CSV import only populates the URL value; the title column is
   not importable via CSV (the same limitation applies to `coordinates` fields' backing
   columns).
+- **Upgrading from an older release.** Existing `url` fields gain the `<name>_title`
+  column automatically on the next `manage.py migrate` (or immediately via
+  `manage.py upgrade_custom_objects`). On a NetBox Branching branch, it's healed the
+  next time that branch itself is migrated. See the [release notes](releases.md) for
+  details.
--- a/docs/field-attributes.md
+++ b/docs/field-attributes.md
@@ -134,5 +134,7 @@ Behaviour:
   columns).
 - **Upgrading from an older release.** Existing `url` fields gain the `<name>_title`
   column automatically on the next `manage.py migrate` (or immediately via
-  `manage.py upgrade_custom_objects`). On a NetBox Branching branch, it's healed the
-  next time that branch itself is migrated.
+  `manage.py upgrade_custom_objects`) -- this also heals every existing NetBox
+  Branching branch's own schema, not just main's. A branch migrating on its own
+  afterward re-heals itself too, as a secondary safeguard, but isn't required for
+  the column to already be there.
--- a/docs/field-attributes.md
+++ b/docs/field-attributes.md
@@ -135,5 +135,4 @@ Behaviour:
 - **Upgrading from an older release.** Existing `url` fields gain the `<name>_title`
   column automatically on the next `manage.py migrate` (or immediately via
   `manage.py upgrade_custom_objects`). On a NetBox Branching branch, it's healed the
-  next time that branch itself is migrated. See the [release notes](releases.md) for
-  details.
+  next time that branch itself is migrated.
--- a/docs/field-attributes.md
+++ b/docs/field-attributes.md
@@ -128,3 +128,6 @@ Behaviour:
 - **REST API.** The pair is exposed as two flat fields, `<name>` and `<name>_title`.
 - `Must be unique` and `Default` continue to apply to the URL value itself, exactly
   as for any other `url` field.
+- **CSV import.** Bulk CSV import only populates the URL value; the title column is
+  not importable via CSV (the same limitation applies to `coordinates` fields' backing
+  columns).
--- a/docs/field-attributes.md
+++ b/docs/field-attributes.md
@@ -13,7 +13,7 @@ The following attributes are available when creating or editing a Custom Object 
 | `boolean` | True/false |
 | `date` | Date |
 | `datetime` | Date and time |
-| `url` | URL |
+| `url` | URL, with an optional link title |
 | `json` | Arbitrary JSON value |
 | `select` | Single selection from a choice set |
 | `multiselect` | Multiple selections from a choice set |
@@ -107,3 +107,24 @@ Behaviour:
 - **Map link.** Detail views render the coordinates with a **Map** button that opens the location
   using NetBox's `MAPS_URL` configuration parameter (Google Maps by default).
 - `Must be unique` and `Default` are not supported for `coordinates` fields.
+
+## URL Fields
+
+Field type: `url`
+
+A `url` field stores the URL value plus an optional human-readable **link title**.
+Adding one `url` field named `website` creates two backing columns:
+
+- `website` — the URL itself
+- `website_title` — optional display text shown in place of the raw URL on the
+  object's detail page
+
+Behaviour:
+
+- **Detail page only.** The title is used as the visible link text on the object
+  detail page. List/table views continue to show the raw URL as plain text.
+- **Optional independently.** The title has no relationship to the URL value —
+  setting one without the other is allowed and harmless.
+- **REST API.** The pair is exposed as two flat fields, `<name>` and `<name>_title`.
+- `Must be unique` and `Default` continue to apply to the URL value itself, exactly
+  as for any other `url` field.
--- a/netbox_custom_objects/__init__.py
+++ b/netbox_custom_objects/__init__.py
@@ -66,13 +66,16 @@ def _reset_deferred_co_field_data(sender, **kwargs):
 def _heal_branch_on_migrate(sender, branch, **kwargs):
     """netbox-branching's own ``post_migrate`` receiver.
 
-    ``Branch.migrate()`` applies migrations directly via Django's
-    ``MigrationExecutor`` against the branch's own connection and never calls
-    ``call_command('migrate', ...)``, so Django's core ``post_migrate`` signal
-    (handled by ``_heal_mixin_columns`` below) never fires for it. Without this
-    receiver, a branch provisioned before a plugin release that adds a new base
-    column (e.g. a multi-column field type's extra sub-column) never gets that
-    column healed into its own schema — only into the default connection's.
+    Secondary/defense-in-depth path only. The reliable trigger for healing
+    *existing* branches is ``heal_all_branches()``, called unconditionally
+    from ``_heal_mixin_columns`` below (Django's core ``post_migrate``, fired
+    by every ``manage.py migrate`` run) -- a schema change like a new
+    multi-column sub-column ships with no actual Django migration file, so
+    there's nothing for netbox-branching's ``MigrationExecutor`` to detect as
+    "pending" against a branch, and ``Branch.migrate()`` never runs for it.
+    This receiver still re-heals the one branch that was just migrated,
+    covering a *future* plugin release that does ship a real migration
+    alongside a schema change.
     """
     try:
         from netbox_custom_objects.mixin_migration import heal_branch  # noqa: PLC0415
@@ -161,6 +164,12 @@ def _heal_mixin_columns(sender, **kwargs):
     process so the cost is negligible on normal server starts where no
     migrations run.
 
+    Also heals every live Branch's own schema (heal_all_branches()), not just
+    the default connection's.  This can't be delegated to Branch.migrate()'s
+    own post_migrate signal (see heal_branch()'s docstring in
+    mixin_migration.py for why) -- the main upgrade path is the only reliable
+    trigger, so it runs unconditionally here alongside the main-schema heal.
+
     Skipped during makemigrations and collectstatic (DB may be unavailable or
     in an inconsistent state for our purposes).
     """
@@ -172,8 +181,11 @@ def _heal_mixin_columns(sender, **kwargs):
         return
 
     try:
-        from netbox_custom_objects.mixin_migration import heal_all_cots  # noqa: PLC0415
+        from netbox_custom_objects.mixin_migration import heal_all_branches, heal_all_cots  # noqa: PLC0415
         heal_all_cots(verbosity=kwargs.get("verbosity", 1))
+        # No-ops (zero-result) when netbox-branching isn't installed -- see
+        # heal_all_branches()'s own apps.is_installed() check.
+        heal_all_branches(verbosity=kwargs.get("verbosity", 1))
     except Exception:
         import logging  # noqa: PLC0415
         logging.getLogger(__name__).exception(
--- a/netbox_custom_objects/__init__.py
+++ b/netbox_custom_objects/__init__.py
@@ -63,6 +63,27 @@ def _reset_deferred_co_field_data(sender, **kwargs):
     _deferred_co_field_data.set(None)
 
 
+def _heal_branch_on_migrate(sender, branch, **kwargs):
+    """netbox-branching's own ``post_migrate`` receiver.
+
+    ``Branch.migrate()`` applies migrations directly via Django's
+    ``MigrationExecutor`` against the branch's own connection and never calls
+    ``call_command('migrate', ...)``, so Django's core ``post_migrate`` signal
+    (handled by ``_heal_mixin_columns`` below) never fires for it. Without this
+    receiver, a branch provisioned before a plugin release that adds a new base
+    column (e.g. a multi-column field type's extra sub-column) never gets that
+    column healed into its own schema — only into the default connection's.
+    """
+    try:
+        from netbox_custom_objects.mixin_migration import heal_branch  # noqa: PLC0415
+        heal_branch(branch, verbosity=kwargs.get("verbosity", 1))
+    except Exception:
+        logging.getLogger(__name__).exception(
+            "netbox_custom_objects: unexpected error healing branch %s after migrate",
+            getattr(branch, "pk", branch),
+        )
+
+
 def _register_branching_hooks_once():
     """Register netbox-branching integration hooks at most once per process.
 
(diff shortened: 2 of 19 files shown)


bctiemann:

netbox_custom_objects/mixin_migration.py
```diff
@@ -47,6 +47,14 @@ def _expected_base_fields(cot, model=None):
...
     fields because they are never created with db_column overrides.
 
+    Multi-column custom field types (e.g. CoordinatesFieldType's "_latitude"/
+    "_longitude", URLFieldType's "_title") are NOT excluded here, since their
+    backing columns' attribute names never equal the user field's own f.name --
+    so this heal pass also covers them as a deliberate side effect. This is why
+    a newly introduced sub-column for one of those types needs no dedicated
+    migration/upgrade code: as long as it's nullable, the next post_migrate run
+    (or `upgrade_custom_objects`) auto-adds it to every existing COT table.
+
```
pheus, line 64: Could we add a regression test for upgrading a branch that already existed before this plugin version?

This change has no migration and relies on the Django `post_migrate` heal pass, which uses the default connection. Branch migrations run on the branch connection and emit the Branching plugin's own signal, so I don't see how the new `_title` column reaches an existing branch schema.

netbox_custom_objects/models.py
```diff
@@ -3516,6 +3541,25 @@ def save(self, *args, **kwargs):
                     else:
                         _schema_add_field(self, model, schema_editor, schema_conn)
                         _apply_deferred_co_field(self)
+                        if self.type == CustomObjectFieldTypeChoices.TYPE_URL:
+                            # A url field's title column has no deferred-CO-field
+                            # replay support (matches the existing coordinates gap).
```
pheus: Could we include the title in deferred replay as well? During a squash merge, `_apply_deferred_co_field()` restores the URL before the `_title` column is added, so an existing title appears to be replaced by the default empty string.

Please add squash merge and revert coverage for both values.

netbox_custom_objects/models.py
```diff
@@ -3536,6 +3580,19 @@ def save(self, *args, **kwargs):
...
+                        ):
+                            # Only a rename touches the title column; other attribute
+                            # changes are persisted by super().save() below.
+                            old_title = field_type.title_field_name(self.original)
+                            new_title = field_type.title_field_name(self)
+                            old_title_field = field_type.get_model_field(self.original)[old_title]
+                            new_title_field = field_type.get_model_field(self)[new_title]
+                            old_title_field.contribute_to_class(model, old_title)
+                            new_title_field.contribute_to_class(model, new_title)
+                            schema_editor.alter_field(model, old_title_field, new_title_field)
```
pheus: Could the title column use the same conflict-aware rename path as the URL column? A direct `alter_field()` still assumes the original `_title` column exists, which can fail after independent main and branch renames.

We also need to rewrite `<old>_title` in the audit data; the existing helper is called only for the base field name. Please cover both columns in the branch rename and replay/revert tests.

netbox_custom_objects/field_types.py
```diff
@@ -524,18 +524,70 @@ def get_filterform_field(self, field, **kwargs):
...
+                help_text=_("Human-readable text shown instead of the raw URL."),
+            ),
+        }
 
     def get_form_field(self, field, **kwargs):
         return LaxURLField(
             assume_scheme="https", required=field.required, initial=field.default
         )
 
+    def get_form_fields(self, field):
```
pheus, line 590: Could we preserve the configured field description here? This path bypasses `get_annotated_form_field()`, so the URL field's help text disappears from edit and bulk-edit forms. The title input should also expose its own help text.

docs/field-attributes.md
```diff
@@ -107,3 +107,27 @@ Behaviour:
...
+- `website` — the URL itself
+- `website_title` — optional display text shown in place of the raw URL on the
+  object's detail page
+
+Behaviour:
+
+- **Detail page only.** The title is used as the visible link text on the object
+  detail page. List/table views continue to show the raw URL as plain text.
+- **Optional independently.** The title has no relationship to the URL value —
+  setting one without the other is allowed and harmless.
```
pheus, line 128: Since the URL and title may be set independently, the deletion impact should consider `<name>_title` too. The current count and dependent-object list only check `<name>`, so title-only values are omitted even though deleting the field removes them.

pheus (changes requested):
Thanks for working on this. The normal URL and title flow looks good.

I found three cases that still need another pass: upgrading existing branches, deferred replay during squash operations, and rename/history handling for the second backing column. I also left two smaller comments about form help text and the field-deletion warning.

I’m requesting changes for now.

bctiemann:
Thanks for the thorough review, @pheus! Pushed 14d6bbb addressing all five points:

1. **Branch-upgrade heal gap** — added `heal_branch()` in `mixin_migration.py`, wired to netbox-branching's own `post_migrate` signal (the one `Branch.migrate()` actually fires, since it never triggers Django's core `post_migrate`). This runs the existing `heal_all_cots()` pass against the branch's own connection inside `activate_branch()`, so a branch provisioned before this release now gets the `_title` column added to its own schema. Added a regression test (`BranchUpgradeHealTestCase`) that simulates a pre-upgrade branch by dropping the column from only the branch's schema and asserts `heal_branch()` restores it there without touching main's.

2. **Deferred replay dropping the title value** — `_apply_deferred_co_field()` now matches and applies the `<name>_title` key independently of the base URL key, since the two can be set independently. Added URL fields (with both URL and title values) to the shared `test_comprehensive_merge_and_revert` test, which runs under both the iterative and squash strategies, plus to `MissingFieldTypesTestCase`.

3. **Naive title-column rename + missing audit-key rewrite** — extracted the conflict-aware rename logic already used by `_schema_alter_field()` into a shared `_alter_column_with_rename_conflict_resolution()` helper, and reused it for the title column, so an independent-rename conflict (branch renames A→B, main independently renames A→C) resolves the title column the same way it already resolves the primary column. Also added a second `_rename_objectchange_field_key()` call for the title key. Added two regression tests: a straightforward rename that replays/reverts both values, and a rename-conflict test mirroring the existing `test_sequential_renames_both_sides_merge` pattern but asserting the title column converges correctly too.

4. **Missing help_text** — `URLFieldType.get_form_fields()` now renders the field's configured description as help_text on the URL input, and the title input has its own help_text.

5. **Deletion-impact preview omitting title-only objects** — both the count and the dependent-objects list in `CustomObjectTypeFieldDeleteView` now match on URL-set OR title-set, since either can hold a value independently.

netbox_custom_objects/__init__.py
```diff
@@ -88,6 +109,12 @@ def _register_branching_hooks_once():
     for sig in (pre_merge, post_merge, pre_sync, post_sync, pre_revert, post_revert):
         sig.connect(_reset_deferred_co_field_data, weak=False)
 
+    try:
+        from netbox_branching.signals import post_migrate as branching_post_migrate
+        branching_post_migrate.connect(_heal_branch_on_migrate, weak=False)
```
pheus, line 117: Thanks for adding the branch-aware heal. I think the trigger is still missing.

Because this change adds no migration, an existing branch will not be marked as having pending migrations, so `Branch.migrate()` and this signal normally won't run. The new test calls `heal_branch()` directly and bypasses that path.

Could we add a migration-visible trigger, or heal existing branches from the main upgrade path, and cover that flow end to end?

netbox_custom_objects/models.py
```diff
@@ -3516,6 +3584,25 @@ def save(self, *args, **kwargs):
                     else:
                         _schema_add_field(self, model, schema_editor, schema_conn)
                         _apply_deferred_co_field(self)
```
pheus: `_apply_deferred_co_field()` now replays `<name>_title`, but it runs before that column is added below. A deferred title would therefore issue an `UPDATE` against a column that doesn't exist yet.

Could we add both backing columns first and replay the deferred values afterward? The comment below should be updated as well.

netbox_custom_objects/models.py
```diff
@@ -2811,46 +2872,53 @@ def clean(self):
...
+        ):
             raise ValidationError(
-                {"type": _("Cannot change a field's type to or from coordinates after creation.")}
+                {"type": _(
+                    "Cannot change a field's type to or from a multi-column type "
+                    "(coordinates, URL) after creation."
+                )}
             )
 
         # Guard against backing-column name collisions.
```
pheus, line 2977: Could we also handle collisions that already exist before this upgrade?

An existing Custom Object Type may already have a URL field named `website` and a separate field named `website_title`. The new validation prevents creating that combination going forward, but it doesn't protect the upgrade path, where both fields can map to the same backing column.

Could the heal step detect this and fail safely with an actionable message? A regression test for an existing collision would help.

pheus (changes requested):
Thanks for the follow-up.

I found three remaining upgrade and replay issues: existing branches still need a reliable trigger for the schema heal, deferred title values are applied before the backing column exists, and pre-existing `<url>_title` field collisions need to be handled safely.

I’ve left the details inline. I’m requesting changes for now, but this looks close.

netbox_custom_objects/__init__.py
bctiemann: Fixed in f53799f. Added `heal_all_branches()` (`mixin_migration.py`), wired unconditionally into the main upgrade path — the `post_migrate` signal handler in `__init__.py` and the `upgrade_custom_objects` management command — rather than relying on any branch-specific migration signal. That's the same path an admin already runs after upgrading, so no new manual step is introduced. `heal_branch()` / `_heal_branch_on_migrate` (netbox-branching's own `post_migrate`) remain as a secondary/defense-in-depth path for a future release that does ship a real migration alongside a schema change.

Added an end-to-end regression test, `test_upgrade_custom_objects_command_heals_existing_branch`, that drives the actual `manage.py upgrade_custom_objects` command (not `heal_branch()` directly) against a branch with the title column dropped from only its own schema, and asserts it gets healed.

I did also consider your "migration-visible trigger" option and looked into what that would take: shipping a real (data) migration would get existing branches flagged `PENDING_MIGRATIONS` via netbox-branching's own `check_pending_migrations`, but applying it isn't automatic on that side — it only happens via an explicit admin action (a "Migrate" button, single or bulk), and a branch in that state is blocked from normal use until someone does. For a fully backward-compatible additive column, that felt like a more disruptive upgrade path than healing silently and automatically, so I went with the main-upgrade-path trigger instead. Happy to reconsider if you feel differently.

netbox_custom_objects/models.py
bctiemann: Fixed in f53799f. Reordered `CustomObjectTypeField.save()`'s adding-branch so the title column is added *before* `_apply_deferred_co_field()` runs, not after — and updated the stale comment that claimed the title column had no deferred-replay support.

Added a focused regression test, `test_title_column_exists_before_deferred_title_replay_runs`, that populates `_deferred_co_field_data` directly and then creates the field — reproducing the exact ordering scenario without depending on netbox-branching's squash-merge timing to trigger it naturally.

netbox_custom_objects/models.py
bctiemann: Fixed in f53799f. Added `detect_backing_column_collisions()`, shared with this `clean()` guard, and wired it into `heal_cot()` so a pre-existing collision (e.g. a plain field literally named `<url_field>_title` from before this validation existed) is surfaced as an actionable `backing_column_collision` warning instead of proceeding silently.

Added `BackingColumnCollisionTestCase` with 5 tests: detecting a pre-existing url/title collision and a coordinates/latitude collision, confirming no false positive for a normal url field or two unrelated plain fields, and confirming `heal_cot()` actually surfaces the warning.

netbox_custom_objects/mixin_migration.py
```diff
@@ -274,3 +306,100 @@ def heal_all_cots(verbosity=1, dry_run=False):
...
+    from netbox_branching.models import Branch  # noqa: PLC0415
+
+    no_schema_statuses = (
+        BranchStatusChoices.NEW,
+        BranchStatusChoices.PROVISIONING,
+        BranchStatusChoices.ARCHIVED,
+    )
+
+    total = healed = warnings = 0
+    for branch in Branch.objects.exclude(status__in=no_schema_statuses):
```
pheus, line 427: Could we validate and isolate each branch before healing it?

`FAILED` can mean provisioning failed, in which case the schema has already been dropped. Since the branch connection uses `<branch>,<main>` as its search path, those queries can then fall through to main. This also includes branches with pending migrations, where the current ORM may not match the branch schema yet.

Please verify the schema exists, skip branches with a pending migration plan, and handle failures per branch. The branch `post_migrate` hook can heal skipped branches after migration.

netbox_custom_objects/models.py
```diff
@@ -285,14 +302,100 @@ def _apply_deferred_co_field(field_instance):
...
+                collisions.append({
+                    "field": field.name,
+                    "other": sibling.name,
+                    "column": column,
+                    "message": (
+                        f"Custom Object Type {cot.name!r}: field {field.name!r} "
+                        f"({field.type}) and field {sibling.name!r} ({sibling.type}) "
+                        f"both map to backing column {column!r}. This predates "
+                        f"validation that now blocks creating this combination; "
+                        f"rename one of the two fields to resolve the collision, "
```
pheus: Could we make the recovery guidance more specific? Renaming either field is not safe.

For an existing `website` / `website_title` pair, renaming `website` also renames the shared `website_title` column. That moves the sibling field's data into the URL title and leaves the sibling without its own column.

The safe path is to rename the field whose own name matches the colliding column, then rerun the heal. A data-preservation test for that recovery flow would be helpful.

docs/field-attributes.md
```diff
@@ -107,3 +107,32 @@ Behaviour:
...
+- **REST API.** The pair is exposed as two flat fields, `<name>` and `<name>_title`.
+- `Must be unique` and `Default` continue to apply to the URL value itself, exactly
+  as for any other `url` field.
+- **CSV import.** Bulk CSV import only populates the URL value; the title column is
+  not importable via CSV (the same limitation applies to `coordinates` fields' backing
+  columns).
+- **Upgrading from an older release.** Existing `url` fields gain the `<name>_title`
+  column automatically on the next `manage.py migrate` (or immediately via
+  `manage.py upgrade_custom_objects`). On a NetBox Branching branch, it's healed the
+  next time that branch itself is migrated.
```
pheus: Could we update this sentence to match the new upgrade path? Existing branch schemas are now healed by the main `migrate` or `upgrade_custom_objects` run; the branch migration hook is only the secondary path.

netbox_custom_objects/mixin_migration.py
```diff
@@ -110,8 +126,23 @@ def heal_cot(cot, verbosity=1, dry_run=False):
     added = []
     warned = []
 
+    # Detect pre-existing field-name collisions with a multi-column type's
+    # synthesized backing column (e.g. a plain field literally named
+    # "<url_field>_title" predating the validation that now blocks this).
+    # Independent of DB introspection -- purely a field-definition check --
+    # so it runs even if the table itself can't be introspected below.
+    from netbox_custom_objects.models import detect_backing_column_collisions  # noqa: PLC0415
```
pheus: Could we keep only the optional `netbox_branching` imports local?

I don't see an import-cycle or app-loading reason for delaying `detect_backing_column_collisions`, and `django.apps.apps` is also safe to import at module scope. Moving those to the regular imports would make it clearer that the remaining lazy imports are specifically required for the optional Branching integration.

pheus (changes requested):
Thanks for the follow-up. The automatic branch trigger and deferred replay ordering look good now, and surfacing existing column collisions is a useful improvement.

I found two remaining upgrade-safety issues: the branch sweep can operate against a missing or outdated schema, and the collision warning currently suggests a recovery path that can move another field's data. I also left one small documentation correction.

I’m requesting changes for those upgrade cases, but the main URL/title implementation looks close.

docs/field-attributes.md
bctiemann: Corrected to reflect the primary nature of the main upgrade path, with the branch migrate as a secondary safeguard.

netbox_custom_objects/models.py
bctiemann: Added a `safe_to_rename` key which tracks which sub-field is to be renamed; fixed the message and added a test.

netbox_custom_objects/mixin_migration.py
bctiemann: Validated schemas individually and skipped missing ones. Defers to netbox-branching's own per-branch `post_migrate` hook.

netbox_custom_objects/mixin_migration.py
```diff
@@ -274,3 +327,147 @@ def heal_all_cots(verbosity=1, dry_run=False):
...
+            logger.warning(
+                "upgrade_custom_objects: skipping branch %r (id=%s, status=%s) -- schema %r "
+                "does not exist (likely a failed or interrupted provision); it will be healed "
+                "once it has a live schema.",
+                branch.name, branch.pk, branch.status, branch.schema_name,
+            )
+            warnings += 1
+            continue
+
+        if branch.pending_migrations:
```
pheus: Could we wrap the pending-migration check and heal in one per-branch `try` and close `connections[branch.connection_name]` in `finally`?

`pending_migrations` opens a branch connection and builds a `MigrationExecutor`. NetBox Branching recently fixed the same loop because retaining those connections during `manage.py migrate` could exhaust PostgreSQL resources with many branches. Keeping the check inside the `try` would also prevent one broken branch from aborting the rest of the sweep.

pheus (changes requested):
Thanks for the follow-up. 

I found one remaining Branching issue: the new branch sweep needs the same per-branch connection cleanup and exception isolation used by NetBox Branching itself.

Otherwise this looks close.

netbox_custom_objects/mixin_migration.py
bctiemann: Added a new `connections[branch.connection_name].close()` in `finally`.

pheus (approved):

PR #644, opened 2026-08-04
Fixes #639: Respect filter_logic in generated filtersets

### LINKED ISSUE #639: Field lookup expressions ignored
### Plugin Version

0.6.0

### NetBox Version

4.5.3

### Python Version

3.12.3

### Steps to Reproduce

1. Create custom object type
2. Add field f1 of type text
3. Create new object with f1=foobar
4. Create new object with f1=barfoo
5. Try query it with parameter f1__isw=foo

### Expected Behavior

Only object 1 to be returned, as it used to work in version 0.4.4

### Observed Behavior

All objects returned.

bctiemann:
## Summary

- CustomObjectTypeField.filter_logic has existed since this plugin's earliest commits, but was never actually read by filterset generation. build_filter_for_field() always built the base text/longtext/url filter with a hardcoded lookup_expr="icontains", regardless of what filter_logic was set to.
- This silently broke suffix lookups (e.g. ?f1__isw=foo, ?f1__iew=foo, ?f1__ic=foo): NetBox core's BaseFilterSet.get_additional_lookups() only augments a filter whose own lookup_expr is one of a small fixed set ('exact', 'iexact', 'in', 'contains') -- "icontains" is not among them -- so those suffix filters were never registered at all, and the unrecognized query param was silently ignored, returning every row unfiltered instead of raising an error or actually filtering.
- The fix mirrors NetBox core's own CustomField.to_filter() convention exactly, since this plugin's filter_logic field was modeled on it in the first place:
  - "loose" (the default): base filter keeps using icontains, unchanged from current behavior.
  - "exact": drops the lookup_expr override (CharFilter's own default is "exact"), which is what makes the base filter eligible for get_additional_lookups() to generate the __n/__ic/__isw/__iew/__ie/__empty/__regex suffix filters.
  - "disabled": skip building a filter for the field entirely, matching NetBox core's own universal (type-agnostic) skip for disabled custom fields. This was also silently ignored before; now it does what the UI has always documented it as doing.
- Applies to TYPE_TEXT, TYPE_LONGTEXT, and TYPE_URL -- the same three types NetBox core's own filter_logic switch covers. TYPE_JSON keeps its existing icontains-only behavior (no core-NetBox "exact" analog to mirror).

## Test plan

- [x] Added TextFieldFilterLogicTestCase covering: loose bare-filter unchanged, loose __isw not registered (documents the intentional NetBox-matching limit), exact bare-filter is exact match, exact __isw/__iew/__ic all correctly filter, disabled has no filter registered at all, disabled query param has no effect.
- [x] Confirmed the new tests actually fail without the fix (5 of 8 fail against the unmodified code), and pass with it.
- [x] Full test_filtersets.py module (128 tests) passes clean.
- [x] test_api.py, test_forms.py, test_views.py pass clean apart from 4 pre-existing, unrelated local-environment failures (netbox_branching importable but not enabled in this environment's PLUGINS list).
- [x] ruff check clean.

### DIFF
--- a/netbox_custom_objects/filtersets.py
+++ b/netbox_custom_objects/filtersets.py
@@ -1,3 +1,4 @@
+import copy
 import django_filters
 from dataclasses import dataclass
 from decimal import Decimal, InvalidOperation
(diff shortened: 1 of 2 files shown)


bctiemann:
@claude /review - report only on potential concerns.

arthanson (changes requested):
this still fails the scenario given in the bug report, I think the test:
```
def test_loose_field_isw_suffix_not_registered(self):
      fs = self._filterset({'loose_field__isw': 'foo'})
      self.assertEqual(fs.qs.count(), 2)   # unfiltered
```
actually shows that it will fail.

arthanson (approved):

PR #648, opened 2026-08-07
Closes #658: Harden polymorphic multiobject through-model registration against concurrent creation/reads

### LINKED ISSUE #640: Deleting a Custom Object with a multiobject field raises ValueError (recurrence of #477 in v0.6.0)
### Plugin Version

v0.6.0

### NetBox Version

v4.6.6

### Python Version

3.14

### Steps to Reproduce

1. Create a Custom Object Type, e.g. name=`internal_service`,<br>verbose_name="Internal Service", slug=`internal-services`.
2. Add a field to it:
   * name: `depends_on`
   * type: `multiobject`
   * either self-referencing (related type = the same Custom Object<br>Type) or referencing another Custom Object Type — any<br>multiobject field reproduces it, self-reference is not required.
3. Create one instance of this Custom Object Type via the UI or via<br>`POST /api/plugins/custom-objects/internal-services/`.
4. Navigate to the object's detail page and click "Delete", OR send<br>`GET /plugins/custom-objects/internal-services/<id>/delete/`<br>directly (this is the confirmation page — no POST/actual deletion<br>is even required to trigger it).
5. Alternatively: send `DELETE /api/plugins/custom-objects/internal-services/<id>/`<br>via the REST API directly (raw HTTP, no client library).

### Expected Behavior

The detail page renders normally and shows the `depends_on` field<br>(empty or populated), exactly as the REST API already does for the<br>same object.

### Observed Behavior

Server Error. Full exception:

```
FieldDoesNotExist

Table<A>Model has no field named 'depends_on'
```

This happens on a **freshly created field that has never been****<br>****edited**, and persists across:

* restarting the `netbox` and `netbox-worker` containers
* a full `docker compose down && docker build --no-cache && docker compose up -d` (i.e. a completely fresh Python process, ruling out<br>any in-memory cache)

We confirmed directly via `./manage.py nbshell`, on a fresh process<br>right after the rebuild above, that the dynamically generated model<br>class genuinely lacks the field:

```python
>>> from netbox_custom_objects.models import CustomObjectType
>>> cot = CustomObjectType.objects.get(id=<A>)
>>> model = cot.get_model()
>>> [f.name for f in model._meta.get_fields()]
['id', 'created', 'last_updated', 'owner', 'service_id', 'product', 'status', 'tag', 'info_yetiforce', 'tags', 'bookmarks', 'contacts', 'journal_entries', 'subscriptions', 'tagged_items']
# 'depends_on' is simply absent from this list, despite the field
# existing and being correctly returned by the REST API for the same
# object at the same time.
```

At container/process startup, the following warnings appear in the<br>logs, once per Custom Object Type that has such a polymorphic field:

```
RuntimeWarning: Model 'netbox_custom_objects.through_custom_objects_<N>_depends_on' was already registered. Reloading models is not advised as it can lead to inconsistencies, most notably with related models.
  _apps.register_model(APP_LABEL, through_model)
```

This suggests the intermediate/through model backing the polymorphic<br>relation is registered more than once during app startup, and the<br>duplicate `apps.register_model()` call (which Django treats as a<br>no-op for an already-registered model name) silently keeps whichever<br>class got registered *first* — which appears to be missing the field<br>in question.

### Empirically confirmed trigger condition

We tested this extensively across 4 different Custom Object Types in<br>the same NetBox instance, changing only the field configuration:

| related_object_types_input length | Detail view result |
| -- | -- |
| 1 (non-polymorphic, single `app_label`+`model`) | Works correctly, every time |
| 2 | Crashes, every time |
| 3 | Crashes, every time |

This held regardless of:

* whether one of the 2-3 types was the field's own Custom Object Type<br>(self-reference) or not
* how many times the field had previously been deleted and recreated<br>(we confirmed it on fields created for the very first time, never<br>touched again, immediately after creation)
* whether `is_cloneable`, `weight`, or other non-structural metadata<br>had been set at creation time vs patched afterward

The only reliable fix we found was reducing every such field to<br>exactly **one** related type (converting it from<br>`is_polymorphic: true` with N types, to a plain non-polymorphic<br>`object`/`multiobject` field with a single `app_label`+`model`).

### Notes / Root cause hint

This looks related to (but may be a distinct or regressed variant of)<br>the previously-reported and closed netboxlabs/netbox-custom-objects#384 ("Model generation<br>problem: fields disappear when Custom Object Types reference each<br>other"), fixed via PR [#530](<https://github.com/netboxlabs/netbox-custom-objects/issues/530>) in milestone v0.5.2 for a narrower<br>repro (two Custom Object Types referencing each other via single-type<br>`object` fields, in an NGINX Unit multi-process environment, symptom<br>being a field intermittently present/absent). Our case:

* reproduces with `is_polymorphic: true` fields specifically (not<br>plain single-type `object` fields)
* reproduces deterministically (100% of the time), not intermittently
* reproduces even with only ONE Custom Object Type in the system<br>having such a field (no mutual/circular reference between multiple<br>types is required — a single type with a 2-type polymorphic field<br>pointing at itself + one other type is enough)
* we are on v0.6.0, newer than the v0.5.2 milestone netboxlabs/netbox-custom-objects#384 was fixed in

We were unable to find a way to trigger `clear_model_cache()` or<br>force a correct rebuild without reducing the field to a single type;<br>happy to provide DB schema dumps, the exact API payloads used to<br>create every field, and/or grant reproduction access if useful for<br>maintainers.

### Workaround found

Model dependencies using **multiple separate non-polymorphic fields**<br>(one per related type, e.g. `depends_on_service_a`,<br>`depends_on_service_b`) instead of one polymorphic field with several<br>related types. Each field individually works reliably; the modeling<br>goal (an object depending on instances of several different Custom<br>Object Types) is achieved by having several fields instead of one.

### LINKED ISSUE #658: Race and deadlock in polymorphic multiobject through-model registration
### Plugin Version

main (unreleased, found in development past v0.6.0)

### NetBox Version

v4.6.6

### Python Version

3.12

### Steps to Reproduce

While investigating netboxlabs/netbox-custom-objects#640, I found two independent, concretely reproducible concurrency bugs in `MultiObjectFieldType.create_polymorphic_m2m_table()` (called once, when a polymorphic `multiobject` field is first created). Neither reproduces netboxlabs/netbox-custom-objects#640's specific reported symptom (see the discussion on PR netboxlabs/netbox-custom-objects#648) -- these are separate bugs in the same function, verified with deterministic regression tests added in PR netboxlabs/netbox-custom-objects#648:

1. `PolymorphicMultiObjectConcurrencyTestCase.test_forced_registration_interleaving_stays_consistent` -- forces a concurrent `get_model(no_cache=True)` call to land inside the writer's build-then-repoint window (bug 1 below), via a mocked `apps.register_model()` hook.
2. `PolymorphicMultiObjectConcurrencyTestCase.test_concurrent_double_submit_does_not_deadlock` -- two threads calling `CustomObjectTypeField.objects.create()` for the identical `(custom_object_type, name)` at once (bug 2 below).

**Bug 1: registration-before-repoint race against a concurrent reader**

`create_polymorphic_m2m_table()` built and registered a through-model class with Django's app registry, and only *afterward* repointed its `source` FK at the caller's model class -- all without holding `CustomObjectType._global_lock`. `CustomObjectType.get_model()`'s own reuse-or-create check for polymorphic through models (`_after_model_generation()`) *is* lock-protected on its own side. A concurrent `get_model(no_cache=True)` call could land in the writer's build-then-repoint window, find the through model already registered, and repoint `source` at its own (different, but table-equivalent) model instance instead -- leaving the through's FK and whatever `get_model()` subsequently caches pointing at two different Python classes for the same table.

**Bug 2: lock-ordering deadlock on a double-submit**

Fixing bug 1 by holding `_global_lock` across the whole `create_polymorphic_m2m_table()` call (including the table-existence probe and `schema_editor.create_model()` DDL) introduces a *different* real deadlock: two threads racing to create the same field (e.g. a doubly-clicked "save" button, or a retried request) each build+register a through model for the same physical table before either knows which one will win the `(name, custom_object_type)` `UniqueConstraint`. Whichever thread's `schema_editor.create_model()` runs second blocks at the Postgres level waiting on the first thread's uncommitted `CREATE TABLE` (same table name) to resolve -- but the first thread's own `CustomObjectTypeField.save()` needs `_global_lock` again in `clear_model_cache()` before it can commit and release that wait.

### Expected Behavior

A polymorphic `multiobject` field's through-model registration should stay consistent under concurrent field creation and reads, and concurrent field-creation attempts should not be able to deadlock.

### Observed Behavior

**Bug 1**: the through model's `source` FK and the model class `get_model()` subsequently returns/caches can point at two different Python classes for the same table. This class-identity mismatch produces `ValueError: Cannot query "X": Must be "TableYModel" instance.` or a `RecursionError` when later deleting an object through the affected relation.

**Bug 2**: confirmed live via `pg_stat_activity` -- one thread idle-in-transaction waiting on `_global_lock` (needed again in `clear_model_cache()` to commit), the other actively blocked on `Lock/transactionid` executing `CREATE TABLE` for the same physical table, waiting on the first thread's uncommitted transaction. Neither thread can make progress; the request hangs.

### Relationship to netboxlabs/netbox-custom-objects#640

netboxlabs/netbox-custom-objects#640 reports a different, more severe symptom (a deterministic `FieldDoesNotExist` on a fresh, single, non-concurrent process) that I was unable to reproduce on current `main` -- see the investigation notes on PR netboxlabs/netbox-custom-objects#648. This issue is scoped narrowly to the two concurrency bugs above, which PR netboxlabs/netbox-custom-objects#648 fixes with regression coverage. netboxlabs/netbox-custom-objects#640 remains open pending further reproduction details.

bctiemann:
## Scope (updated per review)

This PR does **not** close netboxlabs/netbox-custom-objects#640. I investigated whether it does and could not reproduce netboxlabs/netbox-custom-objects#640's reported symptom (a deterministic `FieldDoesNotExist` on a fresh, single, unconcurrent process) on current `main` -- see "Investigation of netboxlabs/netbox-custom-objects#640" below. This PR instead fixes two independently real, verified concurrency bugs in the same function that I found while investigating (now filed separately as netboxlabs/netbox-custom-objects#658), and is scoped as registry-race hardening. netboxlabs/netbox-custom-objects#640 remains open pending further reproduction details.

## Summary

* `MultiObjectFieldType.create_polymorphic_m2m_table()` (called once, when a polymorphic `multiobject` field is first created) built and registered a through-model class with Django's app registry, and only *afterward* repointed its `source` FK at the caller's model class -- all without holding `CustomObjectType._global_lock`. A concurrent `get_model(no_cache=True)` call (lock-protected only on its own side) could land in that window, find the through model already registered, and repoint `source` at its own (different, but table-equivalent) model instance instead -- leaving the through's FK and whatever `get_model()` subsequently caches pointing at two different Python classes for the same table. Fixed by holding `_global_lock` across the build+register+repoint sequence.
* That same lock, if held across the table-existence probe/DDL that follows, creates a *different* real deadlock: two threads racing to create the same field (e.g. a retried request) each build+register a through model for the same physical table before either knows which one will win the `(name, custom_object_type)` `UniqueConstraint`. Whichever thread's `schema_editor.create_model()` runs second blocks at the Postgres level waiting on the first thread's uncommitted `CREATE TABLE` (same table name) to resolve -- but the first thread's own `save()` needs `_global_lock` again in `clear_model_cache()` before it can commit and release that wait. Fixed by narrowing the lock to stop before the DDL.

## Investigation of netboxlabs/netbox-custom-objects#640

Per review, I tried to reproduce netboxlabs/netbox-custom-objects#640's actual reported symptom directly: a **fresh, single Python process**, no concurrency, self-referencing polymorphic multiobject field with 2-3 related types, `FieldDoesNotExist`/model missing the field on the very first `get_model()` call. I could not reproduce it on current `main`:

* Created a COT with a self-referencing 3-type polymorphic multiobject field in one process, then in a **separate, fresh process** (no shared in-memory state) issued real `GET` requests against both the object detail page and the delete-confirmation page (the exact reported repro steps) -- both returned 200, no crash, both with and without this PR's lock fix.
* `model._meta.get_fields()` does not list `depends_on` for a polymorphic multiobject field regardless -- these fields are implemented as a plain descriptor (`PolymorphicM2MDescriptor`), not a real Django `Field`, so they were never expected to appear there. That's consistent with the snippet in netboxlabs/netbox-custom-objects#640's own report, but isn't itself evidence of breakage (`obj.depends_on.set(...)` and the actual views work fine via descriptor access).

Since `main` has moved substantially since v0.6.0 (the version netboxlabs/netbox-custom-objects#640 was filed against) via other fixes, it's plausible netboxlabs/netbox-custom-objects#640's actual root cause was already resolved by an unrelated change, or is specific to the reporter's serving environment (NGINX Unit, Python 3.14, multi-worker-process). I don't have a way to confirm either without reproduction access. Given I can't produce a regression that demonstrates netboxlabs/netbox-custom-objects#640's actual symptom, I'm leaving netboxlabs/netbox-custom-objects#640 open rather than closing it here, per the suggested alternative.

## Test plan

* `PolymorphicMultiObjectConcurrencyTestCase.test_forced_registration_interleaving_stays_consistent`: deterministically forces the registration-before-repoint race via a mocked `apps.register_model()` hook. Fails against the unfixed code (source FK / `get_model()` class mismatch), passes with the fix. Asserts both threads actually completed (not just timed out) before checking results.
* `PolymorphicMultiObjectConcurrencyTestCase.test_concurrent_double_submit_does_not_deadlock`: two threads independently call `CustomObjectTypeField.objects.create()` for the identical `(custom_object_type, name)`. Hangs against the wide-lock version (confirmed via `pg_stat_activity`: one thread idle-in-transaction waiting on the Python lock, the other actively blocked on Postgres waiting for the first's uncommitted `CREATE TABLE`); completes in \~1.5s with the narrowed lock.
* `PolymorphicMultiObjectConcurrencyTestCase.test_field_creation_racing_concurrent_readers_yields_consistent_through_model` / `..._with_two_type_setup_...`: real-concurrency variants (12 looping `get_model()` readers against real field creation), including the reported two-type Custom Object setup, through the public `CustomObjectTypeField.objects.create()` path.
* Delete-confirmation `GET` coverage: the reported UI path (issue netboxlabs/netbox-custom-objects#640, step 4) doesn't call `obj.delete()`, which itself realigns the through's FK before Django's collector runs -- a plain `GET` exercises the unrepaired state directly.
* `ruff check` clean on both changed files.
* Full `test_schema_operations.py` passes (4 pre-existing, unrelated errors from a `netbox_branching` import artifact under the non-branching test configuration, present on unmodified `main` too).


bctiemann:
@claude /review and limit response to potential problems of medium or higher severity.

bctiemann:
I fixed item 2 (reduced the timeout to avoid adding to the CI runtime), but left the other two alone—especially item 3 which basically would amount to neutralizing the very race-condition fix we're trying to implement.

netbox_custom_objects/tests/test_schema_operations.py
```diff
@@ -275,3 +281,212 @@ def test_coordinates_field_delete_drops_both_columns(self):
...
+            return result
+
+        writer_result = {}
+
+        def run_writer():
+            threading.current_thread().name = 'W'
+            field_type = FIELD_TYPE_CLASS[CustomFieldTypeChoices.TYPE_MULTIOBJECT]()
+            try:
+                with connection.schema_editor() as schema_editor:
+                    field_type.create_polymorphic_m2m_table(field, writer_model, schema_editor)
```
pheus, line 453: Could we add a regression through the public field-save path using the reported two-type Custom Object setup? This test starts from an already-persisted field, unregisters its through model, and calls the private helper directly. On creation, the helper runs before the field row is saved, so another connection cannot yet discover the new field.

netbox_custom_objects/tests/test_schema_operations.py
```diff
@@ -275,3 +281,212 @@ def test_coordinates_field_delete_drops_both_columns(self):
...
+                reader_done.set()
+                connection.close()
+
+        with patch.object(apps, 'register_model', side_effect=ordered_register_model):
+            t_w = threading.Thread(target=run_writer, name='W')
+            t_r = threading.Thread(target=run_reader, name='R')
+            t_w.start()
+            t_r.start()
+            t_w.join(timeout=10)
+            t_r.join(timeout=10)
```
pheus, line 478: Could we assert that both threads completed after these joins? If either times out, the result dictionaries can remain empty, and the fresh `get_model()` call below may let the test pass without completing the intended interleaving.

netbox_custom_objects/tests/test_schema_operations.py
```diff
@@ -275,3 +281,212 @@ def test_coordinates_field_delete_drops_both_columns(self):
...
+        source_field = through_model._meta.get_field('source')
+        self.assertIs(
+            source_field.remote_field.model, final_model,
+            "the registered through model's source FK must point at the model class "
+            "get_model() currently returns -- a mismatch here is issue #640",
+        )
+
+        obj = final_model.objects.create(name='Instance 1')
+        obj.depends_on.set([Site.objects.create(name='Force Site', slug='force-site')])
+        obj.delete()  # Must not raise ValueError: "Cannot query ...: Must be ... instance."
```
pheus, line 526: Could we also cover the delete-confirmation GET? The class-identity assertion above exercises the underlying registry invariant, but `obj.delete()` does not cover the reported UI path because the Custom Object delete override realigns the through-model FK before Django's collector runs.

pheus (changes requested):
Thanks for taking this on. The locking change itself looks reasonable, but I’d like the regression to cover the actual Custom Object lifecycle before approval. The current test uses a persisted field and calls the private schema helper directly, then verifies deletion through a path that repairs the FK first. I’ve left inline notes on matching the reported save and confirmation workflow, plus one on making the thread assertions deterministic.

netbox_custom_objects/tests/test_schema_operations.py
bctiemann: Done, though development of this test induced a deadlock requiring an alternative concurrency-safe variant.

netbox_custom_objects/tests/test_schema_operations.py
```diff
@@ -277,6 +286,326 @@ def test_coordinates_field_delete_drops_both_columns(self):
...
+        # Without the fix, this reliably produces a mismatch: reader's model
+        # cached while writer's model is left on the through's "source" FK,
+        # or vice versa.
+        final_model = CustomObjectType.objects.get(pk=cot.pk).get_model()
+        through_model = apps.get_model(APP_LABEL, field.through_model_name)
+        source_field = through_model._meta.get_field('source')
+        self.assertIs(
+            source_field.remote_field.model, final_model,
+            "the registered through model's source FK must point at the model class "
+            "get_model() currently returns -- a mismatch here is issue #640",
```
pheus: I’m still missing the connection to #640’s reported failure. The issue reproduces as a `FieldDoesNotExist` after rebuilding the registry in a fresh process, while this test covers an in-process class-identity race during the initial field save.

Could we add a regression for the reported fresh-startup failure that fails without this change, or scope this PR as separate registry hardening rather than closing #640?

netbox_custom_objects/field_types.py
```diff
@@ -1787,23 +1787,41 @@ def create_polymorphic_m2m_table(self, field_instance, model, schema_editor):
...
+        # ModelBase metaclass inside generate_model() below, but before its
+        # "source" FK is repointed at `model` on the next line -- and race to
+        # point the registered class's FK at its OWN (different) model
+        # instance. Whichever thread's mutation and whichever thread's
+        # get_model() cache-write happen last aren't guaranteed to be the
+        # same thread, leaving the through's "source" FK and the cached model
+        # class mismatched. Confirmed live under concurrent load:
+        # intermittent "ValueError: Cannot query 'X': Must be 'TableYModel'
+        # instance." and RecursionError (issue #640).
+        with CustomObjectType._global_lock:
```
pheus, line 1819: Could we limit this lock to building and registering the through model and repointing its `source` FK, then perform the table check and DDL after releasing it?

A concurrent attempt to create the same field can otherwise hold `_global_lock` while waiting on the first transaction’s uncommitted `CREATE TABLE`. The first save then needs the same lock in `clear_model_cache()` before it can commit, leaving both operations blocked.

pheus (changes requested):
Thanks for the additional coverage. I think this is now down to two blockers: the regression still does not demonstrate #640’s fresh-process `FieldDoesNotExist`, and the lock currently spans transactional DDL, which introduces a credible deadlock path.

bctiemann:
Thanks -- both addressed.

**On the deadlock (`field_types.py:1808`):** confirmed, and reproduced it directly -- two threads double-submitting field creation for the same `(custom_object_type, name)` hung a real test run, and `pg_stat_activity` showed exactly the cycle you described: one thread idle-in-transaction waiting on `_global_lock` (needed again in `clear_model_cache()` to commit), the other actively blocked on Postgres waiting for the first thread's uncommitted `CREATE TABLE` on the same physical table. Narrowed the lock to just the build+register+repoint step; the table-existence probe/DDL now runs after it's released. Added `test_concurrent_double_submit_does_not_deadlock`, which hangs against the old wide-lock version and passes in ~1.5s with the fix.

**On #640's actual symptom:** I tried to reproduce it directly -- a fresh single process, no concurrency, self-referencing 2-3 type polymorphic multiobject field, then a real `GET` to both the object detail page and the delete-confirmation page from a *separate* fresh process. Both returned 200 on current `main`, with or without this PR's fix. I couldn't get `FieldDoesNotExist` to happen at all this way.

Rather than keep asserting this closes #640 without being able to demonstrate the reported failure, I've rescoped the PR: title/description updated, dropped the `Closes:` keyword, and left #640 open. What's left is two independently real, verified concurrency bugs in the same function (the original registration-before-repoint race, plus the deadlock above) -- registry-race hardening, not a fix for #640's specific report. Happy to revisit #640 itself separately if reproduction access or more detail becomes available; my best guess is it's either already fixed by something else merged since v0.6.0, or specific to the reporter's serving environment (NGINX Unit, Python 3.14, multi-worker-process), neither of which I can confirm from here.

netbox_custom_objects/field_types.py
bctiemann: Narrowed the lock to release before the DDL. New test (verified locally to fail before this commit and pass after) added.

bctiemann:
Because this PR does not actually fix #640 directly, it's not completely clear to me whether we should merge this anyway because it's not in response to any specific raised bug; but it does at least eliminate a possible path for the deadlock or multi-worker race condition to manifest. We can keep #640 open and request further reproduction details.

netbox_custom_objects/tests/test_schema_operations.py
bctiemann: See below.

netbox_custom_objects/tests/test_schema_operations.py
```diff
@@ -277,6 +286,386 @@ def test_coordinates_field_delete_drops_both_columns(self):
...
+            except Exception as e:  # noqa: BLE001 - surfaced via the assertion below
+                results[key] = {'error': e}
+            finally:
+                connection.close()
+
+        t_a = threading.Thread(target=create_field, args=('A',), name='A')
+        t_b = threading.Thread(target=create_field, args=('B',), name='B')
+        t_a.start()
+        t_b.start()
+        t_a.join(timeout=15)
```
pheus, line 575: Suggestion: This test may leave both worker threads blocked when it detects the old deadlock, causing the test process to hang rather than fail cleanly. It might be worth covering the lock boundary with a safely releasable interleaving or an isolated subprocess, but I don’t think this needs to hold the PR.

bctiemann:
Filed the two concurrency bugs this PR fixes as a separate issue, #658, and updated this PR to close it instead of referencing #640 directly. #640 stays open pending further reproduction details, per the discussion above.

netbox_custom_objects/field_types.py
```diff
@@ -1787,12 +1787,41 @@ def create_polymorphic_m2m_table(self, field_instance, model, schema_editor):
...
+        # through model mid-construction here -- registered by Django's
+        # ModelBase metaclass inside generate_model() below, but before its
+        # "source" FK is repointed at `model` on the next line -- and race to
+        # point the registered class's FK at its OWN (different) model
+        # instance. Whichever thread's mutation and whichever thread's
+        # get_model() cache-write happen last aren't guaranteed to be the
+        # same thread, leaving the through's "source" FK and the cached model
+        # class mismatched. Confirmed live under concurrent load:
+        # intermittent "ValueError: Cannot query 'X': Must be 'TableYModel'
+        # instance." and RecursionError (issue #640).
```
pheus: Could we update this reference, along with the matching test references, to #658?

netbox_custom_objects/tests/test_schema_operations.py
```diff
@@ -277,6 +286,386 @@ def test_coordinates_field_delete_drops_both_columns(self):
...
+        it can genuinely deadlock rather than just race. Both threads target the same physical
+        through table, so the second thread's CREATE TABLE blocks at the Postgres level on the
+        first thread's still-open transaction; CustomObjectType._global_lock is held by the first
+        thread across that same window (with the fix in place); and if anything downstream in the
+        first thread's own save() needs that lock again (e.g. a signal handler calling
+        get_model()), neither thread can make progress -- confirmed by hanging an actual test
+        run. Real thread-scheduling luck, exercised here instead via 12 looping readers (matching
+        the existing single-type test above), cannot deadlock this way: no reader ever holds
+        transaction.atomic() open across a paused lock acquisition.
+        """
```
pheus, line 609: Suggestion: Could we update and shorten this explanation? The current implementation releases `_global_lock` before the table probe and DDL, so the statement that it is held across this window “with the fix in place” is no longer accurate. The full deadlock analysis now lives in #658; the docstring only needs to describe the behavior asserted by this test.

pheus (approved):
Thanks for separating these findings into #658.

I’m happy to approve this. I’ve left two small documentation notes: please update the remaining #640 references to #658 and correct the test explanation that still describes the previous, wider lock scope. Please also ensure the PR and final merge message close #658 rather than #640.

bctiemann:
Thanks! Both addressed in 7a47878:

- Updated the remaining #640 references (`field_types.py`, and the matching test docstrings/assertion messages in `test_schema_operations.py`) to #658 -- kept the one reference to #640 that specifically attributes the delete-confirmation-GET coverage to that issue's own numbered steps, since #658 doesn't have an equivalent.
- Shortened `test_forced_registration_interleaving_stays_consistent`'s docstring and corrected the now-inaccurate "holds `_global_lock` for that whole call" claim (true before the lock was narrowed) to describe just the build+register+repoint step, with a pointer to #658 for the full analysis.

Left the double-submit deadlock test's potential-hang risk as-is per your note that it's not blocking -- agreed a safely-releasable interleaving or subprocess isolation would be the right way to harden it, but that's more surface area than this PR needs.

PR #654, opened 2026-08-12
Fixes #647: heal mixin fields unmasked by renaming/deleting a colliding CO field

### LINKED ISSUE #647: Renaming NCO fields
### Plugin Version

0.6.0

### NetBox Version

NetBox Community v4.6.3-Docker-5.0.1

### Python Version

3.14.4

### Steps to Reproduce

Create an NCO with an owner field mapped to Object > Tenancy >Contact. 

Image

Image

Create a few of these objects. Rename the owner field to Client Contact 

Image

### Expected Behavior

The field will be renamed. 

### Observed Behavior

When opening the Custom Object, the following error occurs 

Image

```
Request Method:	GET
Request URL:	http://dev-netbox.ei.rutgers.edu/plugins/custom-objects/slbs/
Django Version:	6.0.6
Exception Type:	ProgrammingError
Exception Value:	
column custom_objects_2.owner_id does not exist
LINE 1: ..._2"."created", "custom_objects_2"."last_updated", "custom_ob...
```

On attempt to change it back, this error occurs. 

Image

bctiemann:
## Summary

- `CustomObject` inherits NetBox core's `OwnerMixin`, which contributes its own
  concrete `owner` field (FK to `users.Owner`, column `owner_id`). A user field
  named `owner` (as in this issue, an object field pointing at
  `tenancy.Contact`) shadows that mixin field for as long as the collision
  lasts — the physical column belongs to the user field, not the mixin.
- Renaming or deleting that field unmasks the mixin field on the next model
  regeneration. Its column was never created (since the column belonged to
  the user field), so the next query against the table raises
  `ProgrammingError: column ... owner_id does not exist`, exactly as
  reported.
- Adds `_heal_unmasked_base_fields()`, called right after a rename or delete,
  which adds back any base-mixin column newly unmasked by the operation
  (reusing the existing `mixin_migration._expected_base_fields()` /
  `_can_auto_add()` diff logic, made branch-aware via `schema_conn`).

## Recovering already-broken installations

This fix only prevents the crash going forward. For a COT that's already
broken (renamed/deleted before this fix was deployed), running
`manage.py upgrade_custom_objects` (or simply the next `manage.py migrate`,
which fires the existing `post_migrate` heal hook) recovers it — verified
that the plugin's pre-existing `heal_cot()` already adds the missing base
column back for an already-broken table, independent of this change.

## Test plan

- [x] Added `OwnerFieldNameCollisionTestCase` covering both the rename and
  delete paths — asserts the query succeeds afterward and the mixin's own
  column is healed back in.
- [x] Full `netbox_custom_objects` suite passes (1128 tests).
- [x] Manually verified `heal_cot()` recovers a simulated pre-existing broken
  COT (table with the mixin column missing) independent of this fix.

### DIFF
--- a/netbox_custom_objects/mixin_migration.py
+++ b/netbox_custom_objects/mixin_migration.py
@@ -90,6 +90,47 @@ def _can_auto_add(field):
 # Public API
 # ---------------------------------------------------------------------------
 
+def heal_unmasked_fields(cot, model, schema_conn):
+    """
+    Add missing columns for CustomObject mixin fields unmasked by renaming or
+    deleting a same-named user field (e.g. 'owner' shadowing OwnerMixin.owner).
+
+    Schema-connection-aware (branch-safe) counterpart to the add-column loop
+    in heal_cot(), meant to be called right after a CustomObjectTypeField
+    rename/delete rather than waiting for the next post_migrate heal pass.
+    """
+    expected = _expected_base_fields(cot, model)
+    with schema_conn.cursor() as cursor:
+        actual_cols = {
+            col.name
+            for col in schema_conn.introspection.get_table_description(cursor, model._meta.db_table)
+        }
+
+    missing = []
+    for col_name, field in expected.items():
+        if col_name in actual_cols:
+            continue
+        if not _can_auto_add(field):
+            logger.warning(
+                "heal_unmasked_fields: unmasked base column %r (field %r) on %s is not "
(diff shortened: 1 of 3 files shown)


bctiemann:
@claude /review and limit response to potential problems.

bctiemann:
@claude /review the last commit, same guardrails as before.

arthanson (approved):

Issue #42, opened 2025-07-01
Default value should be auto-populated for multi-object fields


mrmrcoleman:
### Plugin Version

commit 722bbd624c8af515dc584b719e2864cbfcc14dd0 (HEAD -> public-preview, origin/public-preview) Author: btiemann <btiemann@netboxlabs.com> Date:   Thu Jun 26 15:31:01 2025 -0400

### NetBox Version

v4.2.8-Docker-3.2.0

### Python Version

3.12

### Steps to Reproduce

* Create a custom object type X
* Add a field of type multi-object and set default values
* Create a custom object of type X

### Expected Behavior

* The `DynamicModelMultipleChoiceField` is populated in the form

### Observed Behavior

* The default values are not populated

mrmrcoleman:
Tested at:

```
NetBox: 4.4.0
Custom Objects: commit af857f546c4fd099398c4a0be6e3f29135e27bd2 (HEAD -> main, origin/main, origin/HEAD)
```

---

Fixed - will be released in 0.4.0

PR #663, opened 2026-08-17
Closes #645: Omit hidden fields from custom object create/edit and bulk edit forms

### LINKED ISSUE #645: Field UI editable value of "Hidden" not honored
### Plugin Version

0.6.0

### NetBox Version

4.6.7

### Python Version

3.12.8

### Steps to Reproduce

1. Create a new custom named my_obj.
2. Create a new text field for my_obj named "name" to represent the object's primary key.
3. Create a new text field for my_obj named "hidden" and set the "UI editable" value to "Hidden".

Image

4. Open the model form to create a new instance of my_obj.

### Expected Behavior

The hidden field is not rendered in the form as an input.

### Observed Behavior

The hidden field is rendered in the form as an input.

Image

bctiemann:
## Summary

- A `CustomObjectTypeField` with `ui_editable` set to "Hidden" was rendered as an (unlabeled-looking, but present) input on the object create/edit form and the bulk edit form, instead of being omitted entirely.
- Root cause: `get_annotated_form_field()` only ever sets `disabled = True` for any `ui_editable` value other than "Yes" -- it doesn't distinguish "Hidden" (should not be displayed at all) from "No"/read-only (should be displayed, but disabled). The templates render every field listed in `custom_object_type_field_groups` unconditionally, without checking `disabled`, so both `CustomObjectEditView.get_form()` and `CustomObjectBulkEditView.get_form()` ended up including hidden fields as still-visible inputs.
- This mirrors core NetBox's own convention (`extras.forms.mixins.NetBoxModelMixin._get_custom_fields`), which filters out `ui_editable=HIDDEN` custom fields before building the form, while leaving `NO` (read-only) fields in place as disabled.

## Changes

- Added `_hidden_field_raw_columns()` helper in `views.py`: given a queryset of fields, returns the backing model column name(s) for any field whose `ui_editable` is `HIDDEN` (handling coordinates' two-column and polymorphic-object raw-column cases).
- Both `CustomObjectEditView.get_form()` and `CustomObjectBulkEditView.get_form()` now:
  - Add those column(s) to `Meta.exclude`, so the `ModelForm` metaclass doesn't auto-generate a plain (non-disabled) field for them from `fields="__all__"`.
  - Skip hidden fields entirely in the per-field loop that builds `attrs`, so they're never added to `custom_object_type_fields`/`_field_groups`/`_rendered_names`.
- CSV bulk import already excluded hidden fields entirely (fixed under #626); this brings the interactive forms in line with the same behavior.

## Testing

- Added `test_edit_form_omits_hidden_field` and `test_bulk_edit_form_omits_hidden_field` in `test_views.py`, asserting a hidden field's name is absent from `form.base_fields`.
- Verified manually against netbox-community/netbox `main` (Django 6.0.7) and a `feature`-branch commit with Django 6.1 pinned.

### DIFF
--- a/netbox_custom_objects/views.py
+++ b/netbox_custom_objects/views.py
@@ -797,8 +797,14 @@ class CustomObjectEditView(generic.ObjectEditView):
 
         # Process custom object type fields (with grouping)
         for field in cot_fields:
-            # Hidden fields are omitted entirely, not just disabled.
+            # Hidden fields are omitted entirely, not just disabled -- but a
+            # non-polymorphic MultiObject field is a real M2M model attribute
+            # whose configured default must still be applied on create (#42).
+            # Track it in custom_object_type_fields (bookkeeping only, not
+            # rendered anywhere) so custom_init/custom_save still process it.
             if field.ui_editable == CustomFieldUIEditableChoices.HIDDEN:
+                if field.type == CustomFieldTypeChoices.TYPE_MULTIOBJECT and not field.is_polymorphic:
+                    attrs["custom_object_type_fields"][field.name] = field
                 continue
 
             field_type = field_types.FIELD_TYPE_CLASS[field.type]()
@@ -888,6 +894,10 @@ class CustomObjectEditView(generic.ObjectEditView):
             self.custom_object_type_poly_obj_ct_names = attrs["custom_object_type_poly_obj_ct_names"]
             self.custom_object_type_poly_obj_pairs = attrs["custom_object_type_poly_obj_pairs"]
             self.custom_object_type_coordinates_fields = attrs["custom_object_type_coordinates_fields"]
+            # A hidden MultiObject field has no rendered form field, so its resolved
+            # default can't reach cleaned_data via kwargs['initial'] below -- custom_save
+            # applies these directly on create instead. See the note in the loop above.
+            self._hidden_multiobject_defaults = {}
 
             instance = kwargs.get('instance', None)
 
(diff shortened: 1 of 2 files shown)


bctiemann:
@claude /review and keep response limited to actionable issues.

bctiemann:
@claude just curious, do you have access to the session context from the review previously requested?

bctiemann:
Added test coverage.

netbox_custom_objects/views.py
```diff
@@ -768,9 +796,11 @@ def get_form(self, model):
         }
 
         # Process custom object type fields (with grouping)
-        for field in self.object.custom_object_type.fields.prefetch_related(
-            'related_object_types'
-        ).order_by("group_name", "weight", "name"):
+        for field in cot_fields:
+            # Hidden fields are omitted entirely, not just disabled.
+            if field.ui_editable == CustomFieldUIEditableChoices.HIDDEN:
+                continue
```
pheus: Could we preserve the configured default for hidden `multiobject` fields here?

Skipping the field at this point also removes it from `custom_object_type_fields`, which is later used to initialize and save non-polymorphic multi-object relations. As these fields do not have a model-level default, creating an object with a hidden field would silently omit its configured default, reintroducing the behavior fixed in #42.

I think we need to keep the field out of the form while still applying its default during creation. It would also be helpful to cover this with a save-level regression test. Existing relations should remain unchanged when editing an object.

pheus (changes requested):
Thanks!

I found one blocking issue with hidden non-polymorphic multi-object fields: skipping them here also bypasses the view-level initialization and persistence path, so their configured defaults are no longer applied when creating an object. This appears to reintroduce the behavior fixed in #42.

Once that path is preserved and covered by a save-level regression test, the rest looks good to me.

netbox_custom_objects/views.py
bctiemann: Thanks for catching this; fixed and covered with a new test.

pheus (approved):
Thanks for addressing this.

PR #664, opened 2026-08-19
Closes: #637 - Fix model-registry errors when get_model() runs outside ready()'s two-pass loop

### LINKED ISSUE #637: Model relation errors when upgrading NetBox
### Plugin Version

0.6.0

### NetBox Version

4.6.6

### Python Version

3.13

### Steps to Reproduce

Starting with NetBox 4.6.2 and netboxlabs-netbox-custom-objects 0.6.0.

Having two custom object types:
- `dns_zone` (table1model as shown in some places)
- `forwarder_profile` (table2model as shown in some places)

The `dns_zone` model has a `forwarder_profile` field that refers to a `forwarder_profile` object:
- Field type: Object
- Required: no (unchecked)
- Polymorphic: no (unchecked)
- Related object type: Custom Objects > Forwarder Profile

Trying to upgrade NetBox:

1. git fetch --tags
2. git checkout v4.6.6
3. ./upgrade.sh

### Expected Behavior

NetBox is upgraded and migrations complete successfully.

### Observed Behavior

...
Installing local dependencies (pip install -r local_requirements.txt)...
...
Installing collected packages: ..., netboxlabs-netbox-custom-objects
Successfully installed ... netboxlabs-netbox-custom-objects-0.6.0
Applying database migrations (python3 netbox/manage.py migrate)...
SystemCheckError: System check identified some issues:

ERRORS:
netbox_custom_objects.Table1Model.forwarder_profile: (fields.E300) Field defines a relation with model 'netbox_custom_objects.Table2Model', which is either not installed, or is abstract.
netbox_custom_objects.Table1Model.forwarder_profile: (fields.E307) The field netbox_custom_objects.Table1Model.forwarder_profile was declared with a lazy reference to 'netbox_custom_objects.table2model', but app 'netbox_custom_objects' doesn't provide model 'table2model'.

### Background information

The COTs were originally created with plugin version 0.5.2, if that matters. migrate + collectstatic were run successfully (in NetBox 4.6.2) when upgrading the plugin from 0.5.2 to 0.6.0 earlier.

bctiemann:
## Summary

A COT with a cross-COT Object-type field, when `.get_model()` is called directly on the model instance (the pattern shown in `docs/index.md` -- a module-level `CustomObjectType.objects.get(...).get_model()`) while it's unsafe to touch the DB (e.g. during `manage.py migrate`), left the field's `LazyForeignKey` target unresolved. Django's system checks then raised `fields.E300`/`fields.E307` for that field, aborting the upgrade.

Root cause: every other dynamic-model entry point (`ready()`, `PluginConfig.get_model()`/`get_models()`) already guards against generating COT models before migrations have completed. `CustomObjectType.get_model()` -- the one method third-party plugins call directly -- didn't check this at all, so it could register a COT model outside of `ready()`'s two-pass cross-COT FK resolution, leaving a dangling reference that nothing else ever re-resolves during that `migrate` process.

## Changes

- `CustomObjectType.get_model()` now checks a new `_dynamic_model_creation_unsafe()` predicate and forces `skip_object_fields=True` when it's set, omitting the cross-COT field entirely rather than leaving a dangling lazy reference. The resulting degraded model is never cached (existing behavior for `skip_object_fields=True`), so a real request after startup regenerates it in full via `ready()`'s normal two-pass path in the actual (separate) worker process.
- `_dynamic_model_creation_unsafe()` is the actual DB-readiness check (mid-migration, `migrate`/`makemigrations`/`collectstatic`, or this app's own migrations incomplete) -- deliberately excluding `"test"`, since `get_model()` is the method tests use directly and the test database is always fully migrated by the time a test body runs.
- `should_skip_dynamic_model_creation()` -- used by `ready()`, `PluginConfig.get_model()`/`get_models()`, and `navigation` -- keeps its existing public signature and behavior, now expressed as `_dynamic_model_creation_unsafe()` OR `"test"` in `sys.argv`.

## Testing

- New `CrossCOTGetModelOutsideReadyTestCase` in `test_models.py`:
  - `test_get_model_omits_cross_cot_field_when_unsafe` -- the field is omitted, and the degraded model isn't cached.
  - `test_get_model_system_checks_pass_after_migrate_time_call` -- reproduces the exact `fields.E300`/`E307` symptom and confirms it's gone.
  - `test_get_model_regenerates_in_full_on_next_normal_call` -- once the target COT is registered again (mirroring a real restart's `ready()` pass), a subsequent call fully resolves the field.
  - `test_get_model_unaffected_during_ordinary_test_run` -- ordinary test-suite `get_model()` calls stay fully hydrated.
- New `test_dynamic_model_creation_unsafe_excludes_test` in `PluginConfigGetModelTestCase`.
- Verified locally against netbox-community/netbox `main` (Django 6.0.7) and `feature` (Django 6.1), plus the broader `test_models`/`test_api`/`test_filtersets`/`test_polymorphic_fields`/`test_field_types`/`test_views` suites.

### DIFF
--- a/netbox_custom_objects/__init__.py
+++ b/netbox_custom_objects/__init__.py
@@ -282,17 +282,27 @@ class CustomObjectsPluginConfig(PluginConfig):
     graphql_schema = "graphql.schema.schema"
 
     @staticmethod
-    def should_skip_dynamic_model_creation():
+    def should_skip_dynamic_model_creation(include_test_skip=True):
         """
         Determine if dynamic model creation should be skipped.
 
         Returns True if dynamic models should not be created/loaded due to:
         - Currently running migrations
-        - Running tests
+        - Running tests (if include_test_skip)
         - All migrations not yet applied
         - Running collectstatic
 
         Returns False if it's safe to proceed with dynamic model creation.
+
+        :param include_test_skip: Also skip when "test" is in sys.argv. The
+            plugin's own startup-time entry points (ready(), PluginConfig.get_model/
+            get_models) never need to run during tests -- test code generates COT
+            models directly via CustomObjectType.get_model() instead. That method
+            also consults this check (to avoid #637: a dangling cross-COT FK
+            reference when get_model() runs outside ready()'s two-pass resolution,
+            e.g. a third-party plugin importing it at migrate time), but must pass
+            False here since it's the one entry point tests DO rely on for a fully
+            hydrated model.
         """
         global _migrations_checked, _checking_migrations
 
@@ -300,17 +310,17 @@ class CustomObjectsPluginConfig(PluginConfig):
         if _is_migrating.get():
             return True
 
-        skip_commands = (
+        skip_commands = [
             # Running migrations should skip.
             "makemigrations",
             "migrate",
 
             # The database isn't accessible during collect static so should skip.
             "collectstatic",
-
+        ]
+        if include_test_skip:
             # Skip during tests.
-            "test",
-        )
+            skip_commands.append("test")
 
         if any(cmd in sys.argv for cmd in skip_commands):
             return True
--- a/netbox_custom_objects/__init__.py
+++ b/netbox_custom_objects/__init__.py
@@ -284,18 +284,11 @@ class CustomObjectsPluginConfig(PluginConfig):
     @staticmethod
     def _dynamic_model_creation_unsafe():
         """
-        True if the DB can't safely be queried yet to generate a dynamic COT model:
-        mid-migration, running makemigrations/migrate/collectstatic (schema may be
-        incomplete or absent), or this app's own migrations aren't fully applied.
-
-        Deliberately excludes "test" -- unlike should_skip_dynamic_model_creation()
-        below, this is the check CustomObjectType.get_model() itself uses (#637: a
-        bare get_model() call outside ready()'s two-pass cross-COT FK resolution,
-        e.g. a third-party plugin importing it at migrate time, leaves a dangling
-        LazyForeignKey that Django's system checks flag). By the time get_model()
-        runs on an already-fetched COT instance inside a test method, the test
-        database is fully migrated and safe to query -- and test code relies on
-        get_model() returning a fully-hydrated model, not a degraded one.
+        True if the DB can't safely be queried to generate a dynamic COT model:
+        mid-migration, running makemigrations/migrate/collectstatic, or this
+        app's own migrations aren't fully applied. Unlike
+        should_skip_dynamic_model_creation() below, deliberately excludes "test" --
+        this is the narrower check CustomObjectType.get_model() itself uses (#637).
         """
         global _migrations_checked, _checking_migrations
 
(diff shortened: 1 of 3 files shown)


bctiemann:

bctiemann:
Addressed items 1 and 2.

arthanson (changes requested):
From Claude:

The PR's guard drops all Object/Multi-object fields when it fires, but only COT-targeting non-polymorphic ones can dangle.

Consequence: migration 0003_ensure_fk_constraints calls get_model() inside a migrate process, so it now finds no object fields and creates no FK constraints - silently (one logger.error). Core-model targets (dcim.Site etc.) worked there before; nothing downstream backfills a missing constraint, so affected COTs lose DB-level on_delete_behavior enforcement. Reachable on installs still at plugin ≤0.4.0 that have COTs.

Fix: gate on "target is another COT" instead of "is an object field"

Secondary: the degradation isn't logged, and _migrations_checked caches True for the process lifetime, so a worker started mid-upgrade serves field-less models until restart.

Here is what is came up for as a fix:

One file for the behavior: netbox_custom_objects/models.py

1. A new predicate on CustomObjectTypeField (next to related_object_type_label, which already open-codes this pattern at models.py:2605, as does :3049):
```
      @property
      def targets_custom_object_type(self):
          """
          True when this field points at another COT's dynamic model, i.e. its FK is a
          LazyForeignKey that only resolves once the target is in the app registry.
          Polymorphic fields use a GFK and need no resolution, so they are excluded.
          """
          if self.is_polymorphic or not self.related_object_type_id:
              return False
          try:
              object_type = ContentType.objects.get_for_id(self.related_object_type_id)
          except ContentType.DoesNotExist:
              return False
          return (
              object_type.app_label == APP_LABEL
              and extract_cot_id_from_model_name(object_type.model) is not None
          )
```
ContentType and extract_cot_id_from_model_name are already imported there. get_for_id() is the process-cached lookup, so this doesn't add a query per field the way _get_related_content_type() (field_types.py:293) would. The extract_cot_id_from_model_name half matters: an Object field targeting netbox_custom_objects.customobjecttype shares the app label but isn't a dynamic model, and field_types.py:780 treats it differently.

2. _fetch_and_generate_field_attrs() gains a second, narrower mode - the existing skip_object_fields=True semantics are untouched, because field_types.py:1446 and views.py:922 rely on them to build stub models that break FK recursion:
```
           for field in fields:
  -            if skip_object_fields:
  -                if field.type in [CustomFieldTypeChoices.TYPE_OBJECT, CustomFieldTypeChoices.TYPE_MULTIOBJECT]:
  +            if field.type in [CustomFieldTypeChoices.TYPE_OBJECT, CustomFieldTypeChoices.TYPE_MULTIOBJECT]:
  +                if skip_object_fields:
  +                    continue
  +                # Only a COT target produces a LazyForeignKey that needs ready()'s
  +                # second pass; core-model targets and polymorphic GFKs resolve on
  +                # their own, so they stay even in the unsafe window (#637).
  +                if skip_cot_object_fields and field.targets_custom_object_type:
                       continue
```
3. get_model() sets the narrow flag instead of overwriting the caller's - three small hunks: the guard becomes skip_cot_object_fields = apps.get_app_config(APP_LABEL)._dynamic_model_creation_unsafe(), that flag is passed through to _fetch_and_generate_field_attrs(), and the cache condition becomes if not (skip_object_fields or skip_cot_object_fields) so a narrowed model still never gets cached.

bctiemann:
Thanks @arthanson 

1. Added `CustomObjectTypeField.targets_custom_object_type`
2. Narrowed the migrate-time skip logic to only drop COT-targeting Object/Multi-object fields, leaving core-model references and polymorphic GFKs intact even in the unsafe window
3. `get_model()`'s cache guard now correctly accounts for both skip flags

arthanson (approved):

PR #666, opened 2026-08-19
Merge main to feature

bctiemann:
Merge 'main' into 'feature'.


bctiemann:
Superseded by #667 — this PR's head branch is `main` itself, which isn't safe to push conflict-resolution commits onto. #667 has the same merge (feature ← main) resolved on a proper throwaway branch, plus fixups for issues the merge introduced, with the full suite verified passing.

Issue #665, opened 2026-08-19
Change exportable schema interchange format from JSON to YAML


bctiemann:
### Plugin Version

v0.6.0

### Proposed functionality

Use YAML as the standard interchange format for exporting and importing custom object type schema definitions.

### Use case

YAML is a more readable and ergonomic format than JSON and would be a better choice for encapsulating the object definitions between systems.

### External dependencies

PyYAML

PR #668, opened 2026-08-19
Closes: #665 - Support YAML for the portable schema export/preview/apply feature

bctiemann:
### Closes: #665

Adds YAML support for the portable schema export/preview/apply endpoints (`SchemaPreviewView`, `SchemaApplyView`), per #665. JSON remains fully supported and is still the default when no `Accept`/`Content-Type` header is given, so this is backward compatible.

- `api/parsers.py`: `YAMLParser(BaseParser)`, `media_type = 'application/yaml'`, parses with `yaml.safe_load`.
- `api/renderers.py`: `YAMLRenderer(BaseRenderer)`, `media_type = 'application/yaml'`, `format = 'yaml'`, renders with `yaml.safe_dump(data, sort_keys=False, default_flow_style=False)`.
- `api/views.py`: `_SCHEMA_PARSER_CLASSES = [YAMLParser, JSONParser]` and `_SCHEMA_RENDERER_CLASSES = [JSONRenderer, YAMLRenderer, FormlessBrowsableAPIRenderer]` (JSON first so a request with no `Accept` header still gets JSON).
- `pyproject.toml`: adds `PyYAML` as a dependency.
- `docs/portable-schema.md`: documents both formats.
- Tests: `tests/schema/test_yaml_codec.py` (parser/renderer round-trip), plus new cases in `tests/schema/test_schema_api.py` covering content-negotiated preview/apply requests.

Rebased onto the now-merged `feature` (post #666/#667), which brought in the Django 6.1 `from_db()` fix. Full suite verified in a clean venv (no stray `netbox_branching` package on the path): 1218 tests, 0 failures/errors, 8 skipped.

### DIFF
--- a/docs/portable-schema.md
+++ b/docs/portable-schema.md
@@ -1,36 +1,31 @@
 # Portable Schema
 
 The portable schema feature allows Custom Object Type (COT) definitions to be exported as
-structured JSON documents, versioned in source control, and applied to other NetBox instances.
+structured YAML documents, versioned in source control, and applied to other NetBox instances.
 This makes COT schemas shareable, auditable, and deployable across environments in a consistent
-and repeatable way.
+and repeatable way. YAML is the recommended format; JSON is also accepted (see
+[Schema Document Format](#schema-document-format)).
 
 ## Concepts
 
 ### Schema Documents
 
-A schema document is a JSON object that fully describes one or more Custom Object Types —
-their names, metadata, and all field definitions. The document is self-contained: a reader
-does not need access to the originating NetBox instance to understand or validate it.
+A schema document is a mapping that fully describes one or more Custom Object Types — their
+names, metadata, and all field definitions. The document is self-contained: a reader does not
+need access to the originating NetBox instance to understand or validate it.
 
-```json
-{
-  "schema_version": "1",
-  "types": [
-    {
-      "name": "circuit",
-      "slug": "circuit",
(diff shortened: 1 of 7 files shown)


bctiemann:

bctiemann:
Addressed all three:

1. `YAMLRenderer.render()` now returns bytes per DRF convention.
2. The null/empty-body `ParseError` message no longer misleadingly claims the body is "empty" when it's actually a valid null document.
3. `SchemaPreviewView` now has the same non-dict-body guard as `SchemaApplyView`.

jnovinger (approved):
Approved as is, but with one potential suggestion.

`_StrictBoolLoader` inherits `SafeLoader`'s alias resolution, so an anchor/alias bomb can expand in memory at parse time before `_validate_schema_doc` runs. You're already subclassing the loader here, so it's cheap to close: reject `AliasEvent` in `compose_node`. `ComposerError` is a `YAMLError`, so your existing `except yaml.YAMLError` already turns it into a clean 400.

```python
def compose_node(self, parent, index):
    if self.check_event(yaml.events.AliasEvent):
        event = self.peek_event()
        raise yaml.composer.ComposerError(
            None, None, "YAML anchors/aliases are not permitted in schema documents", event.start_mark
        )
    return super().compose_node(parent, index)
```

Not blocking, but suggesting it since schema docs don't seem to have a legitimate use for anchors.

PR #676, opened 2026-08-20
Fixes #629: Delete stale model registration before regenerating, not after

### LINKED ISSUE #629: Phantom Tagged Objects
### Plugin Version

0.5.2

### NetBox Version

4.6.3

### Python Version

3.14.4

### Steps to Reproduce

Create a custom object type. Create several of those objects and apply some tags to them. I'm actually unsure what actual steps it takes to reproduce this fully as I'm actively using this in production, so custom objects are being added, deleted, modified every day. In our case we have a custom obect type called ssl-profiles, and a tag called unbound with ID 359. This tag shows 7 tagged items but they are blank. Filtering on this tag shows a 7 on the tag name but no results in the resultant query 

### Expected Behavior

For there to be a (0) after this tag. 

### Observed Behavior

tag shows 7 items, where none exist. 

Image

Image

bctiemann:
## Summary

Reproduces and fixes #629 (Phantom Tagged Objects): deleting a tagged custom object left its `TaggedItem` row behind. The tag's detail page reported a nonzero item count, but filtering by that tag returned zero results, since the referenced object no longer existed.

## Root cause

`CustomObjectType.get_model()` deleted the old `apps.all_models` registry entry for a COT's dynamic model *after* `generate_model()` had already built the replacement class, not before.

`TagsMixin`'s `tags` field resolves its `through` model lazily: taggit's `contribute_to_class()` runs mid-construction, while Django's `ModelBase.__new__()` is still adding fields to the new class — well before the new class itself gets registered. It calls `lazy_related_operation()`, which treats the *owning class* as a dependency, resolved by looking it up in `apps.all_models` under the model's name (not the literal class object passed in). Since the *old* class was still registered under that name at that exact moment, the lookup resolved immediately against the stale, about-to-be-replaced class instead of deferring — so the new class's own `tags` field's `post_through_setup()` never ran against itself, and it never got a `tagged_items` `GenericRelation`. Without that `GenericRelation`, Django's deletion collector has no way to cascade-delete a custom object's `TaggedItem` rows on delete — regardless of whether the object is deleted via a single instance's `.delete()`, a bulk `queryset.delete()`, or a freshly-regenerated class.

## Fix

Delete the stale registry entry *before* calling `generate_model()` instead of after, so `generate_model()`'s own `type()` call is what correctly registers (and satisfies the lazy dependency for) the new class. The now-redundant explicit `apps.register_model()` call afterward is guarded to skip re-registering when `generate_model()` already did so, avoiding a spurious "already registered" `RuntimeWarning`.

## Testing

New `PhantomTaggedObjectsTestCase` in `test_models.py`:
- `test_generated_model_has_tagged_items_generic_relation` — the root cause: the generated model must actually have a `tagged_items` `GenericRelation`.
- `test_direct_delete_removes_tagged_item` / `test_queryset_bulk_delete_removes_tagged_item` / `test_delete_after_model_regeneration_removes_tagged_item` — all three deletion paths must clean up the `TaggedItem` row.
- `test_filtering_by_tag_matches_actual_object_count` — the user-visible symptom: filtering by tag after deleting a tagged object must return the correct count.

All 5 fail without the fix (reproducing the exact reported symptom) and pass with it.

Full suite verified in a clean venv (fresh install, no stray environment state): 1154 tests, 0 failures/errors, 8 skipped.

### DIFF
--- a/netbox_custom_objects/models.py
+++ b/netbox_custom_objects/models.py
@@ -1738,19 +1738,10 @@ class CustomObjectType(NetBoxModel):
 
         # Delete any stale registration for this model name *before* generating
         # the replacement class, not after (issue #629). TagsMixin's 'tags' field
-        # resolves its 'through' model lazily: contribute_to_class() (which runs
-        # mid-construction, while Django's ModelBase.__new__() is still adding
-        # fields to the new class) calls lazy_related_operation(), which treats
-        # the OWNING class itself as a dependency, satisfied by looking it up in
-        # apps.all_models under this exact model name. If the *old* class was
-        # still registered under that name at that point, the lookup resolves
-        # immediately against the stale class instead of deferring — so the new
-        # field's post_through_setup() runs with cls=<old, about-to-be-replaced
-        # class>, and the new class's own 'tags' field never gets its
-        # tagged_items GenericRelation set up. Without that GenericRelation,
-        # Django's deletion collector has no way to cascade-delete a custom
-        # object's TaggedItem rows, leaving orphaned ("phantom") tagged items
-        # behind whenever a custom object is deleted.
+        # resolves its 'through' model lazily against whatever's registered under
+        # this model name at contribute_to_class() time; leaving the old class
+        # registered there let the new field's setup bind to the old class
+        # instead of itself, silently breaking tag cascade-delete.
         if branch_id is None:
             model_key = model_name.lower()
             with _suppress_clear_cache():
@@ -1796,11 +1787,8 @@ class CustomObjectType(NetBoxModel):
             # would return a class with the wrong column set across contexts.
             model_key = model_name.lower()
             if branch_id is None:
-                # generate_model()'s own type() call already registered this class
-                # under model_key (the pre-deletion above ensured nothing else held
-                # that slot). This call is a safety net for the (unexpected) case
-                # where it didn't -- skip it when redundant to avoid a spurious
-                # "already registered" RuntimeWarning.
+                # generate_model() already registered this class; skip the
+                # redundant call to avoid a spurious "already registered" warning.
                 if apps.all_models[APP_LABEL].get(model_key) is not model:
                     apps.register_model(APP_LABEL, model)
             else:
--- a/netbox_custom_objects/models.py
+++ b/netbox_custom_objects/models.py
@@ -1736,6 +1736,27 @@ class CustomObjectType(NetBoxModel):
         # Wrap the existing post_through_setup method to handle ValueError exceptions
         from taggit.managers import TaggableManager as TM
 
+        # Delete any stale registration for this model name *before* generating
+        # the replacement class, not after (issue #629). TagsMixin's 'tags' field
+        # resolves its 'through' model lazily: contribute_to_class() (which runs
+        # mid-construction, while Django's ModelBase.__new__() is still adding
+        # fields to the new class) calls lazy_related_operation(), which treats
+        # the OWNING class itself as a dependency, satisfied by looking it up in
+        # apps.all_models under this exact model name. If the *old* class was
+        # still registered under that name at that point, the lookup resolves
+        # immediately against the stale class instead of deferring — so the new
+        # field's post_through_setup() runs with cls=<old, about-to-be-replaced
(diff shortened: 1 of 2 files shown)


bctiemann:

bctiemann:
1. Added a note to the existing concurrency comment at the top of `get_model()` explaining that a concurrent, unrelated lookup can now transiently see the model name as unregistered during regeneration
2. Removed redundant call
3. Fixed

netbox_custom_objects/models.py
```diff
@@ -1736,6 +1739,16 @@ def get_model(
...
 
+        # Delete any stale registration for this model name *before* generating
+        # the replacement class, not after (issue #629). TagsMixin's 'tags' field
+        # resolves its 'through' model lazily against whatever's registered under
+        # this model name at contribute_to_class() time; leaving the old class
+        # registered there let the new field's setup bind to the old class
+        # instead of itself, silently breaking tag cascade-delete.
+        model_key = model_name.lower()
+        if branch_id is None and model_key in apps.all_models[APP_LABEL]:
+            del apps.all_models[APP_LABEL][model_key]
```
pheus: Would you be open to fixing the binding directly instead of temporarily removing the registry entry? `TaggableManager` retains the class being built as `manager.model`, so the wrapper could use that rather than the class resolved through the app registry:

```python
def wrapped_post_through_setup(manager, _resolved_model):
    try:
        return original_post_through_setup(manager, manager.model)
    except ValueError:
        pass
```

This would keep the model registered during generation and ensure each generated class receives its own `tagged_items` relation.

netbox_custom_objects/models.py
```diff
@@ -1736,6 +1739,16 @@ def get_model(
...
         from taggit.managers import TaggableManager as TM
 
+        # Delete any stale registration for this model name *before* generating
+        # the replacement class, not after (issue #629). TagsMixin's 'tags' field
+        # resolves its 'through' model lazily against whatever's registered under
+        # this model name at contribute_to_class() time; leaving the old class
+        # registered there let the new field's setup bind to the old class
+        # instead of itself, silently breaking tag cascade-delete.
+        model_key = model_name.lower()
+        if branch_id is None and model_key in apps.all_models[APP_LABEL]:
```
pheus: Could we cover the branch path as well? This is limited to `branch_id is None`, so the registered main class can still be resolved while constructing a branch model.

Using `manager.model` in the wrapper would address both contexts without changing the canonical registry entry. A branch regression test that tags and deletes an object, then verifies that its `TaggedItem` is removed, would help confirm the behavior.

netbox_custom_objects/models.py
```diff
@@ -1773,11 +1786,11 @@ def wrapped_post_through_setup(self, cls):
...
-            model_key = model_name.lower()
             if branch_id is None:
-                if model_key in apps.all_models[APP_LABEL]:
-                    del apps.all_models[APP_LABEL][model_key]
-                apps.register_model(APP_LABEL, model)
+                # generate_model() already registered this class; skip the
+                # redundant call to avoid a spurious "already registered" warning.
+                if apps.all_models[APP_LABEL].get(model_key) is not model:
+                    apps.register_model(APP_LABEL, model)
             else:
```
pheus, line 1792: Could we cover the branch path as well? This is limited to `branch_id is None`, so the registered main class can still be resolved while constructing a branch model.

Using `manager.model` in the wrapper would address both contexts without changing the canonical registry entry. A branch regression test that tags and deletes an object, then verifies that its `TaggedItem` is removed, would help confirm the behavior.

pheus (changes requested):
Thanks!
I’ve left two inline comments around binding the setup to the generated model and covering the branch path. I’m requesting changes for those points, but the overall direction looks good.

netbox_custom_objects/models.py
bctiemann: Good point; reverted this to use `manager.model` instead of relying on `cls` and now no longer touches the registry at all.

netbox_custom_objects/models.py
bctiemann: The above comment covers this one too — no longer using the registry at all for this, so the branch path is covered.

pheus (approved):

PR #680, opened 2026-08-25
Fixes #653: Fix Branch.sync() cross-connection deadlock on ChangeDiff rewrites

### LINKED ISSUE #653: Branch.sync() deadlocks when a field rename's ChangeDiff rewrite blocks record_change_diff's own ChangeDiff write
## Plugin Version

0.6.0 (main, 16 commits past the 0.6.0 tag)

## NetBox Version

4.6.8

## Python Version

3.12.1

## Steps to Reproduce

1. With `netbox-branching` installed (tested against 1.1.2) and the plugin's `configuration_co_branching_testing` test config, run:
   ```
   cd netbox/
   NETBOX_CONFIGURATION=netbox.configuration_co_branching_testing \
     python manage.py test netbox_custom_objects.tests.test_branching.SequentialRenameTestCase.test_sequential_renames_both_sides_sync
   ```
2. This test provisions a branch, renames a Custom Object Type field on the branch (alpha→beta→gamma), renames the *same* field on main (alpha→delta), then calls `branch.sync(user=self.user, commit=True)` to replay main's change onto the branch.

Reproduced 4/4 attempts on a fresh database. The hang occurs partway through `sync()`, not at test start.

## Expected Behavior

`sync()` completes (or raises a normal exception) within a few seconds.

## Observed Behavior

The process hangs indefinitely (confirmed hung for 22+ minutes before being killed manually; there is no timeout that resolves it). It is **not** a livelock or infinite loop — it is a genuine cross-connection lock wait that never gets released, confirmed via `pg_stat_activity`, `pg_blocking_pids()`, `pg_locks`, and a live Python thread dump (`faulthandler`, since `py-spy` doesn't attach cleanly to this Python build).

### Root cause

`Branch.sync()` (`netbox_branching/models/branches.py`) wraps its entire body in **two simultaneously-open transactions**:

```python
with (
    activate_branch(self),
    transaction.atomic(using=DEFAULT_DB_ALIAS),
    transaction.atomic(using=self.connection_name),
):
    ...
```

Both stay open for the whole call — the branch connection's transaction is not released until `sync()` itself returns.

While replaying main's changes onto the branch, one of them is a `CustomObjectTypeField` rename. Applying it calls `CustomObjectTypeField.save()`, which calls `_rename_objectchange_field_key()` (`netbox_custom_objects/models.py:2159`). That function does a **raw SQL `UPDATE netbox_branching_changediff ...`** against `_get_schema_connection()` — i.e. the *branch's own connection* — to rewrite renamed JSON keys inside `ChangeDiff.original`/`.modified`/`.current`:

```python
conn = _get_schema_connection()   # branch connection when a branch is active
...
with transaction.atomic(using=conn.alias):
    with conn.cursor() as cursor:
        for json_col in ('original', 'modified', 'current'):
            cursor.execute(cd_sql.format(col=json_col), [...])
```

`netbox_branching_changediff` only exists in the main/public schema (it is **not** a per-branch-replicated table — `ChangeDiff` is a plain `models.Model`, not a `ChangeLoggingMixin` subclass, so `BranchAwareRouter.supports_branching()` returns `False` for it and every ORM-driven access always routes to `DEFAULT_DB_ALIAS`). It's still reachable and *writable* from the branch connection purely because the branch connection's `search_path` falls back to the main schema. This raw UPDATE therefore takes a real row lock on the physical `netbox_branching_changediff` table, held until the branch connection's transaction commits — which, per above, won't happen until `sync()` returns.

Later in the *same* `sync()` call, `_flush_sync_buffer()` (`branches.py:676`) creates synthetic `ObjectChange` rows on the branch connection. Each creation synchronously fires the `record_change_diff` signal (`signal_receivers.py:168`), which does `ChangeDiff.objects.filter(...).first()` / `diff.save()` — and *this* write, being plain ORM access with no explicit `.using()`, always resolves to `DEFAULT_DB_ALIAS` (again, because `ChangeDiff` isn't branch-aware).

If that later `ChangeDiff` row happens to be one already locked by the earlier raw-SQL update (both target rows for the same `object_type_id`, matching the renamed field's Custom Object Type), the default-connection UPDATE blocks waiting for the branch connection's transaction to finish. That transaction can only finish once `sync()`'s outer `with` block exits — which requires the very statement that's now blocked to complete first.

This is a **single-thread, dual-connection lock-ordering deadlock**, not a classic multi-session deadlock — confirmed by a live thread dump showing exactly one Python thread. PostgreSQL's own deadlock detector never fires: only the default-connection session is actually waiting on a lock; the branch connection is simply idle (`idle in transaction`, `wait_event=ClientRead`), passively holding its lock while its owning thread is stuck elsewhere. From Postgres's perspective there is no lock cycle to detect, so nothing ever times it out.

### Evidence

`pg_stat_activity` mid-hang:
```
  pid  |        state         | wait_event_type |  wait_event   |  query
-------+----------------------+------------------+---------------+---------------------------------------------
 65164 | active               | Lock             | transactionid | UPDATE "netbox_branching_changediff" SET "branch_id" = 1, ...
 65417 | idle in transaction  | Client           | ClientRead    | SELECT "netbox_custom_objects_customobjecttype"...
```

`pg_blocking_pids(65164)` → `{65417}`. `pg_locks` shows `65417` holding a `RowExclusiveLock` on `public.netbox_branching_changediff` (confirmed via `pg_class`/`pg_namespace` lookup against the actual test database — `65417`'s other locks are all on `branch_<schema>`-local tables, confirming it is the branch connection).

Live thread dump (`faulthandler.dump_traceback(all_threads=True)`, single thread present):
```
File ".../psycopg/connection.py", line 484 in wait
File ".../psycopg/cursor.py", line 113 in execute
...
File ".../django/db/models/base.py", line 874 in save
File "netbox_branching/models/changes.py", line 257 in save          # ChangeDiff.save()
File "netbox_branching/signal_receivers.py", line 168 in record_change_diff
File ".../django/dispatch/dispatcher.py", line 209 in send
File ".../django/db/models/base.py", line 874 in save                # ObjectChange.save()
File "netbox/core/models/change_logging.py", line 140 in save
File ".../django/db/models/query.py", line 669 in create
File "netbox_branching/models/branches.py", line 676 in _flush_sync_buffer
File "netbox_branching/models/branches.py", line 837 in sync
File "netbox_custom_objects/tests/test_branching.py", line 2251 in test_sequential_renames_both_sides_sync
```

### Suggested fix

`_rename_objectchange_field_key()`'s `ObjectChange` block correctly uses the branch connection, because `core.ObjectChange` *is* per-branch-replicated (documented in the existing comment there). But its `ChangeDiff` block reuses the same `conn`/`conn.alias` without the same justification — `ChangeDiff` has no branch-schema copy and is always accessed via `DEFAULT_DB_ALIAS` everywhere else in `netbox-branching`. Writing to it via the branch connection here is what introduces the second, conflicting connection into the picture.

Using the default connection for just the `ChangeDiff` block instead — matching how `record_change_diff` itself accesses `ChangeDiff` — removes the cross-connection reference entirely (since `sync()` already holds `transaction.atomic(using=DEFAULT_DB_ALIAS)` open for its whole duration, this becomes a savepoint within that existing transaction, not a new one):

```python
from django.db import DEFAULT_DB_ALIAS, connections
...
try:
    with transaction.atomic(using=DEFAULT_DB_ALIAS):
        with connections[DEFAULT_DB_ALIAS].cursor() as cursor:
            for json_col in ('original', 'modified', 'current'):
                cursor.execute(cd_sql.format(col=json_col), [...])
except ProgrammingError:
    ...
```

I have not yet implemented or tested this fix — flagging the root cause here first per repo convention before opening a PR.

bctiemann:
## Summary

Fixes a same-process, cross-connection deadlock in `Branch.sync()` reported in #653. `_rename_objectchange_field_key()`'s `ChangeDiff` block wrote via the branch connection — the same connection already correctly used for `ObjectChange`, which *is* per-branch-replicated. `ChangeDiff` has no per-branch schema copy: it's a plain `models.Model`, not a `ChangeLoggingMixin` subclass, so `BranchAwareRouter.supports_branching()` returns `False` for it and every other access to it (including netbox-branching's own `record_change_diff` signal receiver) always resolves to `DEFAULT_DB_ALIAS`.

## Root cause

`Branch.sync()` holds two simultaneously-open transactions for its whole duration — `DEFAULT_DB_ALIAS` and the branch connection. Writing `ChangeDiff` via the branch connection introduced a second, unrelated connection into that pair:

1. A field rename replayed onto the branch calls `_rename_objectchange_field_key()`, which does a raw `UPDATE netbox_branching_changediff ...` via the **branch connection**, taking a row lock held until `sync()` returns.
2. Later in the *same* `sync()` call, `_flush_sync_buffer()` creates `ObjectChange` rows, each synchronously firing `record_change_diff`, which writes the *same* `ChangeDiff` row via the **default connection** (plain ORM access, no `.using()`).
3. If that row is already locked by step 1, step 2 blocks waiting for the branch connection's transaction to commit — which can't happen until `sync()` returns, which needs step 2 to finish first.

Postgres's own deadlock detector never fires here: only one session is actually waiting on a lock; the other is idle (`idle in transaction`), passively holding its lock while its owning thread (the same Python thread, in this single-threaded scenario) is blocked elsewhere. From Postgres's perspective there's no lock cycle to detect.

## Fix

Route the `ChangeDiff` write through `DEFAULT_DB_ALIAS`, matching every other `ChangeDiff` access in the codebase. Since `sync()` already holds `transaction.atomic(using=DEFAULT_DB_ALIAS)` open for its whole duration, this becomes a plain savepoint within that existing transaction rather than a second, independent one — removing the cross-connection reference entirely.

## Testing

Verified against a disposable NetBox `main` checkout (Django 6.0.7, matching `netbox_branching`'s `max_version` pin) with `netbox-branching` 1.1.2:

- The exact reproduction from the issue (`test_sequential_renames_both_sides_sync`), which previously hung indefinitely (confirmed hung 22+ minutes via `pg_stat_activity`/`pg_locks`/a live thread dump before being killed), now completes in a few seconds and passes.
- Full `test_branching.py` module: 61 tests, 0 failures/errors — no regression from moving this write off the branch connection.

### DIFF
--- a/netbox_custom_objects/models.py
+++ b/netbox_custom_objects/models.py
@@ -2229,9 +2229,14 @@ def _rename_objectchange_field_key(fi, old_name, new_name):
         'SET {col} = ({col} - %s) || jsonb_build_object(%s, {col}->%s) '
         'WHERE object_type_id = %s AND {col} IS NOT NULL AND {col} ? %s'
     )
+    # Unlike core.ObjectChange, ChangeDiff has no per-branch schema copy, so every
+    # other access to it (including netbox-branching's own record_change_diff
+    # receiver) uses DEFAULT_DB_ALIAS. Writing it via the branch connection here
+    # instead introduced a second connection into sync()'s open transaction,
+    # causing a same-process cross-connection deadlock (issue #653).
     try:
-        with transaction.atomic(using=conn.alias):
-            with conn.cursor() as cursor:
+        with transaction.atomic(using=DEFAULT_DB_ALIAS):
+            with connections[DEFAULT_DB_ALIAS].cursor() as cursor:
                 for json_col in ('original', 'modified', 'current'):
                     cursor.execute(
                         cd_sql.format(col=json_col),


bctiemann:
@claude /review and limit response to actionable findings.

arthanson (approved):

PR #681, opened 2026-08-25
Fixes #677: Set polymorphic GFK field values before full_clean(), not after

### LINKED ISSUE #677: A Custom Validators instance doesn't include the values Polymorphic objects
### Plugin Version

0.6.1, main branch build

### NetBox Version

4.6.5

### Python Version

3.12.3

### Steps to Reproduce

1. Create a Custom object type ("foo")
2. Add a field, named `member`, that is a `object` selection and check `polymorphic`, for virtual machine and device.
3. Add another field that is a `object` selection that isn't polymorphic, I used another custom object called `application_cluster` here and used that as the field name
4. Add a custom validator that is called by the `foo` custom object, debug the `instance`
_note: the config `table[n]model` needs to reference the object id_
```
  CUSTOM_VALIDATORS: 
    {
      'netbox_custom_objects.table1model': (
          'custom_validators.FooValidator',
        )
     }
```
_validator_
```
class FooValidator(CustomValidator):

    def validate(self, instance, request):
        logger.debug(f"instance dict = {instance.__dict__}")
```
5. Create  `foo` custom object, fill out both selection fields with their own object
_debug will show something like this, note the member returns *None*_
```
instance dict = {
    '_state': < django.db.models.base.ModelState object at 0x7f8e3c989910 > ,
    'id': None,
    'created': None,
    'last_updated': None,
    'owner_id': None,
    'display_name': '',
    'application_cluster ': 35,
    'member_content_type_id': None,
    'member_object_id': None,
    '_changelog_message': None,
    '_m2m_values': {
        'tags': []
    }
}
```
6. Edit the object and, without changing anything, save it again
_debug will show this, note the member has values now_
```
instance dict = {
    '_state': < django.db.models.base.ModelState object at 0x7f8e3ca0cdd0 > ,
    'id': 5,
    'created': datetime.datetime(2026, 8, 24, 1, 3, 31, 306363, tzinfo = datetime.timezone.utc),
    'last_updated': datetime.datetime(2026, 8, 24, 1, 3, 31, 306377, tzinfo = datetime.timezone.utc),
    'owner_id': None,
    'application_cluster ': 35,
    'member_content_type_id': 123,
    'member_object_id': 4,
    '_changelog_message': None,
    '_prechange_snapshot': {
        'created': '2026-08-24T01:03:31.306Z',
        'owner': None,
        'tenant': 35,
        'member_content_type': 123,
        'member_object_id': 4,
        'tags': [],
        '__nco_poly_m2m_fields__': [{
            'name': 'member',
            'pk': 27
        }]
    },
    '_m2m_values': {
        'tags': []
    }
}
```

### Expected Behavior

Member objects added or edited exist on the `instance` in custom validation.
in the specific example above this would be 

```
instance dict = {
    '_state': < django.db.models.base.ModelState object at 0x7f8e3c989910 > ,
    'id': None,
    'created': None,
    'last_updated': None,
    'owner_id': None,
    'display_name': '',
    'application_cluster ': 35,
    'member_content_type_id': 123,
    'member_object_id': 4,
    '_changelog_message': None,
    '_m2m_values': {
        'tags': []
    }
}
```

### Observed Behavior

The `instance`'s loaded values on a polymorphic field are from the object itself, not the values from the form.

For a new object that means the value is `Null`

For a edited object that means the value is the saved value of the object before it was edited, if the edit changes the value of the polymorphic object the `instance` value is the old object before the edit.

bctiemann:
## Summary

Fixes #677: a `CUSTOM_VALIDATORS` validator's `validate(instance, request)` always saw a polymorphic single-object (GFK) field as blank on create, or as its pre-edit value on update — regardless of what was actually submitted in the request.

## Root cause

`CUSTOM_VALIDATORS` runs inside `instance.clean()`, called by Django's `full_clean()`. In both the form and API paths, the polymorphic field's value was applied to the instance strictly *after* `full_clean()` had already run:

- **Form path**: `custom_save()` (called by `form.save()`, itself called *after* `form.is_valid()` — which already triggers `full_clean()` via `ModelForm._post_clean()`) was the only place that did `setattr(instance, field_name, obj_val)` for the field.
- **API path**: the serializer's `validate()` popped the polymorphic field entirely out of the data dict *before* delegating to `ValidatedModelSerializer.validate()` (which builds/mutates the instance and calls `instance.full_clean()`), only restoring the field to the data dict *after* that call returned.

## Fix

- `views.py`: `custom_clean()` (the form's own `clean()`, which runs *before* Django's `_post_clean()` calls `instance.full_clean()`) now also sets the polymorphic field's value on `self.instance`. `custom_save()`'s existing assignment stays in place — reapplying the same value there is a harmless no-op.
- `api/serializers.py`: `validate()` now substitutes the field's two real backing scalar columns (`<name>_content_type_id`, `<name>_object_id`) for the resolved object before delegating to the parent `validate()`, then restores the resolved object under the original field name afterward for `create()`/`update()` to consume as before. Those two columns are real Django model fields (unlike the field itself, a private/virtual GFK descriptor), so an unsaved instance can hold them fine — unlike the polymorphic M2M case (left unchanged), which genuinely has no home on an instance without a PK yet.

## Testing

Added 4 regression tests in `test_polymorphic_fields.py`, covering create and edit/update through both the form and API paths. All 4 fail without the fix — reproducing the exact reported symptom (`None` on create, stale pre-edit value on update) — and pass with it.

Verified:
- `test_polymorphic_fields.py`: 82 tests, 0 real failures (1 error is the confirmed pre-existing, unrelated `netbox_branching` shared-environment artifact, present on unmodified `main` too).
- Full plugin suite in a clean venv: 1165 tests, 0 failures/errors, 2 skipped.

### DIFF
--- a/netbox_custom_objects/api/serializers.py
+++ b/netbox_custom_objects/api/serializers.py
@@ -688,12 +688,10 @@ def get_serializer_class(model, skip_object_fields=False):
         if type(data) is not dict:
             return NetBoxModelSerializer.validate(self, data)
 
-        # NetBoxModelSerializer.validate() calls Model(**attrs) (create) or
-        # setattr(instance, k, v) per attrs key (update), then instance.full_clean()
-        # -- which is what runs CUSTOM_VALIDATORS. Polymorphic GFK and M2M fields
-        # aren't real Django model fields, so passing either straight through
-        # would raise TypeError. Pop them before delegating to the parent, then
-        # restore them afterward.
+        # NetBoxModelSerializer.validate() calls Model(**attrs)/setattr then
+        # instance.full_clean() (which runs CUSTOM_VALIDATORS). Polymorphic M2M
+        # fields aren't real Django model fields, so pop them before delegating
+        # and restore them after.
         # super() is unavailable here because this function is defined outside a
         # class body (no __class__ cell). The generated class has a single base
         # (NetBoxModelSerializer), so calling it directly is equivalent.
@@ -702,14 +700,11 @@ def get_serializer_class(model, skip_object_fields=False):
             if field_name in data:
                 saved[field_name] = data.pop(field_name)
 
-        # Polymorphic single-object (GFK) fields, unlike the M2M case above, ARE
-        # backed by two real scalar columns (<name>_content_type_id,
-        # <name>_object_id) that an unsaved instance can hold. Substitute those
-        # for the resolved object before delegating, so the instance
-        # full_clean() (and CUSTOM_VALIDATORS) actually validates sees the
-        # submitted value instead of a blank/stale one (#677); restore the
-        # resolved object under the original field name afterward, since
-        # create()/update() expect to find it there.
+        # Polymorphic single-object (GFK) fields ARE backed by two real scalar
+        # columns an unsaved instance can hold. Substitute those for the
+        # resolved object before delegating, so full_clean()/CUSTOM_VALIDATORS
+        # see the submitted value instead of a blank one (#677); restore the
+        # resolved object afterward for create()/update() to consume.
         poly_obj_originals = {}
         for field_name in _poly_obj_fields:
             if field_name in data:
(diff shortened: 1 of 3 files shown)


bctiemann:
@claude /review and limit response to actionable findings.

jnovinger (approved):

PR #656, opened 2026-08-14
insert fields marked as 'context' into the SearchIndex results

biwhite:
Fixes: #655

Provide Context tagged fields within the global Search results, in the Attributes column.
This slightly modifies the stock 'SearchIndex' registered on CustomObjects, to include CustomObjectFields tagged as 'Context'

PR #684, opened 2026-08-27
Closes #655: Surface context fields as attributes in global search results

### LINKED ISSUE #655: Global search results do not populate 'attributes' fields for Custom Objects
### Plugin Version

v0.6.0

### Proposed functionality

When returning search results in the global search, stock models provide entries in the 'attributes' column.
We're already capturing 'Context' fields, and as such these can be returned in the attributes field for additional, relevant information to appear on global search.

### Use case

Global Search at the moment only shows primary key link to custom object.  This adds more information into existing space intended to display such information.

### External dependencies

_No response_

bctiemann:
## Summary

Global search results for custom objects only ever showed a bare link to the object — none of its "context" fields appeared in the Attributes column the way core NetBox models do. This wires `CustomObjectType`'s dynamically generated `SearchIndex` to surface `context=True` fields there, via NetBox core's existing `SearchIndex.display_attrs` mechanism.

## Credit

This builds directly on [#656](https://github.com/netboxlabs/netbox-custom-objects/pull/656) by @biwhite, who filed #655 and correctly identified `display_attrs` as the right hook. His PR is out of sync with `main` and failing CI as a result — not because of any problem with his approach (see the comment thread on #655) — and I don't have push access to his fork to rebase it directly, so I'm opening this instead.

## What changed vs. #656

Reviewing #656 for correctness turned up two issues, both addressed here:

1. **Dead code**: #656 added a `get_display_attrs` classmethod to the dynamically generated `SearchIndex` class, but nothing in NetBox core or this plugin ever calls a method by that name. The actual renderer is [`CachedValue.display_attrs`](https://github.com/netbox-community/netbox/blob/main/netbox/extras/models/search.py) (a core `@property`), which only reads the plain `display_attrs` tuple on the indexer class and does its own `getattr()` / `get_<field>_display()` dispatch per entry — the added method was inert.
2. **A real correctness gap**: nothing prevents `context=True` from being set on a `multiobject` (real M2M) field. `CachedValue.display_attrs`'s `getattr()`-based renderer has no per-type dispatch — for a real `ManyToManyField` it returns the `RelatedManager` instance itself, not its contents, so a multiobject context field would leak a broken object repr into search results instead of the related objects. This PR excludes multiobject context fields from `display_attrs` to avoid that. Polymorphic and coordinates fields need no equivalent exclusion — they have no real backing column under the field's own name, so the existing `present`-set guard (already in place to protect stub models generated with `skip_object_fields=True`) filters them out automatically.

Also consolidated the two separate `self.fields` queries (one filtered by `search_weight`, one by `context`) into a single pass.

## Testing

Adds two regression tests to `test_models.py`:
- a `context=True` text field appears in the generated `SearchIndex.display_attrs`
- a `context=True` multiobject field does not

Full plugin suite verified in a clean venv: 1173 tests, 0 failures/errors, 2 skipped.

### DIFF
--- a/netbox_custom_objects/models.py
+++ b/netbox_custom_objects/models.py
@@ -1639,15 +1639,28 @@ class CustomObjectType(NetBoxModel):
             | {f.name for f in model._meta.local_many_to_many}
         )
         fields = []
-        for field in self.fields.filter(search_weight__gt=0):
+        display_attrs = []
+        for field in self.fields.all():
             if field.name not in present:
                 continue
-            fields.append((field.name, field.search_weight))
+            if field.search_weight > 0:
+                fields.append((field.name, field.search_weight))
+            # Context fields surface as supplementary "attributes" on global search
+            # results (issue #655), via the same generic CachedValue.display_attrs
+            # mechanism every other NetBox model's SearchIndex uses. That mechanism
+            # renders a field with plain getattr() (falling back to
+            # get_<field>_display() for Django choices=), so it has no way to render
+            # a MultiObject field's RelatedManager as anything meaningful -- exclude
+            # those. Polymorphic and coordinates fields need no such exclusion: they
+            # have no real backing column under the field's own name, so `present`
+            # already filters them out above.
+            if field.context and field.type != CustomFieldTypeChoices.TYPE_MULTIOBJECT:
+                display_attrs.append(field.name)
 
         attrs = {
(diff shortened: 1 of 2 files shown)


bctiemann:

bctiemann:
Improved both tests as suggested.

arthanson (approved):

Issue #686, opened 2026-08-28
RecursionError when accessing COT models with polymorphic reverse descriptors


vsudevopsam:
### Plugin Version

0.6.0

### NetBox Version

4.6.2

### Python Version

3.12

### Steps to Reproduce

I am encountering a RecursionError: `maximum recursion depth exceeded` when accessing Custom Object Types which have relational links involving polymorphic fields.

This appears to be related to the dynamic model/relation graph recursion described in #685 but the recursion enters through a different path.

In my case, NetBox starts normally. The error occurs later when acting upon a dynamically generated Custom Object model which has relational links, including relationships multiple nodes away. That is, attempting to save, modify, or delete any field on the COT or COTs on a related node upon graph construction causes infinite recursion.

The significant schema changes immediately preceding the problem were:

- Changing `internal_name` values on Custom Object Type Fields
- Explicitly setting/changing `related_name` values on relational fields, including polymorphic fields

No other structural schema changes were made and my schema was previously stable and functional. 

### Expected Behavior

Various related names can be successfully saved on polymorphic fields of COTs.

### Observed Behavior

The problem appears to be associated with polymorphic relationship handling and the reverse-descriptor wiring introduced/fixed in the current plugin code. The recursion follows this path:

`django.db.models.options.Options._relation_tree`
      ↓
`django.db.models.options.Options._populate_directed_relation_graph`
      ↓
`django.apps.registry.Apps.get_models`
      ↓
`netbox_custom_objects.CustomObjectsPluginConfig.get_models`
      ↓
`CustomObjectType.get_model`
      ↓
`CustomObjectType._after_model_generation`
      ↓
`_wire_polymorphic_reverse_descriptors`
      ↓
`field_instance.related_object_types.all()`
      ↓
`Django QuerySet evaluation`
      ↓
`Query.add_q()`
      ↓
`Query.names_to_path()`
      ↓
`Options.get_field()`
      ↓
`Options.fields_map`
      ↓
`Options._get_fields`
      ↓
`Options._relation_tree`
      ↓
`django.apps.registry.Apps.get_models`
      ↓
...

The relevant portion of the traceback is:
File ".../django/db/models/options.py", line 810, in _populate_directed_relation_graph
        all_models = self.apps.get_models(include_auto_created=True)
File ".../django/apps/registry.py", line 185, in get_models
        result.extend(app_config.get_models(include_auto_created, include_swapped))
[...]

Issue #335, opened 2025-12-16
Dynamic Custom Object Models Not Loading in apps.get_models() After v0.4.2 → v0.4.3, Duplicate models 0.4.2


Sergicus:
### Plugin Version

0.4.3

### NetBox Version

4.4.6

### Python Version

3.12.9

### Steps to Reproduce

# 0.4.2

```
./manage.py dbshell
from django.apps import apps
for model in apps.get_models():
    print(f"{model._meta.app_label}.{model.__name__}")
```
  
```
netbox_custom_objects.CustomObjectType
netbox_custom_objects.CustomObjectTypeField
netbox_custom_objects.CustomObjectObjectType
netbox_custom_objects.Table2Model
netbox_custom_objects.Through_custom_objects_1_sites
netbox_custom_objects.Through_custom_objects_1_devices
netbox_custom_objects.Table1Model
**netbox_custom_objects.Table2Model**
**netbox_custom_objects.Table1Model**
```

 v0.4.2 output shows duplicates:
 **netbox_custom_objects.Table2Model  # Duplicate
 netbox_custom_objects.Table1Model  # Duplicate**

# 0.4.3
```
./manage.py dbshell
from django.apps import apps
for model in apps.get_models():
    print(f"{model._meta.app_label}.{model.__name__}")
```

```
# Output (incomplete):
# netbox_custom_objects.CustomObjectType
# netbox_custom_objects.CustomObjectTypeField
# netbox_custom_objects.CustomObjectObjectType
```
**Dynamic models like Table1Model, Table2Model, through models are missing**

### Expected Behavior

```
# Expected output should include:
# netbox_custom_objects.CustomObjectType
# netbox_custom_objects.CustomObjectTypeField
# netbox_custom_objects.CustomObjectObjectType
# netbox_custom_objects.Table1Model
# netbox_custom_objects.Table2Model
# netbox_custom_objects.Through_custom_objects_1_sites
# netbox_custom_objects.Through_custom_objects_1_devices
# etc.
```

### Observed Behavior

Dynamic models like Table1Model, Table2Model, through models are missing

NetBox [Attachments ](https://github.com/Kani999/netbox-attachments) Plugin: Fails to show attachment sections for custom objects because scope filters can't find the models

Other Plugins: Any plugin that enumerates models via apps.get_models() won't see custom object models

Sergicus:
[...]

PR #687, opened 2026-08-31
Fixes #685: guard get_models() against re-entrancy

### LINKED ISSUE #685: `schema/apply/` raises RecursionError when a document creates multiple new, cross-referencing Custom Object Types in one request
### Plugin Version

0.6.1

### NetBox Version

4.6.9

### Python Version

3.14.4

### Steps to Reproduce

1. On a NetBox instance with no existing Custom Object Types, `POST` a portable
   schema document (see `docs/portable-schema.md`) to
   `/api/plugins/custom-objects/schema/apply/` that defines **two or more brand-new
   Custom Object Types in the same request**, where at least one type has an
   `object`/`multiobject` field whose `related_object_type` points at another
   new type in the same document via `custom-objects/<slug>`.

   Minimal example body:

   ```json
   {
     "allow_destructive": false,
     "schema": {
       "schema_version": "1",
       "types": [
         {
           "name": "ospf_instance",
           "slug": "ospf-instances",
           "fields": [
             { "id": 1, "name": "name", "type": "text", "primary": true, "required": true, "unique": true }
           ]
         },
         {
           "name": "ospf_area",
           "slug": "ospf-areas",
           "fields": [
             { "id": 1, "name": "name", "type": "text", "primary": true, "required": true, "unique": true },
             {
               "id": 2,
               "name": "instance",
               "type": "object",
               "required": true,
               "related_object_type": "custom-objects/ospf-instances"
             }
           ]
         }
       ]
     }
   }
   ```

2. Observe the response.

Notes on scope:
- Applying the exact same two-type document works fine if you split it into two
  separate `schema/apply/` requests — one per type, in dependency order (create
  `ospf-instances` first, then `ospf-areas`).
- Calling `netbox_custom_objects.schema.executor.apply_document()` directly from
  `manage.py nbshell` with the same multi-type document also succeeds — the bug
  only manifests when the call happens inside a live Django request/response
  cycle (i.e. via the `SchemaApplyView` API endpoint).
- Reproduced with a 2-type and a 3-type document (3rd type
  `ospf_interface` -> `custom-objects/ospf-areas`); both fail identically.

## Expected Behavior

`POST /api/plugins/custom-objects/schema/apply/` should create all of the new,
cross-referencing Custom Object Types described in the document in a single
atomic call, returning `200` with `{"applied": true, "diffs": [...]}`.

This is not just an assumption — `docs/portable-schema.md` documents multi-COT,
single-request apply as the supported path. Its "Dependency Ordering" section
(under "Executor (Developer Reference)") states:

> When a schema document contains multiple new COTs that reference each other
> via `related_object_type: "custom-objects/<slug>"`, the executor performs a
> topological sort to ensure referenced COT tables exist before any referencing
> field is added. Cycles among new COTs raise `CircularDependencyError`.

The only failure case called out there is a dependency cycle. Nothing in
`docs/portable-schema.md` (Usage, Typical End-to-End Workflow, or the API
request/response reference) suggests that a document with multiple new,
non-circular, cross-referencing COTs needs to be split into separate requests.

### Expected Behavior

`POST /api/plugins/custom-objects/schema/apply/` should create all of the new,
cross-referencing Custom Object Types described in the document in a single
atomic call, returning `200` with `{"applied": true, "diffs": [...]}`.

This is not just an assumption — `docs/portable-schema.md` documents multi-COT,
single-request apply as the supported path. Its "Dependency Ordering" section
(under "Executor (Developer Reference)") states:

> When a schema document contains multiple new COTs that reference each other
> via `related_object_type: "custom-objects/<slug>"`, the executor performs a
> topological sort to ensure referenced COT tables exist before any referencing
> field is added. Cycles among new COTs raise `CircularDependencyError`.

The only failure case called out there is a dependency cycle. Nothing in
`docs/portable-schema.md` (Usage, Typical End-to-End Workflow, or the API
request/response reference) suggests that a document with multiple new,
non-circular, cross-referencing COTs needs to be split into separate requests.

### Observed Behavior

The request fails with `500 Internal Server Error`:

```json
{
  "error": "maximum recursion depth exceeded",
  "exception": "RecursionError",
  "netbox_version": "4.6.9-Docker-5.0.2",
  "python_version": "3.14.4"
}
```

Full traceback (captured by calling the view directly via Django's test
`Client`, bypassing the API's generic 500 handler):

```
  File ".../netbox_custom_objects/__init__.py", line 597, in get_models
    model = custom_type.get_model()
  File ".../netbox_custom_objects/models.py", line 1819, in get_model
    self._after_model_generation(attrs, model)
  File ".../netbox_custom_objects/models.py", line 1538, in _after_model_generation
    own_ot = ObjectType.objects.get_for_model(model)
  File ".../netbox/core/models/object_types.py", line 98, in get_for_model
    ot = self.get_or_create(
         ...
    )[0]
  File ".../django/db/models/manager.py", line 87, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File ".../django/db/models/query.py", line 994, in get_or_create
    return self.create(**params), True
  File ".../netbox/core/models/object_types.py", line 34, in create
    return super().create(**kwargs)
  File ".../django/db/models/query.py", line 659, in create
    self.model._meta._reverse_one_to_one_field_names
  File ".../django/utils/functional.py", line 47, in __get__
    res = instance.__dict__[self.name] = self.func(instance)
  File ".../django/db/models/options.py", line 1030, in _reverse_one_to_one_field_names
    field.name for field in self.related_objects if field.one_to_one
  File ".../django/db/models/options.py", line 621, in related_objects
    all_related_fields = self._get_fields(
  File ".../django/db/models/options.py", line 925, in _get_fields
    for obj in parent._meta._get_fields(
  File ".../django/db/models/options.py", line 943, in _get_fields
    all_fields = self._relation_tree
  File ".../django/utils/functional.py", line 47, in __get__
    res = instance.__dict__[self.name] = self.func(instance)
  File ".../django/db/models/options.py", line 843, in _relation_tree
    return self._populate_directed_relation_graph()
  File ".../django/db/models/options.py", line 810, in _populate_directed_relation_graph
    all_models = self.apps.get_models(include_auto_created=True)
  File ".../django/apps/registry.py", line 185, in get_models
    result.extend(app_config.get_models(include_auto_created, include_swapped))
  File ".../netbox_custom_objects/__init__.py", line 597, in get_models
    model = custom_type.get_model()
  File ".../netbox_custom_objects/models.py", line 1819, in get_model
    self._after_model_generation(attrs, model)
  ... (repeats, cycling back through get_for_model / get_or_create / create /
       _relation_tree / _populate_directed_relation_graph / get_models /
       get_model, growing the stack each time) ...
  File ".../django/db/models/fields/related.py", line 1191, in get_prep_value
    return self.target_field.get_prep_value(value)
RecursionError: maximum recursion depth exceeded
```

The `transaction.atomic()` wrapper in the executor does correctly roll back —
after the 500, `GET /api/plugins/custom-objects/custom-object-types/` shows
`"count": 0`, so no partial/orphaned COTs or tables are left behind. The bug is
a hard failure of the request, not data corruption.

### Suspected Cause

`CustomObjectsPluginConfig.get_models()` (`netbox_custom_objects/__init__.py:597`)
overrides Django's app-registry model enumeration and, for each `CustomObjectType`,
calls `get_model()` to lazily generate the dynamic model class. `get_model()` in
turn calls `_after_model_generation()`, which calls
`ObjectType.objects.get_for_model(model)` to ensure a NetBox `ObjectType`
(ContentType wrapper) exists for the newly generated model.

`ObjectType.objects.get_or_create()` → `.create()` accesses
`model._meta._reverse_one_to_one_field_names`, which is a cached property that,
on first access, triggers Django to build the global relation graph via
`self._meta._relation_tree` → `apps.get_models(include_auto_created=True)`.
Because `get_models()` is overridden by the plugin, this re-enters
`custom_type.get_model()` for **every** `CustomObjectType` row, including ones
newly created earlier in the same `apply_document()` call whose FK
(`related_object_type` pointing at `custom-objects/<slug>`) has not finished
being fully wired up yet.

For those still-being-generated models, `get_model()` re-triggers
`_after_model_generation()` → `get_for_model()` → the same relation-graph build
→ `get_models()` again, and so on — an unbounded mutual recursion between
Django's relation-graph cache population and the plugin's dynamic model
generation, specific to the window where 2+ new, cross-referencing COTs exist
in the same uncommitted transaction and haven't each individually completed
`_after_model_generation()` yet.

This does not reproduce when:
- Only one new COT is created per `apply_document()`/API call (no other
  not-yet-finalized COT for `get_models()` to trip over), or
- The call happens outside a fresh request context where Django's relation-tree
  cache (`Options._relation_tree`) may already be warm from prior activity in
  the same process (e.g. an `nbshell` session that has already touched the app
  registry), which appears to avoid re-entering
  `_populate_directed_relation_graph()` altogether.

### Proposed Fix

A few options, roughly in order of invasiveness:

1. In the executor (`netbox_custom_objects/schema/executor.py`), when applying
   a document with multiple new COTs, fully finalize each COT (including its
   `ObjectType`/`ContentType` registration) one at a time, in the existing
   topological order, before moving on to the next — rather than creating all
   COT rows/fields first and letting model generation/`ObjectType` resolution
   happen lazily and interleaved.
2. Pre-create the `ObjectType` row for each new COT's model up front (before
   any field with a `related_object_type` reference to it is added), so that by
   the time a later COT's FK needs to resolve that `ObjectType`,
   `get_for_model()` short-circuits on the existing row instead of calling
   `.create()` and rebuilding the relation graph.
3. Guard `CustomObjectsPluginConfig.get_models()` against re-entrancy (e.g. a
   thread-local/context-var flag set while a `get_model()` call for a given COT
   is already in progress), returning already-registered models without
   recursing into generation for COTs currently mid-construction.

As a workaround in the meantime, callers of `schema/apply/` can submit each new,
cross-referencing Custom Object Type as its own request, in dependency order
(referenced types before referencing types) — this was verified to succeed
reliably against 4.6.9 / plugin 0.6.1.

bctiemann:
## Summary

`POST /api/plugins/custom-objects/schema/apply/` raised `RecursionError` when a document created two or more new, cross-referencing Custom Object Types in a single request. Generating a brand-new COT's model calls `_after_model_generation()`, which can itself need Django's global relation graph (`Options._relation_tree`) -- e.g. via `ObjectType.objects.get_for_model()`'s `.create()`. Rebuilding that graph calls `apps.get_models()`, which re-enters this plugin's own `get_models()` while it's still mid-generation. Without a guard, that re-entrant call walks `CustomObjectType.objects.all()` and calls `get_model()` again for every COT -- including the one still under construction -- with no way to ever finish.

#686 (closed as a duplicate) hit the *identical* `get_models()` recursion through a completely different call site: `_wire_polymorphic_reverse_descriptors()` -> `field_instance.related_object_types.all()`, triggered during ordinary polymorphic-field save/modify, not the executor at all.

## Evaluating the proposed fix

The issue's own investigation laid out three options, roughly by invasiveness: (1) serialize COT finalization in the executor, (2) pre-create `ObjectType` rows before any FK reference resolves, (3) a re-entrancy guard on `get_models()` itself. @jnovinger's review recommended option 3 specifically *because* of #686 -- an executor-only fix (1 or 2) would resolve #685 but leave #686, and any future path into the same hazard, live. This PR implements option 3.

A `contextvars.ContextVar`-based guard (matching the existing `_is_migrating` idiom already in this file) is safe here because `generate_model()`'s `type()` call already registers a COT's model with Django's app registry *synchronously*, before `get_model()` ever calls `_after_model_generation()` (the method that can trigger this re-entrancy). So a re-entrant call can simply fall back to `super().get_models()` (everything already registered) without needing to regenerate anything -- any COT the outer loop hasn't reached yet is just absent from that transient snapshot, and self-heals via `get_model()`'s own `apps.clear_cache()` once the remaining COTs finish generating.

## Testing

Per jnovinger's explicit ask, adds a regression test for **each** re-entry path:

- `SchemaApplyMultiCOTRecursionTestCase` (`test_schema_api.py`) -- the executor path from #685's own repro (2-type and 3-type chained-reference documents posted through the real API view).
- `PolymorphicReverseDescriptorRecursionTestCase` (`test_polymorphic_fields.py`) -- the descriptor-wiring path from #686.

Both needed one non-obvious adjustment to actually exercise the vulnerable code: `get_models()`'s `CustomObjectType`-enumeration loop is unconditionally disabled under `manage.py test` (`should_skip_dynamic_model_creation()` returns `True` whenever `"test" in sys.argv`), so without patching around that, these tests silently no-op regardless of the fix. Both tests now patch `_app_ready` / `should_skip_dynamic_model_creation` to replicate a non-test process, and each was verified to reproduce a genuine `RecursionError` against this commit's parent, then pass cleanly with the fix.

`test_schema_api.py` also gets a third, deterministic white-box test (`test_get_models_guards_against_reentrant_cot_generation`) that simulates the re-entrant trigger directly via a call-counting spy on `get_model()`, so the invariant is verified independent of whatever incidental factors (cache warmth, interpreter version, stack depth) determine whether a given environment happens to blow past Python's recursion limit.

Full `netbox_custom_objects` suite verified in the shared dev venv: 1179 tests, 0 new failures (15 pre-existing errors, all attributable to the sibling `netbox-branching` checkout's known incompatibility with this venv's Django/NetBox version -- confirmed to fail identically in isolation, unrelated to this change).

### DIFF
--- a/netbox_custom_objects/__init__.py
+++ b/netbox_custom_objects/__init__.py
@@ -18,6 +18,15 @@ logger = logging.getLogger(__name__)
 # Context variable to track if we're currently running migrations
 _is_migrating = contextvars.ContextVar('is_migrating', default=False)
 
+# Guards get_models() against re-entrancy (issues #685/#686): generating a COT
+# model can itself trigger Django to rebuild its global relation graph (e.g. via
+# ObjectType.objects.get_for_model()'s .create(), or a polymorphic field's
+# related_object_types.all() query), which calls apps.get_models() again. Without
+# this guard, that re-entrant call would call get_model() again for every COT --
+# including the one still mid-construction -- recursing without ever reaching the
+# point where the first call finishes and registers/caches it.
+_generating_models = contextvars.ContextVar('generating_models', default=False)
+
 # Cache for migration check to avoid repeated expensive filesystem/database operations
 _migrations_checked = None
 _checking_migrations = False
@@ -560,52 +569,76 @@ class CustomObjectsPluginConfig(PluginConfig):
         for model in super().get_models(include_auto_created, include_swapped):
             yield model
 
-        # Suppress warnings about database calls during model loading.
-        # See the corresponding block in ready() for the rationale for using
-        # module-based filtering instead of message-content matching.
-        with warnings.catch_warnings():
-            warnings.filterwarnings(
-                "ignore", category=RuntimeWarning,
-                module=r"django\.db\.backends\..*",
-            )
-            warnings.filterwarnings(
-                "ignore", category=UserWarning,
-                module=r"netbox_branching\..*",
-            )
+        # Re-entrant call (issues #685/#686): something invoked while WE are
+        # already generating COT models needed Django's model registry -- most
+        # commonly Options._relation_tree, via ObjectType.objects.get_for_model()
+        # or a polymorphic field's related_object_types.all() query. Recursing
+        # into get_model() again here would keep regenerating the same
+        # still-under-construction model forever, since the outer call hasn't
+        # reached the point where it finishes and registers/caches it.
+        #
+        # super().get_models() above already covers everything this caller can
+        # safely see right now: generate_model()'s type() call registers a COT's
+        # model with Django's app registry synchronously (inside ModelBase.__new__),
+        # before get_model() ever calls _after_model_generation() -- the method
+        # that can trigger this re-entrancy. Any COT the outer loop below hasn't
+        # reached yet is simply absent from this transient, incomplete snapshot;
+        # get_model()'s own apps.clear_cache() call (once each COT finishes)
+        # invalidates any relation-tree computed against that incomplete view, so
+        # it self-heals as soon as the remaining COTs are generated.
+        if _generating_models.get():
+            return
 
-            # Skip dynamic model generation until ready() has completed.
-            # Other apps' ready() calls (e.g. dcim) trigger _relation_tree →
-            # apps.get_models() before our ready() runs.  At that point _model_cache
-            # is empty, so get_model() would regenerate every COT from scratch —
-            # including ContentType DB lookups that may fail.  After our ready()
-            # finishes, _app_ready is True and get_model() returns cached models
-            # without any ContentType lookups.
-            if not _app_ready:
-                return
+        token = _generating_models.set(True)
+        try:
+            # Suppress warnings about database calls during model loading.
+            # See the corresponding block in ready() for the rationale for using
+            # module-based filtering instead of message-content matching.
+            with warnings.catch_warnings():
+                warnings.filterwarnings(
+                    "ignore", category=RuntimeWarning,
+                    module=r"django\.db\.backends\..*",
+                )
+                warnings.filterwarnings(
+                    "ignore", category=UserWarning,
+                    module=r"netbox_branching\..*",
+                )
 
-            # Skip custom object type model loading if dynamic models can't be created yet
-            if self.should_skip_dynamic_model_creation():
-                return
+                # Skip dynamic model generation until ready() has completed.
+                # Other apps' ready() calls (e.g. dcim) trigger _relation_tree →
+                # apps.get_models() before our ready() runs.  At that point _model_cache
+                # is empty, so get_model() would regenerate every COT from scratch —
+                # including ContentType DB lookups that may fail.  After our ready()
+                # finishes, _app_ready is True and get_model() returns cached models
+                # without any ContentType lookups.
+                if not _app_ready:
+                    return
 
-            # Add custom object type models
-            from .models import CustomObjectType
+                # Skip custom object type model loading if dynamic models can't be created yet
+                if self.should_skip_dynamic_model_creation():
+                    return
 
-            try:
-                with transaction.atomic():
-                    custom_object_types = CustomObjectType.objects.all()
-                    for custom_type in custom_object_types:
-                        model = custom_type.get_model()
-                        if model:
-                            yield model
+                # Add custom object type models
+                from .models import CustomObjectType
 
-                            # If include_auto_created is True, also yield through models
-                            if include_auto_created and hasattr(model, '_through_models'):
-                                for through_model in model._through_models:
-                                    yield through_model
-            except (ProgrammingError, OperationalError):
-                # DB schema is incomplete (unapplied migrations). Yield nothing —
-                # dynamic models will be available once migrations have run.
-                return
+                try:
+                    with transaction.atomic():
+                        custom_object_types = CustomObjectType.objects.all()
+                        for custom_type in custom_object_types:
+                            model = custom_type.get_model()
+                            if model:
+                                yield model
+
+                                # If include_auto_created is True, also yield through models
+                                if include_auto_created and hasattr(model, '_through_models'):
+                                    for through_model in model._through_models:
+                                        yield through_model
+                except (ProgrammingError, OperationalError):
+                    # DB schema is incomplete (unapplied migrations). Yield nothing —
+                    # dynamic models will be available once migrations have run.
+                    return
+        finally:
+            _generating_models.reset(token)
 
 
 config = CustomObjectsPluginConfig
--- a/netbox_custom_objects/__init__.py
+++ b/netbox_custom_objects/__init__.py
@@ -19,12 +19,8 @@ logger = logging.getLogger(__name__)
 _is_migrating = contextvars.ContextVar('is_migrating', default=False)
 
 # Guards get_models() against re-entrancy (issues #685/#686): generating a COT
-# model can itself trigger Django to rebuild its global relation graph (e.g. via
-# ObjectType.objects.get_for_model()'s .create(), or a polymorphic field's
-# related_object_types.all() query), which calls apps.get_models() again. Without
-# this guard, that re-entrant call would call get_model() again for every COT --
-# including the one still mid-construction -- recursing without ever reaching the
-# point where the first call finishes and registers/caches it.
+# model can trigger Django to rebuild its relation graph, which calls
+# apps.get_models() again while we're still mid-generation.
 _generating_models = contextvars.ContextVar('generating_models', default=False)
 
 # Cache for migration check to avoid repeated expensive filesystem/database operations
@@ -569,23 +565,10 @@ class CustomObjectsPluginConfig(PluginConfig):
         for model in super().get_models(include_auto_created, include_swapped):
             yield model
 
-        # Re-entrant call (issues #685/#686): something invoked while WE are
-        # already generating COT models needed Django's model registry -- most
-        # commonly Options._relation_tree, via ObjectType.objects.get_for_model()
-        # or a polymorphic field's related_object_types.all() query. Recursing
-        # into get_model() again here would keep regenerating the same
-        # still-under-construction model forever, since the outer call hasn't
-        # reached the point where it finishes and registers/caches it.
-        #
-        # super().get_models() above already covers everything this caller can
-        # safely see right now: generate_model()'s type() call registers a COT's
-        # model with Django's app registry synchronously (inside ModelBase.__new__),
-        # before get_model() ever calls _after_model_generation() -- the method
-        # that can trigger this re-entrancy. Any COT the outer loop below hasn't
-        # reached yet is simply absent from this transient, incomplete snapshot;
-        # get_model()'s own apps.clear_cache() call (once each COT finishes)
-        # invalidates any relation-tree computed against that incomplete view, so
-        # it self-heals as soon as the remaining COTs are generated.
+        # Re-entrant call (issues #685/#686): generate_model() registers a COT's
+        # model synchronously, before get_model() can trigger this recursion, so
+        # super().get_models() above already covers everything safely visible
+        # here -- fall back to that instead of regenerating.
         if _generating_models.get():
             return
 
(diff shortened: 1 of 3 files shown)


bctiemann:

bctiemann:
Re: finding #1 (Python 3.10 compatibility) -- looked into this. NetBox core itself requires Python 3.12+ as of v4.5 (`requires-python = ">=3.12"`, and the install docs say outright "NetBox supports only Python 3.12 or later"). Since this plugin only ever runs loaded inside a NetBox instance, there's no real deployment path where its test suite executes under 3.10/3.11. The `>=3.10` floor in this repo's `pyproject.toml`/`AGENTS.md`/`ruff.toml` looks like stale metadata that was never updated when NetBox moved its own floor -- CI only ever tests 3.12, and nothing else in the codebase avoids 3.11+ features.

Leaving `enterContext()` as-is here. We'll correct the stale `requires-python`/classifiers/ruff target in the v0.7.0 release PR instead of touching test code for a constraint that isn't real.

Finding #2 (repeated inline imports) is a fair, uncontroversial cleanup -- fixing that now.

jnovinger (approved):
The guard is right and the registration-ordering argument holds, so `super().get_models()` really does cover the model under construction.

Two potential test improvements:

- The tests assert "didn't raise" plus existence, so they'd pass just as well against a `get_models()` that returned zero COT models — which is the degradation mode the guard introduces. `test_schema_operations.py`'s #335 test has the shape I'd want here: `assertIn(model, apps.get_models())`.

- Since `get_models()` is a generator, the `set`/`reset` pair is scoped to iteration rather than a call frame, so an `assertFalse(_generating_models.get())` after the call would pin that too. It can't fail against the `list(...)` the tests use today, but it's what would catch a later refactor that leaves the flag set and starts silently truncating everyone else's model list.

bctiemann:
Both applied, thanks.

- `assertIn(model, apps.get_models())` added to all three end-to-end/deterministic tests, plus `assertFalse(_generating_models.get())` after every guarded call.

- Applying the `assertIn` check to the #686 polymorphic-descriptor test surfaced a real bug, not a flaw in the test: `cot.get_model()` called directly (not via `get_models()`'s own loop) on a not-yet-cached COT can trigger `generate_model()` **twice** for the same COT -- once for the direct call, and once more when `_wire_polymorphic_reverse_descriptors()` re-enters `get_models()` before the direct call has cached anything (this reentrancy guard isn't set yet at that point, since the direct call never went through `get_models()`'s loop itself). The two generations register different classes in different places: `get_model()`'s return value and `CustomObjectType._model_cache` end up holding one class, while Django's app registry (`apps.all_models`, and therefore everything that resolves models through it) ends up holding the other.

  Confirmed via instrumentation -- `generate_model()` fires twice, producing two distinct class objects, with the divergence landing exactly where described. This guard correctly prevents it from recursing infinitely (bounded to 2 generations instead of a crash), but doesn't prevent the resulting cache/registry split. It's been latent in the caching design all along; before this fix, this exact interleaving always hit `RecursionError` before reaching a stable state, so it was never observable.

  Filed as #688 with the full repro and root-cause trace, out of scope for this PR. The test now looks up the actually-registered class via `apps.get_model()` rather than trusting `get_model()`'s return value, with an inline comment pointing at #688.

Verified: all 4 tests fail with a genuine `RecursionError` against the pre-fix code and pass with the fix. Full `test_schema_api.py` + `test_polymorphic_fields.py` run: 113 tests, 0 new failures (1 pre-existing, unrelated `netbox-branching` environment error, same as before).

bctiemann:
The two `tests (main, branching)` CI failures here are unrelated to this PR's changes.

Root cause: NetBox's `main` branch was released as **v4.7.0 today** (2026-09-02). `netboxlabs-netbox-branching`'s latest release (v1.1.3, 2026-08-26) still declares a maximum supported NetBox version of 4.6.99, so this workflow's "main, branching" matrix leg -- which floats both `netbox-ref: main` and `netboxlabs-netbox-branching>=1.0.4,<2.0.0` -- now installs an incompatible combination. The plugin correctly refuses to load at that version mismatch, but `test_branching.py` imports `netbox_branching.models` directly regardless of whether the plugin loaded, and that import crashes before Django's app-registry check even runs.

Confirmed this is repo-wide, not branch-specific: manually dispatched this same workflow directly against `main` (https://github.com/netboxlabs/netbox-custom-objects/actions/runs/33687827160) with zero code changes, and `tests (main, branching)` fails there too, while `tests (main)` and `tests (feature)` both pass. The last green run of this leg was two days ago, before NetBox's 4.7.0 cut.

This needs a fix on the CI/dependency side (pin the branching leg's `netbox-ref` to a version netbox-branching actually supports, or wait for a netbox-branching release compatible with 4.7.0) rather than anything in this PR.

bctiemann:
Update: netbox-branching 1.2.0 fixes the version-gate crash from my earlier comment (reran both failed jobs here after the release -- the plugin now loads correctly under NetBox 4.7.0), but surfaces two new, genuine test failures that are unrelated to this PR's diff: `test_cot_deleted_in_branch_merge_and_revert` (NOT NULL violation on `cache_timestamp` during `branch.revert()`) and `test_single_field_delete_merge_and_revert` (a dropped column isn't restored on revert), both in `IterativeBranchingTestCase` and `SquashBranchingTestCase`.

Confirmed these are pre-existing and not caused by anything here: dispatched `lint-tests.yaml` directly against `main` with zero code changes (https://github.com/netboxlabs/netbox-custom-objects/actions/runs/33695787237) and got the identical failure signature -- 61 tests, 2 failures, 2 errors, same test names.

Filed as #689 with full tracebacks and a working theory (likely related to 1.2.0's mptt->ltree/trigger migration per its release notes). This PR's own tests (`test_schema_api.py`, `test_polymorphic_fields.py`) are unaffected and still pass cleanly.

jnovinger (approved):

PR #691, opened 2026-09-09
Fixes #688: Re-register get_model()'s own class after a re-entrant regeneration

### LINKED ISSUE #688: get_model() can leave _model_cache and apps.all_models pointing at two different classes for the same COT
### Summary

`CustomObjectType._model_cache` and Django's app registry (`apps.all_models`) can end up holding two different Python classes for the same `CustomObjectType`, when generating that COT's model triggers a re-entrant `apps.get_models()` call (e.g. via a polymorphic field's `related_object_types.all()` query in `_wire_polymorphic_reverse_descriptors()`, per #686) while the COT's model isn't cached yet.

Discovered while adding a regression test for #685/#686 in #687 (per @jnovinger's review suggestion to assert the generated model is actually returned by `apps.get_models()`, rather than just checking `get_model()` doesn't raise). The #685/#686 reentrancy guard correctly prevents this from recursing infinitely, but doesn't prevent the underlying interleaving hazard it exposes — before that fix, this exact sequence always crashed with `RecursionError` before reaching a stable (if incorrect) state, so the divergence was never observable.

### Reproduction

```python
cot = CustomObjectType.objects.create(name="Test", slug="test")
CustomObjectTypeField.objects.create(custom_object_type=cot, name="name", type="text", primary=True)
field = CustomObjectTypeField.objects.create(
    custom_object_type=cot, name="target_obj", type="object",
    is_polymorphic=True, related_name="rev_test",
)
field.related_object_types.set([site_ot, prefix_ot])

cot.clear_model_cache(cot.id)
apps.clear_cache()

model = cot.get_model()  # generates the model
```

Instrumenting `generate_model()` shows it's called **twice** for this single `get_model()` call:

```
generate_model() call #1 -> <class 'database.models.Table16Model'> id=...336
generate_model() call #2 -> <class 'database.models.Table16Model'> id=...672
returned model:  id=...336
cached model:    id=...336   (CustomObjectType.get_cached_model(cot.id))
registered model: id=...672  (apps.all_models['netbox_custom_objects']['table16model'])
```

### Root cause

In `CustomObjectType.get_model()` (`netbox_custom_objects/models.py`), `apps.register_model()` runs (~line 1819) *before* `_after_model_generation()` (~line 1832), and the per-COT `_model_cache` write (~line 1894) runs *after* `_after_model_generation()` completes:

1. **Outer call** (`cot.get_model()`) generates model A, registers it via `apps.register_model()`, then runs `_after_model_generation()` → `_wire_polymorphic_reverse_descriptors()` → `related_object_types.all()` → triggers Django's relation-graph rebuild → `apps.get_models()`.
2. This is the *first* entry into `CustomObjectsPluginConfig.get_models()` for this call stack (the outer call never went through `get_models()`'s own loop, so the #685/#686 reentrancy guard isn't set yet). The reentrant call proceeds into its `CustomObjectType.objects.all()` loop and calls `get_model()` again for the *same* COT — which is still mid-generation and not yet cache-hit-eligible (the outer call's `_model_cache` write hasn't happened yet).
3. This **nested call** generates model B and re-registers it via `apps.register_model()`, overwriting model A's registry entry. Its own `_after_model_generation()` triggers another `apps.get_models()` re-entry, but this time the guard *is* set (from step 2's own loop), so it returns early — bounding the recursion instead of crashing.
4. Execution unwinds back to the outer call, which finishes and writes **model A** into `_model_cache` (~line 1894) — overwriting model B there.

End state: `_model_cache` (what `get_model()` returns on a future cache hit) holds model A; `apps.all_models` / `apps.get_models()` (what the ORM, `ContentType.model_class()`, serializers, etc. actually resolve) holds model B. Two distinct, non-identical classes for the same table.

### Impact

Any code that resolves the model via `get_model()`/`get_cached_model()` gets a different Python class than code resolving it via the app registry (`apps.get_model()`, `ContentType.model_class()`, `apps.get_models()` enumeration). FK bindings, reverse descriptors, and identity checks (`isinstance`, `model in some_list`) established against one class won't match the other, which can produce confusing failures anywhere the two code paths meet for the same COT.

### Suggested fix directions

- Re-register the outer call's final model with `apps.register_model()` *after* `_after_model_generation()` completes, so the last write always matches what gets cached, regardless of any nested regeneration that happened in between.
- Alternatively, have the nested get_models() loop skip regenerating a COT that's already mid-generation on the current call stack (a finer-grained, per-COT guard rather than just the whole-loop `_generating_models` flag), so the nested call sees the outer's in-progress generation instead of racing it.

Related: #685, #686, #687.

bctiemann:
## Summary

`CustomObjectType._model_cache` and Django's app registry (`apps.all_models`) could end up holding two different Python classes for the same `CustomObjectType`, when generating that COT's model triggers a re-entrant `apps.get_models()` call (e.g. via a polymorphic field's `related_object_types.all()` query in `_wire_polymorphic_reverse_descriptors()`, per #686) while the COT's model isn't cached yet. The #685/#686 reentrancy guard correctly prevents this from recursing infinitely, but doesn't prevent the underlying interleaving hazard it exposes -- discovered while adding a regression test for #685/#686 in #687, per @jnovinger's review suggestion to assert the generated model is actually returned by `apps.get_models()`.

Full root-cause writeup with instrumented repro output is in #688.

## Root cause (recap)

In `CustomObjectType.get_model()`, `apps.register_model()` runs *before* `_after_model_generation()`, and the per-COT `_model_cache` write runs *after* it completes. When `_after_model_generation()` triggers a re-entrant `get_model()` call for the *same* COT (via the relation-tree-rebuild path described above), that nested call generates and registers a second, different class under the same `apps.all_models` key -- clobbering the outer call's registration. The outer call's own `_model_cache` write happens last and correctly ends up holding the outer call's class, but nothing re-asserts `apps.all_models` afterward, so it's left pointing at the nested call's stray class instead.

End state: `_model_cache` (and `get_model()`'s return value) hold one class; `apps.all_models` / `apps.get_models()` (what the ORM, `ContentType.model_class()`, serializers, etc. actually resolve) hold a different one.

## Fix

Implements the first of #688's two suggested fix directions: re-register the outer call's own `model` in `apps.all_models` immediately after `_after_model_generation()` completes (guarded to the `branch_id is None` case, matching the existing registration logic just above it -- branch contexts never register their own class in `apps.all_models` in the first place). This guarantees the final registered class always matches what's about to be cached and returned, regardless of any nested regeneration that happened in between.

The alternative direction (a finer-grained, per-COT "already generating" guard so the nested call sees the outer's in-progress generation instead of racing it) would also work, but is more invasive -- it would need to thread per-COT state through the reentrancy guard rather than the whole-loop flag added in #687. Re-asserting the final registration is a smaller, more localized change with the same effect: whichever call the caller is actually waiting on becomes the canonical class everywhere, deterministically, rather than "whichever regeneration happened to run last" winning based on incidental Django-internals ordering.

## Testing

Added `GetModelCacheRegistryConsistencyTestCase` in `test_polymorphic_fields.py`, reusing the exact reproduction scenario from #686/#688 (a polymorphic field with a `related_name`, model cache and relation-tree forced cold). Asserts `get_model()`'s return value, `CustomObjectType.get_cached_model()`, and `apps.get_model()` are all the *same* class object (`assertIs`, not just equal).

Verified:
- Against unfixed code, the test fails with exactly the divergence #688 describes (`<class 'database.models.Table1Model'> is not <class 'database.models.Table1Model'>` -- two distinct objects of the same name).
- With this fix, it passes.
- Full `test_polymorphic_fields.py` suite (85 tests) against `netbox.configuration_co_testing`: 0 new failures (1 pre-existing, unrelated error in `ReferencedObjectDeletionTest` -- `CustomObjectType.delete()`'s `netbox_branching` import guard catches `ImportError` but this config's shared dev venv has `netbox_branching` importable-but-not-in-`INSTALLED_APPS`, raising `RuntimeError` instead; confirmed present on `main` with no code changes).

### DIFF
--- a/netbox_custom_objects/models.py
+++ b/netbox_custom_objects/models.py
@@ -1831,6 +1831,22 @@ class CustomObjectType(NetBoxModel):
             with self._global_lock:
                 self._after_model_generation(attrs, model)
 
+                # _after_model_generation() can trigger a re-entrant get_model() for this
+                # SAME (cot_id, branch_id) -- e.g. a polymorphic field's
+                # related_object_types.all() rebuilding Django's relation tree, which
+                # re-enters our get_models() override before the #685/#686 reentrancy
+                # guard is set (this call arrived directly, not via get_models()'s own
+                # loop) and regenerates+re-registers a *different* class object under the
+                # same apps.all_models key (#688). That nested call's own _model_cache
+                # write is superseded below by this call's, but nothing previously
+                # re-asserted apps.all_models afterward, so it could keep pointing at the
+                # nested call's class while _model_cache (and this call's return value)
+                # point at this one. Re-register this call's own model so both agree.
+                if branch_id is None and apps.all_models[APP_LABEL].get(model_key) is not model:
+                    if model_key in apps.all_models[APP_LABEL]:
+                        del apps.all_models[APP_LABEL][model_key]
+                    apps.register_model(APP_LABEL, model)
+
                 # When this COT's model is regenerated (cache miss), non-polymorphic through
                 # models owned by OTHER COTs that point to this COT as their M2M target keep
                 # their target FK stale (pointing at the old model class).  Django's deletion
--- a/netbox_custom_objects/models.py
+++ b/netbox_custom_objects/models.py
@@ -1832,9 +1832,9 @@ class CustomObjectType(NetBoxModel):
                 self._after_model_generation(attrs, model)
 
                 # _after_model_generation() can trigger a re-entrant get_model() for this
-                # same (cot_id, branch_id) (#688) -- e.g. a polymorphic field's
+                # same (cot_id, branch_id) -- e.g. a polymorphic field's
                 # related_object_types.all() rebuilding the relation tree and re-entering
-                # get_models() before the #685/#686 guard is set -- which re-registers a
+                # get_models() before the reentrancy guard is set -- which re-registers a
                 # different class under this key. Re-assert this call's own class so
                 # apps.all_models matches _model_cache and the return value below.
                 if branch_id is None and apps.all_models[APP_LABEL].get(model_key) is not model:
--- a/netbox_custom_objects/models.py
+++ b/netbox_custom_objects/models.py
@@ -1832,16 +1832,11 @@ class CustomObjectType(NetBoxModel):
                 self._after_model_generation(attrs, model)
 
                 # _after_model_generation() can trigger a re-entrant get_model() for this
-                # SAME (cot_id, branch_id) -- e.g. a polymorphic field's
-                # related_object_types.all() rebuilding Django's relation tree, which
-                # re-enters our get_models() override before the #685/#686 reentrancy
-                # guard is set (this call arrived directly, not via get_models()'s own
-                # loop) and regenerates+re-registers a *different* class object under the
-                # same apps.all_models key (#688). That nested call's own _model_cache
-                # write is superseded below by this call's, but nothing previously
-                # re-asserted apps.all_models afterward, so it could keep pointing at the
-                # nested call's class while _model_cache (and this call's return value)
-                # point at this one. Re-register this call's own model so both agree.
+                # same (cot_id, branch_id) (#688) -- e.g. a polymorphic field's
+                # related_object_types.all() rebuilding the relation tree and re-entering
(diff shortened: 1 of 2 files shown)


bctiemann:

arthanson (approved):

PR #698, opened 2026-09-16
Fixes #697: Invalidate COT model cache when a referenced choice set changes

### LINKED ISSUE #697: Unable to select updated Custom Choice Set values with a Custom Object using a multiple selection field
### Plugin Version

0.7.0

### NetBox Version

4.7.1

### Python Version

3.12.13

### Steps to Reproduce

1. Click on Customization->Custom Field Choices and create new choice set.
2. Create a Custom Object then add a field with type "Multiple selection" that references the choice set created in step 1.
3. Create a new object of the type created above, adding one or more values from the existing choice set.  (this works as expected).
4. Click on Customization->Custom Field Choices and modify the custom choice set created in step one.
5. Add one or more new values.
6. Modify the custom object created in step 3 to attempt to add new values added from the choice set.  Click Save.
7. Instead of adding the value, an error pops up saying "Item x in the aray did not validate: Value 'x' is not a valid choice."

### Expected Behavior

The custom object should update to include the new choice set values.

### Observed Behavior

An error message prevents the change to the object from saving.  Modifications can still be made to the choice set as long as the changes only include the choice set values that existed when the custom object was defined.  For example if the choice set was created with values "one","two", and "three", but later changed to include "four" and "five", you can still modify the value to include any combination of "one", "two", and "three", however, any inclusion of "four" or "five" raises the error.

Image

### Suspected Cause

_No response_

### Proposed Fix

_No response_

bctiemann:
## Root cause

`SelectFieldType.get_model_field()` / `MultiSelectFieldType.get_model_field()` bake a `CustomFieldChoiceSet`'s current values into the generated Django model field's `choices=` kwarg at model-generation time. `CustomObjectType`'s model cache is invalidated (via `cache_timestamp`) whenever a `CustomObjectTypeField` is saved or deleted, but nothing was listening for the referenced `CustomFieldChoiceSet` itself changing — so adding a new value to a choice set already in use never invalidated any COT model that depends on it. Every already-cached model kept validating against the old value list, rejecting a legitimate new value with `"Value 'x' is not a valid choice."` until something unrelated happened to bump the COT's `cache_timestamp`.

Reproduced via three separate `manage.py shell` processes (simulating three independent HTTP requests — matching the issue's actual multi-step repro across separate page loads): create an object with a multiselect value, edit the choice set to add a new value, then edit the object to use the new value. Got the exact error message from the report.

## Fix

Added a `post_save` signal on `extras.CustomFieldChoiceSet` (`netbox_custom_objects/models.py`) that finds every `CustomObjectTypeField` referencing the saved choice set and, for each distinct owning COT: clears the in-process model cache and bumps `cache_timestamp` via `snapshot()` + `save(update_fields=['cache_timestamp'])` — the same pattern `clear_cache_on_field_save` already uses. Bumping `cache_timestamp` in the DB (not just clearing the local `_model_cache`) is what makes every worker process, not just the one that handled the choice-set edit, pick up the change on its next `get_model()` call.

## Testing

`ChoiceSetCacheInvalidationRegressionTest` in `test_models.py`:
- One test each for `select` and `multiselect` field types: populate the model cache, edit the choice set, confirm `get_model()` returns a genuinely different (regenerated) class, and confirm the new value now validates and saves.
- One test confirming an unrelated COT (whose fields don't reference the edited choice set) is *not* invalidated.

Both positive tests fail on unfixed code with the expected `AssertionError` (model wasn't regenerated); pass with the fix. The negative test passes both ways by design.

Full `test_models.py` (147 tests) and `test_field_types.py` (104 tests): no new failures (2 pre-existing, unrelated `netbox_branching` import errors in this dev environment's non-branching test config, confirmed present without this change).


bctiemann:

bctiemann:
Renamed the test class; I don't think the other two items really merit a change.

jnovinger (approved):
One thing I noticed while looking at some of the core custom field choice set machinery is that core guards against removing a custom field choice/value still referenced by any object, but it doesn't seem like we do here. I think it's more suitable for a follow up issue, rather than expanding scope here.

Core's removal guard in `CustomFieldChoiceSet.clean()` iterates `choices_for`, which is the related name on core `CustomField` only, so it never sees `CustomObjectTypeField` (`choices_for_object_type`). Removing an in-use value now regenerates the COT model promptly while existing rows keep the removed value, so those rows no longer validate. The upshot is that edits to these rows with now non-existent choices is impossible.

PR #710, opened 2026-09-17
Fixes: #699 - Rebase Custom Object viewsets onto NetBoxModelViewSet

### LINKED ISSUE #699: REST API viewsets bypass NetBox's permission enforcement, query optimization, and bulk-operation support
### Plugin Version

0.7.0

### NetBox Version

4.7.1

### Python Version

3.12

### Steps to Reproduce

1. Create a Custom Object Type with at least one Object (FK) field, one Multi-object (M2M) field, and confirm the type has `tags`/`owner` available (standard on every Custom Object).
2. As a non-superuser, create an `ObjectPermission` granting `view` on that Custom Object Type, constrained to a subset of instances (e.g. `{"name": "Visible"}`).
3. Create two instances — one matching the constraint, one not.
4. `GET /api/plugins/custom-objects/<type-slug>/` (the list endpoint) as that user.
5. Separately, `GET` the *detail* URL of the instance that should be hidden by the constraint.
6. Try `?brief=1`, `?fields=id`, and `?omit=tags` on the same list endpoint.
7. Try a bulk `PATCH`/`DELETE` against the list endpoint for `CustomObjectType`.

### Expected Behavior

- The list response in step 4 should contain only the constraint-matching instance, matching the detail-view behavior in step 5 (which correctly 404s).
- `?brief=`, `?fields=`, `?omit=` should trim the response the same way they do for every core NetBox model.
- Bulk PATCH/DELETE should work the same way it does for core `NetBoxModelViewSet`-based endpoints.

### Observed Behavior

- The list response in step 4 returns **both** instances (`count: 2`) — the permission constraint is silently ignored, even though the detail-view path for the same user and the same constraint correctly returns 404.
- `?brief=`, `?fields=`, and `?omit=` all return an identical, untrimmed response — they have no effect.
- Bulk PATCH/DELETE against `CustomObjectType`'s list endpoint returns a clean `405 Method Not Allowed` rather than performing the operation.
- Listing 12 rows of a Custom Object Type with 4 relational fields (FK, M2M, polymorphic object, polymorphic multi-object) plus `tags`/`owner` costs 119 SQL queries (measured via `CaptureQueriesContext`), scaling linearly at ~7 queries per additional row — extrapolating to roughly 7,000 queries for a single 1,000-row list response.

### Suspected Cause

`CustomObjectViewSet`, `CustomObjectTypeViewSet`, and `CustomObjectTypeFieldViewSet` (`netbox_custom_objects/api/views.py`) all subclass plain `rest_framework.viewsets.ModelViewSet` rather than `netbox.api.viewsets.BaseViewSet`/`NetBoxModelViewSet`. Confirmed directly: `api/views.py` imports `ModelViewSet` from `rest_framework.viewsets`, not from `netbox.api.viewsets`.

This one inheritance gap is the root cause of all four symptoms above, since `BaseViewSet` is where all of this machinery lives in core:

- `BaseViewSet.initial()` (`netbox/api/viewsets/__init__.py`) is what calls `queryset.restrict(user, action)` — its own docstring states it is "responsible for the enforcement of object-based permissions." None of the three viewsets in this plugin call `.restrict()` anywhere, on any action.
- The same `initial()`/`initialize_request()` path is what implements `?brief=`, `?fields=`, `?omit=` via `field_kwargs`.
- `get_queryset()` on `NetBoxModelViewSet` is what calls `get_annotations_for_serializer()` to apply automatic `select_related`/`prefetch_related` based on the serializer's declared relations — `CustomObjectViewSet.get_queryset()` (`api/views.py`) is a bare `self.model.objects.all()` with no optimization of any kind, which is the direct cause of the N+1 scaling.
- Bulk create/update/delete come from `BulkUpdateModelMixin`/`BulkDestroyModelMixin`, which `NetBoxModelViewSet` mixes in and none of these three viewsets do — yet `CustomObjectTypeViewSet`/`CustomObjectTypeFieldViewSet` are still routed through `NetBoxRouter` (`api/urls.py`), which unconditionally maps list-level PATCH/DELETE to `bulk_partial_update`/`bulk_destroy` regardless of whether the viewset actually implements them, producing the 405.

### Proposed Fix

Rebase all three viewsets onto `netbox.api.viewsets.BaseViewSet`/`NetBoxModelViewSet` (or the closest applicable core base for `CustomObjectViewSet`, which needs a dynamic per-COT model/serializer and may need a thin adapter). This single change should restore object-permission enforcement on list, `?brief=`/`?fields=`/`?omit=`, and bulk operations all at once. Query-count scaling will additionally need `select_related`/`prefetch_related` wired into `get_queryset()` for object/multiobject/tag/owner fields (the annotation mechanism `get_annotations_for_serializer()` uses may need COT-aware adaptation, since the serializer class is generated at runtime rather than being static).

Recommend adding query-count regression coverage for the REST list endpoint alongside this fix, the same way `tests/test_views.py` already tracks UI list-view costs in `query_counts.json` — `tests/test_api.py` currently has none, which is how this went uncaught.

bctiemann:
## Summary

`CustomObjectTypeViewSet`, `CustomObjectTypeFieldViewSet`, and `CustomObjectViewSet` inherited directly from DRF's `ModelViewSet` instead of NetBox's `NetBoxModelViewSet`. This bypassed several pieces of core NetBox REST API behavior:

- **Object-level permission enforcement on LIST** — `BaseViewSet.initial()`'s `queryset.restrict(user, action)` call never ran, so a user with a permission constrained to a subset of objects could still list every object of a given Custom Object Type.
- **Query optimization** — none of `get_prefetches_for_serializer()` / `get_annotations_for_serializer()`'s automatic `select_related`/`prefetch_related`/`annotate` resolution ran, so list views scaled with N+1 queries for related/tagged fields.
- **`?brief=`, `?fields=`, `?omit=`** — non-functional, since these are implemented in `BaseViewSet.get_serializer()`/`field_kwargs`.
- **Bulk create/update/destroy** — not exposed at all.

Rebasing all three viewsets onto `NetBoxModelViewSet` restores all of the above. `CustomObjectViewSet` resolves a different dynamically-generated Django model per request (keyed by the Custom Object Type slug in the URL), so its `get_queryset()` now memoizes that resolution, and a new `initial()` override forces it to run before `super().initial()` applies `queryset.restrict()` — otherwise permission restriction would be applied against an unresolved queryset. `urls.py`'s manually-wired list route also needed the same PUT/PATCH/DELETE → bulk-action verb mapping that `NetBoxRouter` already applies automatically to the other two viewsets.

Several now-redundant method overrides on `CustomObjectViewSet` (`list`, `create`, `update`, `partial_update`, `perform_destroy`) were removed, since `NetBoxModelViewSet` already provides equivalent (and in the update/destroy case, more complete — ETag/snapshot-aware) behavior.

**Known residual limitation (out of scope for this PR):** polymorphic multiobject fields still don't get automatic query optimization, since the custom `PolymorphicM2MDescriptor` backing them doesn't implement Django's prefetch-related protocol at all (unlike plain FK/M2M/GFK fields, which are fully fixed by this change). This is a narrower, separately-tracked follow-up.

## Test plan

- [x] `netbox_custom_objects.tests.test_api` (98 tests) passes against a freshly built test database.
- [x] Added/updated regression coverage:
  - `test_bulk_create_objects` / `test_bulk_delete_objects` / `test_bulk_update_objects` now exercise genuine single-request bulk operations (previously looped over individual per-object requests, with docstrings incorrectly claiming bulk endpoints didn't exist).
  - `test_list_objects_respects_object_level_permission_constraints` — new; verifies LIST now honors a constrained `ObjectPermission`.
  - `test_list_objects_brief` / `test_list_objects_fields` / `test_list_objects_omit` — new; verify `?brief=`/`?fields=`/`?omit=` now work on the list endpoint.
  - `test_list_objects_query_count_is_constant` — new; verifies the list endpoint's query count stays flat as row count grows (after warming the per-COT dynamic model/serializer class cache).
- [x] Full `netbox_custom_objects` suite run separately to check for regressions — 17 pre-existing failures observed, all in `test_schema_operations.py`/`test_polymorphic_fields.py`, all sharing the same known, pre-existing root cause (`netbox_branching`'s `ObjectChange` model raising a `RuntimeError` for a missing `app_label` when COT deletion/field-rename code paths run under a non-branching test config) — confirmed unrelated to this change, since neither failing test module references `api/views.py`, `api/urls.py`, or `CustomObjectViewSet` at all.


bctiemann:

bctiemann:
Addressed all three actionable findings in 19850e4:

1. **Slug-enumeration side-channel** — `initial()` no longer resolves the COT (or raises `Http404`) before `super().initial()` runs authentication. It now resolves without raising, falling back to an empty queryset for a bad slug, and the actual 404 is deferred to `get_queryset()` (reached only after auth/permission checks). Added `test_unauthenticated_request_does_not_leak_slug_existence` to lock this in.
2. **Missing permission tests for the two other rebased viewsets** — added `CustomObjectTypeAndFieldViewSetPermissionTest` covering `customobjecttype-list`/`customobjecttypefield-list` returning 403 for an unpermissioned authenticated user.
3. **Query-count test page-boundary sensitivity** — both requests in `test_list_objects_query_count_is_constant` now pass an explicit `?limit=100`.

`netbox_custom_objects.tests.test_api` is at 101/101 passing.

netbox_custom_objects/api/urls.py
```diff
@@ -8,9 +8,15 @@
...
-    {"get": "list", "post": "create"}
-)
+custom_object_list = views.CustomObjectViewSet.as_view({
+    "get": "list",
+    "post": "create",
+    # Matches NetBoxRouter's own bulk-operation mapping (netbox.api.routers.NetBoxRouter),
+    # since this list route is wired up manually rather than via that router.
+    "put": "bulk_update",
+    "patch": "bulk_partial_update",
+    "delete": "bulk_destroy",
```
pheus: Could we also exclude `bulk_update`, `bulk_partial_update`, and `bulk_destroy` in `@extend_schema_view`? Schema generation skips `initial()`, leaving `queryset` unset, so NetBox’s access to `queryset.model` breaks generation of the whole API schema.

Adding these alongside the existing exclusions and a schema-generation regression test should cover this.

netbox_custom_objects/api/views.py
```diff
@@ -147,64 +146,29 @@ def get_serializer_class(self):
...
+        # self.model/queryset must be resolved before super().initial() applies
+        # restrict(). Don't raise Http404 for a bad slug here though - auth hasn't
+        # run yet, and that would leak slug existence to an unauthenticated caller.
+        if self.model is None:
+            try:
+                custom_object_type = CustomObjectType.objects.get(
+                    slug=self.kwargs["custom_object_type"]
+                )
+                self.model = custom_object_type.get_model_with_serializer()
+                self.queryset = self.model.objects.all()
```
pheus: Could we explicitly reject `?background=true` for this viewset until background jobs can restore the custom-object type? NetBox 4.7’s `AsyncAPIJob` creates a fresh viewset with empty URL kwargs and skips `initial()`, so calling `.restrict()` on the unset queryset causes the accepted job to error.

A regression test for that rejection would help keep background support separate from this change.

netbox_custom_objects/api/views.py
```diff
@@ -134,7 +133,7 @@ class CustomObjectTypeViewSet(ModelViewSet):
     partial_update=extend_schema(exclude=True),
     destroy=extend_schema(exclude=True)
 )
```
pheus: Could we use a class-wide exclusion here? The new bulk actions are not covered by the existing decorator. Schema generation skips `initial()`, leaving `queryset` unset, so NetBox fails when generating their descriptions.

```suggestion
# Schema generation cannot resolve the dynamic model without a URL slug.
@extend_schema(exclude=True)
```

Please also remove the unused `extend_schema_view` import on line 12:

```diff
-from drf_spectacular.utils import extend_schema_view, extend_schema
+from drf_spectacular.utils import extend_schema
```

A schema-generation regression test would help cover this.

netbox_custom_objects/api/views.py
```diff
@@ -147,64 +146,29 @@ def get_serializer_class(self):
...
+        if self.model is None:
+            try:
+                custom_object_type = CustomObjectType.objects.get(
+                    slug=self.kwargs["custom_object_type"]
+                )
+                self.model = custom_object_type.get_model_with_serializer()
+                self.queryset = self.model.objects.all()
+            except CustomObjectType.DoesNotExist:
+                self.queryset = CustomObjectType.objects.none()
+        super().initial(request, *args, **kwargs)
```
pheus, line 156: Could we reject background bulk requests for now? NetBox 4.7’s `AsyncAPIJob` creates a fresh viewset without restoring the URL kwargs or calling `initial()`. The queryset therefore remains unset, and the accepted job errors when the worker calls `.restrict()`.

Overriding the enqueue hook would return a clear error without changing synchronous requests:

```suggestion
        super().initial(request, *args, **kwargs)

    def _enqueue_bulk_job(self, request, action, payload, action_kwargs=None):
        # AsyncAPIJob cannot restore this viewset's dynamic model.
        raise ValidationError(
            _("Background processing is not supported for custom objects.")
        )
```

A regression test confirming HTTP 400 and no job being enqueued would help cover this.

pheus (changes requested):
Thanks for this! Could we address the schema generation issue and reject unsupported background bulk requests before merging? I've included suggestions in the inline comments.

netbox_custom_objects/api/views.py
bctiemann: Fixed in f3775b5: replaced the per-action `extend_schema_view` list with a single class-wide `@extend_schema(exclude=True)`, which covers the bulk actions automatically (verified against drf-spectacular's source — a class-level `exclude=True` applies to every action, not just the ones matched by `methods=`). Removed the unused `extend_schema_view` import. Added `SchemaGenerationTest.test_api_schema_generates_without_error`, which hits the real `/api/schema/` endpoint.

netbox_custom_objects/api/views.py
bctiemann: Fixed in f3775b5: added a `_enqueue_bulk_job` override on `CustomObjectViewSet` that unconditionally raises `ValidationError` (matching your suggested approach). Added `test_background_bulk_update_rejected`, confirming a `?background=true` bulk PATCH returns 400 and doesn't create a Job row.

pheus (approved):

PR #712, opened 2026-09-17
Fixes: #711 - Prefetch polymorphic multi-object fields on the REST list endpoint

### LINKED ISSUE #711: Polymorphic Multi-object field causes N+1 queries in REST API
### Plugin Version

0.7.0

### NetBox Version

4.7.1

### Python Version

3.12

### Steps to Reproduce

1. Create a Custom Object Type with a polymorphic Multi-object field pointing at more than one target model.
2. Create several instances (e.g. 3, 6, 12) each populating the polymorphic multi-object field.
3. Query the type's REST API list endpoint (`GET /api/plugins/custom-objects/<slug>/`).
4. Compare total SQL query count (e.g. via `django.db.connection.queries` or `CaptureQueriesContext`) across the different row counts.

### Expected Behavior

Query count should stay flat as row count increases — matching the behavior now confirmed (after #699) for plain Object (FK) fields, plain Multi-object (M2M) fields, polymorphic Object (GenericForeignKey) fields, and `tags` on the same schema, all of which are now automatically batched via `get_prefetches_for_serializer()`'s `select_related`/`prefetch_related` resolution.

### Observed Behavior

Query count scales linearly with row count when a polymorphic multi-object field is requested:

| Rows | Query count |
|---|---|
| 3 | 28 |
| 6 | 34 |
| 12 | 46 |

That's roughly 2 extra queries per row. By contrast, the polymorphic Object (single, GFK-backed) field case is already flat (23 queries measured across the same 3/6/12 row counts) — `GenericForeignKey` is a native Django relation type, and `get_prefetches_for_serializer()` in `utilities/api.py` already includes it in its auto-prefetch-eligible `isinstance()` check, so it inherited the #699 fix automatically. The polymorphic multi-object case did not.

### Suspected Cause

Polymorphic multi-object fields are backed by `PolymorphicM2MDescriptor`/`PolymorphicManyToManyManager` in `netbox_custom_objects/field_types.py` — a custom, non-Django-native descriptor that isn't registered in Django's `_meta` field system and doesn't implement Django's prefetch-related protocol at all (unlike `GenericForeignKey`, which Django supports natively via `GenericForeignKeyDescriptor.get_prefetch_querysets()`). `get_prefetches_for_serializer()`'s introspection has no way to recognize or batch it, so every instance's polymorphic multi-object value is resolved with its own separate query per related object.

### Proposed Fix

Give `PolymorphicM2MDescriptor` an explicit prefetch mechanism — most likely a custom `Prefetch`-compatible interface, or a `get_prefetch_querysets()` implementation mirroring Django's native GFK protocol, batched per distinct target model the same way `GenericPrefetch` does for polymorphic GraphQL fields (see the equivalent GraphQL-side ticket, #705). This is a narrower, standalone follow-up to #699/NPL-1366 — the REST list endpoint's permission enforcement, `brief`/`fields`/`omit` support, and bulk operations are already fixed there; only this one polymorphic-multi-object query-optimization gap remains.

bctiemann:
## Summary

Follow-up to #699/#710: that fix made list-endpoint query optimization automatic for plain object/multi-object fields, tags, and polymorphic (GFK) object fields — but polymorphic multi-object fields were explicitly called out as a residual, narrower gap, tracked here.

`PolymorphicManyToManyManager` didn't implement Django's `prefetch_related()` protocol at all, so every instance's polymorphic multi-object value was resolved with its own separate query (or queries, one per distinct target model referenced by that instance). This PR:

- Adds `get_prefetch_querysets()`/`get_queryset()` to `PolymorphicManyToManyManager`, batching across *all* instances passed by Django's prefetch machinery in one query per distinct target content type (not per instance), using the same `extra()`-based join `CustomManyToManyManager` (the non-polymorphic case) already uses to duplicate each result row per relationship. `.all()` now checks the prefetch cache before falling back to the old per-instance path.
- Adds an explicit `.prefetch_related(...)` call in `CustomObjectViewSet.get_queryset()` for any polymorphic multi-object fields on the model, since `PolymorphicM2MDescriptor` isn't a registered Django relation field — `model._meta.get_field()` can't find it, so core's automatic `get_prefetches_for_serializer()` resolution never discovers it on its own (unlike the GFK case, where `GenericForeignKey` *is* a registered private field). This respects the request's `fields=`/`omit=` constraints the same way core's own resolution does.

## Test plan

- [x] `PolymorphicMultiObjectPrefetchTest.test_query_count_is_constant` — list-endpoint query count for a polymorphic multi-object field stays flat as row count grows (empirically verified flat at 26 queries across 12/24 rows; previously scaled ~2 extra queries/row, e.g. 28/34/46 at 3/6/12 rows).
- [x] `PolymorphicMultiObjectPrefetchTest.test_prefetched_results_are_scoped_per_instance` — each instance's serialized `targets` reflects only its own memberships (mixed content types, no cross-contamination) once results are prefetched in a single batch across instances.
- [x] `netbox_custom_objects.tests.test_api` (105 tests) passes against a freshly built test database.
- [x] Broader regression pass: `test_api`, `test_field_types`, `test_views`, `test_graphql` (438 tests) — 4 pre-existing failures observed, all sharing the same known, pre-existing root cause (`netbox_branching`'s `ObjectChange` app_label `RuntimeError` in `.delete()` code paths under a non-branching test config), confirmed unrelated since none of those tests touch `field_types.py` or `api/views.py`.


bctiemann:

bctiemann:
Addressed in 0e13d56:

1. **Custom queryset guard** — `get_prefetch_querysets()` now accepts a per-content-type queryset override instead of rejecting any non-empty `querysets`, mirroring `GenericForeignKey.get_prefetch_querysets()`'s own handling (one queryset per content type, raising only on a genuine duplicate for the same type).
2. **`get_queryset()` comment** — one correction here: Django's prefetch loop *does* call `manager.get_queryset()` per instance even under the plural `get_prefetch_querysets()` protocol (verified directly against this repo's installed `django/db/models/query.py::prefetch_one_level()` — it's unconditional for the `single=False` case, then immediately overwrites `_result_cache` regardless of what came back). The cache-check branch is still effectively unreachable in practice, just not for that reason — nothing else in this class calls `get_queryset()` directly (`.all()` checks the prefetch cache itself rather than delegating). Reworded the comment to say that accurately rather than repeat the "not called at all" framing.
3. **SQL interpolation** — added a comment explaining why the through-table PK list can't use `%s` (variable-length `IN`-list) and why it's safe (already-fetched instances' integer PKs, matching `CustomManyToManyManager`'s existing style).
4. **Missing `fields=`/`omit=` coverage** — added `test_omit_excludes_field_and_skips_its_prefetch`, confirming `?omit=targets` both drops the field from the response and skips its prefetch (lower query count than with the field included).

`test_api.py` is at 106/106.

arthanson (approved):

PR #714, opened 2026-09-17
Fixes: #700 - Enforce required on scalar fields at the model and REST layers

### LINKED ISSUE #700: Required fields are enforced only in the UI form, never at the model or REST API layer
### Plugin Version

0.7.0

### NetBox Version

4.7.1

### Python Version

3.12

### Steps to Reproduce

1. Create a Custom Object Type with a scalar field (e.g. a `text` field) marked "Required".
2. `POST` to the type's REST API list endpoint, omitting that field entirely.
3. `OPTIONS` the same list endpoint and inspect the field's `required` attribute.

### Expected Behavior

The `POST` in step 2 should be rejected, and `OPTIONS` in step 3 should report `required: true` for the field — matching the "Required" setting configured on the Custom Object Type Field.

### Observed Behavior

- `POST` succeeds with `201 Created`; the object is saved with the field `null`.
- `OPTIONS` reports `required: false` for the field.

This reproduces for every scalar field type: `text`, `longtext`, `integer`, `decimal`, `boolean`, `date`, `datetime`, `url`, `json`, `select`, `multiselect`, `coordinates` (12 of the plugin's 14 field types).

### Suspected Cause

Every scalar field type's `get_model_field()` in `netbox_custom_objects/field_types.py` hardcodes `null=True, blank=True` on the generated Django model field, regardless of the `CustomObjectTypeField.required` flag. Checked against the full commit history of this file — it has never been conditional on `required`. `field.required` is read only inside `get_form_field()`/`get_filterform_field()`, i.e. UI form construction — nowhere at the model or REST serializer layer. Only `ObjectFieldType` and `MultiObjectFieldType` correctly propagate `required=field.required` into the generated DRF serializer field; every other field type lets DRF's `ModelSerializer` auto-infer `required=False` from the model field's `blank=True`.

This may be related to #138 ("Required fields in custom objects are not enforced"), which was closed with the note that scalar values already worked correctly and only object fields didn't — the reverse of what's reproduced here. Worth checking whether that fix only ever touched the UI form layer, or whether this has regressed since.

### Proposed Fix

Make `get_model_field()` for each scalar field type set `blank=not field.required` (and `null` accordingly, or handle NOT NULL enforcement at the serializer/form layer instead, consistent with however NetBox core distinguishes "may be blank in the UI" from "may be omitted via the API" for its own required/optional fields). Add `required=field.required` to each scalar field's serializer-field construction to match the pattern already used for object/multiobject fields.

bctiemann:
## Summary

Every scalar field type (`text`, `longtext`, `integer`, `decimal`, `boolean`, `date`, `datetime`, `url`, `json`, `select`, `multiselect`, `coordinates`) hardcoded `blank=True` on its generated model field regardless of `CustomObjectTypeField.required`, and had no explicit `get_serializer_field()` of its own — so DRF's `ModelSerializer` auto-built a serializer field from the model field and always inferred `required=False`. `required` was therefore enforced only by the UI form, which constructs its own form field with `required=field.required` directly, independent of the model field.

- `get_model_field()` now sets `blank=not field.required` for every affected type. `null` stays `True` unconditionally — this plugin already commits to "toggling `required` must never risk an `ALTER TABLE ... SET NOT NULL`" (see `test_branching.py::test_field_required_toggle_merge`), and DRF's own `required` auto-inference treats a nullable column as non-required regardless of `blank` anyway, so leaving `null=True` and fixing `blank` alone wouldn't have been enough even if DB-level enforcement were in scope.
- Each affected type now implements `get_serializer_field()` explicitly, passing `required=field.required` (mirroring the pattern `ObjectFieldType`/`MultiObjectFieldType` already used correctly).
- `CoordinatesFieldType.get_serializer_field()` returns a dict (one entry per backing column — `<name>_latitude`/`<name>_longitude`), mirroring `get_model_field()`'s existing multi-column convention. `serializers.py`'s field-building loop now merges a dict return into `attrs` and exempts coordinates fields from the single-column `model_field_names` guard (neither backing column is literally named after the field itself).
- `URLFieldType`'s title column is unaffected by this — it stays always-optional by design (see the class docstring), only the URL column itself follows `required`.
- Existing default-value-on-omission behavior is preserved: the serializer still doesn't pass an explicit `default=` of its own, so a model-level default still applies when a key is simply absent from the payload (for a *non-required* field).

## Test plan

- [x] `RequiredFieldEnforcementTestCase` (new, `test_field_types.py`) — model-layer `blank` correctly follows `required` for all 9 plain scalar types, select/multiselect, both coordinates halves, and confirms the URL title column stays optional regardless.
- [x] `RequiredScalarFieldAPITest` (new, `test_api.py`) — REST-layer: `POST` omitting required fields returns 400 with per-field errors; `POST` with all required fields (or only the optional one omitted) returns 201; `OPTIONS` reports `required: true`/`false` correctly, including the URL title column.
- [x] `test_branching.py::test_field_required_toggle_merge` (existing, unmodified) — confirms toggling `required` still does not produce a DB-level `NOT NULL` constraint.
- [x] Full `netbox_custom_objects` suite (1284 tests) run against a freshly built database — 17 pre-existing failures observed, matching the already-established baseline exactly (`netbox_branching`'s `ObjectChange` app_label `RuntimeError` in `.delete()`/rename code paths under a non-branching test config), confirmed unrelated to this change.


bctiemann:

bctiemann:
Addressed all three in f846402:

1. **Missing text/longtext API coverage** — added `req_text`/`req_longtext` to `RequiredScalarFieldAPITest`'s fixture and both assertion loops; previously every test payload's only required text-type field was `name`, which every payload already satisfied.
2. **`allow_blank` bypassing choice validation** — removed `allow_blank=not field.required` from `SelectFieldType.get_serializer_field()`. Confirmed empirically: before the fix, `POST {"opt_select": ""}` on a non-required select field returned 201 and stored `""`; after, it correctly returns 400 (`'"" is not a valid choice.'`), while omitting the field or submitting a real choice still succeeds.
3. **Coordinates exception fallback wrong column name** — the `except Exception` fallback in `get_serializer_class()` now assigns the fallback `JSONField` under both `_latitude`/`_longitude` backing columns for coordinates fields instead of the non-existent `field.name`.

`test_api`/`test_field_types` are at 215/216 (1 pre-existing skip, expected — query-count baselines self-skip when `netbox-branching` is installed).

netbox_custom_objects/field_types.py
```diff
@@ -325,7 +326,14 @@ class TextFieldType(FieldType):
     def get_model_field(self, field, **kwargs):
         field_kwargs = self._safe_kwargs(**kwargs)
         field_kwargs.update({"default": field.default, "unique": field.unique})
-        return models.CharField(null=True, blank=True, **field_kwargs)
+        return models.CharField(null=True, blank=not field.required, **field_kwargs)
```
jnovinger, line 329: Flipping a field `optional -> required` strands existing rows that left it empty. Is that intentional?

`blank=False` gets checked on every `full_clean()`, against every field on the object, so editing the *label* on an old row now fails with "note: This field cannot be blank." Hits the API, bulk edit, the UI form, and branching merge.

`unique` already guards this exact case (`models.py:2919-2958`). Should `required` get the same pre-flight check?

jnovinger (commented):
`test_field_required_toggle_merge`'s docstring (`test_branching.py:2991-2993`) still says `field_types.py` "hardcodes `null=True, blank=True`" and that required is form-layer only. Both changed here. The test still passes because it only asserts `is_nullable`.

netbox_custom_objects/field_types.py
bctiemann: Fixed in 11ad686: added a pre-flight check to `CustomObjectTypeField.clean()`, mirroring the existing non-unique-values check right above it. Toggling optional->required now scans for any existing row with a blank value for this field (per Django's own `blank`/`empty_values` semantics — handles multi-column types like coordinates by checking each half independently, and correctly excludes a url field's always-optional title column) and rejects the toggle with a `ValidationError` if found, before it can strand that row. New tests in `test_models.py` cover: rejected-with-existing-blank-row, allowed-with-no-blank-rows, and the coordinates half-blank case.

bctiemann:
Addressed both in 11ad686:

1. **Required-toggle strands existing blank rows** — added a pre-flight check to `CustomObjectTypeField.clean()`, mirroring the existing non-unique-values check right above it (replied inline on that thread with details).
2. **Stale `test_field_required_toggle_merge` docstring** — updated to describe the current behavior (`blank`/serializer `required` enforce it; `null=True` unconditionally, not `blank=True`) instead of the pre-#700 state. The test's own assertion (DB column stays nullable) was already correct and unchanged.

netbox_custom_objects/models.py
```diff
@@ -2959,6 +2959,42 @@ def clean(self):
...
+            and not self._state.adding
+            and hasattr(self, '_original')
+            and not self.original.required
+        ):
+            model = self.custom_object_type.get_model()
+            model_field = FIELD_TYPE_CLASS[self.type]().get_model_field(self)
+            columns = model_field if isinstance(model_field, dict) else {self.name: model_field}
+            # Only columns actually made non-blank by this field's required flag
+            # (e.g. a url field's title column always stays blank=True).
+            required_columns = {name: f for name, f in columns.items() if not f.blank}
```
jnovinger: Toggling a polymorphic multiobject field optional -> required raises `AttributeError`.

Was able to reach it in the browser with this STR:

1. Create a Custom Object Type.
2. Add a field: type `text`, Primary, Required.
3. Add a field: type `multiobject`, check Polymorphic, select one or more related object types, leave Required unchecked.
4. Edit that multiobject field, check Required, save.

Step 4 returns a 500. No custom object instances needed: it fails while building `required_columns`, before any row scan.

```
File "netbox_custom_objects/models.py", line 2981, in clean
    required_columns = {name: f for name, f in columns.items() if not f.blank}
AttributeError: 'PolymorphicM2MDescriptor' object has no attribute 'blank'
```

Same via the API: `PATCH /api/plugins/custom-objects/custom-object-type-fields/<id>/` with `{"required": true}`.

netbox_custom_objects/models.py
```diff
@@ -2959,6 +2959,42 @@ def clean(self):
...
+            if required_columns:
+                existing = model.objects.values_list(*required_columns.keys())
+                for row in existing.iterator():
```
jnovinger, line 2998: Claude is telling me this is going to require a full-table scan while there is potentially an index-covered query available, that looks something like `model.objects.filter(Q(note__isnull=True) | Q(note='')).exists()`. Measured 11.3ms against 5k populated rows, so it only matters on large types. Also, it would appears that some types (like JSON and multiselect) would need special handling.

Leaving this one as an idea, not a requirement or even necessarily a suggestion.

jnovinger (changes requested):
Object and multiobject fields still hardcode `blank=True` (`field_types.py:932`, `:937`, `:982`, `:1595`), so `required` on those two types stays REST-only. The scalar types now enforce at both layers, which leaves the same "Required" checkbox meaning two different things depending on field type.

#700 is written against the scalar types, so this may be deliberate scope. Worth a follow-up issue if so.

netbox_custom_objects/models.py
bctiemann: Confirmed and fixed in b74fc44, thanks for the exact STR. Root cause: `get_model_field()` for a polymorphic multiobject returns a bare `PolymorphicM2MDescriptor` (no `.blank`), and for a polymorphic object it returns a dict that includes a `GenericForeignKey` entry alongside its two real backing columns — a real `Field`, with `.blank=False` by default, but not itself a directly queryable column (`values_list()` can't resolve it, hence the `FieldError` you'd hit on plain object/multiobject... actually on the polymorphic object case specifically, per my own follow-up testing).

`required_columns` now excludes anything without a usable `.blank`, and `GenericForeignKey` specifically, treating both the same as `ObjectFieldType`/`MultiObjectFieldType`'s own concrete fields (which already hardcode `blank=True` regardless of `required`) — i.e. not checkable by this pre-flight check at all. Empirically re-verified your exact repro (polymorphic multiobject, Required checkbox, save) plus polymorphic object, plain object, and plain multiobject — all four now toggle cleanly. Added `test_required_toggle_does_not_crash_for_relationship_fields` covering all four.

bctiemann:
Both addressed:

1. **Crash on polymorphic relationship fields** — fixed in b74fc44 (replied inline with details). Root cause: `get_model_field()` returns a bare `PolymorphicM2MDescriptor` for polymorphic multiobject (no `.blank` attribute → `AttributeError`), and a dict including a `GenericForeignKey` entry for polymorphic object (a real `Field`, but not a directly-queryable column → `FieldError` from `values_list()`). Both are now explicitly excluded from the pre-flight check, matching how plain object/multiobject's own concrete fields (hardcoded `blank=True` always) were already excluded. Verified all four relationship-field variants (plain object, plain multiobject, polymorphic object, polymorphic multiobject) toggle cleanly now, with a new test covering all four.
2. **Object/multiobject model-layer scoping** — confirmed deliberate: #700's own reproduction excluded these two types from the start, since their *serializer*-level `required` was already correct pre-#700 (only scalar types had that gap). Filed as a follow-up: [#719](https://github.com/netboxlabs/netbox-custom-objects/issues/719) / Linear NPL-1383, linked under the same REST-audit parent as this ticket.

netbox_custom_objects/models.py
bctiemann: Thanks for checking into this — I think it's probably OK to leave as-is for now, since fixing it would be complex for the benefit, and this isn't a hot-path action (it's an infrequent admin operation, changing the `required` state on an existing field). We can address it if it surfaces as a real-world issue.

jnovinger (approved):

PR #716, opened 2026-09-17
Fixes: #715 - Enforce integer/decimal min/max at the model and REST layers

### LINKED ISSUE #715: Integer/Decimal min/max validation is enforced only in the UI form, never at the model or REST API layer
### Plugin Version

0.7.0

### NetBox Version

4.7.1

### Python Version

3.12

### Steps to Reproduce

1. Create a Custom Object Type with an `integer` or `decimal` field, setting "Minimum Value" and/or "Maximum Value".
2. `POST` to the type's REST API list endpoint with a value outside that range.
3. Create an instance directly via the ORM (e.g. `model.objects.create(...)`) with a value outside that range, and call `full_clean()` on it.

### Expected Behavior

Both the `POST` in step 2 and `full_clean()` in step 3 should reject the out-of-range value — matching the "Minimum Value"/"Maximum Value" configured on the Custom Object Type Field, and matching the UI form's own behavior (which already enforces this correctly).

### Observed Behavior

Both succeed and persist the out-of-range value. `IntegerFieldType.get_model_field()` and `DecimalFieldType.get_model_field()` never attach `MinValueValidator`/`MaxValueValidator` to the generated model field, and neither type's `get_serializer_field()` passes `min_value`/`max_value` to the DRF field either — only `get_form_field()` does (`min_value=field.validation_minimum, max_value=field.validation_maximum`), so enforcement is UI-form-only, the same underlying pattern as #700, just for range validation instead of required.

### Suspected Cause

Same root cause shape as #700: `field.validation_minimum`/`field.validation_maximum` are read only inside `get_form_field()`. Neither `get_model_field()` (as `validators=[...]`) nor `get_serializer_field()` (as `min_value=`/`max_value=` kwargs, which `rest_framework.serializers.IntegerField`/`DecimalField` both accept natively) propagate them.

### Proposed Fix

Add `MinValueValidator(field.validation_minimum)`/`MaxValueValidator(field.validation_maximum)` (guarded by `is not None`) to `IntegerFieldType.get_model_field()`/`DecimalFieldType.get_model_field()`'s `validators=[...]`, and pass `min_value=field.validation_minimum, max_value=field.validation_maximum` to both `get_serializer_field()` implementations (added in #700), matching the pattern their own `get_form_field()` already uses.

bctiemann:
## Summary

Follow-up to #700, raised during that work: `IntegerFieldType`/`DecimalFieldType.get_model_field()` never attached `MinValueValidator`/`MaxValueValidator` to the generated model field, and their `get_serializer_field()` (added in #700) never passed `min_value`/`max_value` to the DRF field either — only `get_form_field()` enforced `validation_minimum`/`validation_maximum`. Same underlying gap-shape as #700, just for range validation instead of required.

**Note:** this PR targets the `700-required-fields-not-enforced` branch, not `main` — it builds directly on the `get_serializer_field()` methods #700/#714 introduces, which aren't on `main` yet. Retarget to `main` once #714 merges.

- Both `get_model_field()` implementations now append `MinValueValidator`/`MaxValueValidator` (guarded by `is not None`) to the model field's `validators`.
- Both `get_serializer_field()` implementations pass `min_value=field.validation_minimum, max_value=field.validation_maximum` directly — DRF's `IntegerField`/`DecimalField` already no-op when either is `None`, so no extra guarding is needed on that side.

## Test plan

- [x] `test_integer_field_min_max_enforced_by_full_clean` / `test_decimal_field_min_max_enforced_by_full_clean` (new, `test_field_types.py`) — model-layer `full_clean()` rejects out-of-range values, accepts in-range ones.
- [x] `IntegerDecimalMinMaxAPITest` (new, `test_api.py`) — REST-layer: `POST` above/below the configured range returns 400 with per-field errors; `POST` within range returns 201.
- [x] Empirically verified both the model (`full_clean()`) and REST layers reject out-of-range values on both bounds, for both types.
- [x] Broader regression pass (`test_api`, `test_field_types`, `test_forms`, `test_views` — 427 tests) — 4 pre-existing errors observed, matching the already-established `netbox_branching` app_label baseline, confirmed unrelated.


bctiemann:

bctiemann:
Addressed both in d51453d:

1. **`validators=[]` overwriting caller-supplied validators** — `field_kwargs["validators"]` is now only set when the computed list is non-empty, so a `validators` kwarg passed in via `**kwargs` (none currently do, but as you noted, the pattern was a latent trap) is preserved when neither `validation_minimum` nor `validation_maximum` is set.
2. **Boundary coverage** — added exact-boundary assertions (`count=0`/`count=100`, `price=Decimal("0.00")`/`Decimal("100.00")`) to both `full_clean()` tests, plus a new `test_post_at_exact_boundaries_returns_201` in the REST test covering the same for both int and decimal.

14/14 in the affected test classes.

jnovinger (approved):

PR #718, opened 2026-09-17
Fixes: #708 - Represent select/multiselect fields as {value, label} in REST and GraphQL

### LINKED ISSUE #708: Select/multiselect fields return the raw value instead of {value, label} in REST and GraphQL
### Plugin Version

0.7.0

### NetBox Version

4.7.1

### Python Version

3.12

### Steps to Reproduce

1. Create a Custom Object Type with a `select` or `multiselect` field backed by a Custom Field Choice Set.
2. Create an instance setting a value on that field.
3. Fetch the instance via the REST API, and separately via GraphQL.
4. Compare the field's representation against a core NetBox model's choice field (e.g. `dcim.Site.status`) in the same API.

### Expected Behavior

A select/multiselect field should be represented as `{"value": ..., "label": ...}` (REST) / an equivalent labeled shape (GraphQL) — matching every core NetBox `ChoiceField`'s convention, so a client doesn't need a separate lookup to display the human-readable label.

### Observed Behavior

Both APIs return the raw stored value only:
- REST: no `{value, label}` wrapping — the field is auto-built by DRF directly from the model's `CharField(choices=...)`/`ArrayField`, which `SelectFieldType`/`MultiSelectFieldType` never override with a labeled representation.
- GraphQL: confirmed in the live schema's SDL — `value: String`, `values: [String!]`, plain scalars, read straight from each field type's `get_graphql_annotation()`.

This is one underlying gap surfacing in both APIs, not two independent ones — a single fix in the shared field-type layer should cover both.

### Suspected Cause

`SelectFieldType`/`MultiSelectFieldType` in `netbox_custom_objects/field_types.py` don't override serializer-field or GraphQL-annotation construction to wrap the value in a labeled object the way core's `ChoiceField` (REST) does.

### Proposed Fix

Add a labeled representation for select/multiselect fields in both the REST serializer field and the GraphQL type/resolver, matching core's `{value, label}` convention.

bctiemann:
## Summary

`SelectFieldType`/`MultiSelectFieldType` never wrapped their stored value in a labeled representation, unlike every core NetBox `ChoiceField` (e.g. `dcim.Site.status`) — both APIs returned the raw stored string(s) directly. One underlying gap, surfacing in both REST and GraphQL, as the issue itself notes.

**Note:** this PR targets the `715-integer-decimal-min-max-validation` branch (itself stacked on `700-required-fields-not-enforced`/#714), not `main` — it builds on the `get_serializer_field()` methods those introduce. Retarget to `main` once the earlier PRs in the stack merge.

- **REST**: `SelectFieldType.get_serializer_field()` now returns `netbox.api.fields.ChoiceField` — core's own field that represents `{value, label}` on read and still accepts a bare value on write — instead of DRF's plain `ChoiceField`. `MultiSelectFieldType`'s `ListField` uses the same core `ChoiceField` as its child, so each item in the list reads the same way.
- **GraphQL**: added a small `CustomObjectChoiceType` (`value`, `label`) and a resolver for `SELECT`/`MULTISELECT` fields, mirroring core's `CustomField.resolve_selection_value()` almost exactly, and reusing this plugin's own `CustomObjectTypeField.get_choice_label()` — the same helper already backing the UI table column (`SelectFieldType.get_table_column_field()`) and `get_display_value()`.
- Updated `test_scalar_fields_query`, whose existing assertion on the raw `"status"` value is now intentionally outdated by this fix.

## Test plan

- [x] `SelectMultiSelectValueLabelAPITest` (new, `test_api.py`) — REST GET returns `{value, label}` for `select` and a list of such dicts for `multiselect`; POST still accepts bare values on write.
- [x] `test_multiselect_field_query_returns_value_label_pairs` (new, `test_graphql.py`) — GraphQL query returns `{value, label}` per item for a multiselect field.
- [x] `test_scalar_fields_query` (updated) — GraphQL `select` field now queried/asserted as `{value, label}`.
- [x] Empirically verified both layers end-to-end: REST GET/POST round-trip, and the actual built Strawberry type's resolver called directly against a real instance.
- [x] Full `test_graphql` suite (32 tests, including the schema-assembly smoke test) passes.
- [x] Broader regression pass (`test_api`, `test_graphql`, `test_field_types`, `test_views` — 455 tests) — 4 pre-existing errors observed, matching the already-established `netbox_branching` app_label baseline, confirmed unrelated.


bctiemann:

bctiemann:
Addressed both in 63405a8:

1. **Stale `graphql_annotation`** — removed `= str`/`= List[str]` from `SelectFieldType`/`MultiSelectFieldType` entirely (matching `ObjectFieldType`/`MultiObjectFieldType`, which never set it either, for the same reason: GraphQL resolution goes through a dedicated resolver, not `get_graphql_annotation()`). Dropped the now-unused `List` import.
2. **Null-value coverage** — added `test_select_field_reads_as_none_when_unset` and `test_multiselect_field_reads_as_none_when_unset`. Worth noting: an unset multiselect also reads back as `None`, not `[]` — DRF's `Serializer.to_representation()` short-circuits to `None` before ever calling the field's own `to_representation()` when the underlying attribute is `None`, so this is inherent to how the model field's own default works, unaffected by this fix either way.

259/260 in the affected test classes (1 pre-existing skip, expected).

netbox_custom_objects/field_types.py
```diff
@@ -859,12 +862,14 @@ def get_model_field(self, field, **kwargs):
...
-        # already-generated model field rather than re-querying field.choice_set.
+        # already-generated model field rather than re-querying field.choice_set,
+        # and use core's own ChoiceField as the child so each item reads as
+        # {value, label} instead of the raw value (still writes a bare value).
         choices = (
             model._meta.get_field(field.name).base_field.choices if model else field.choices
         )
         return drf_serializers.ListField(
-            child=drf_serializers.ChoiceField(choices=choices),
+            child=NetBoxChoiceField(choices=choices),
```
jnovinger, line 872: Core's [`ChoiceField`](https://github.com/netbox-community/netbox/blob/abccf4e03650fecd366e676b392729320258c49e/netbox/netbox/api/fields.py#L76-L90) coerces before it checks membership (`netbox/api/fields.py:77-86`): `"1"` becomes `1`, `"true"` becomes `True`. Choice values are stored as strings, so any choice set with numeric or boolean-looking values stops accepting writes.

The API rejects its own output:

```
GET  -> {"value": "1", "label": "1"}
PUT  {"status": "1"}  -> 400  "1 is not a valid choice."
```

Core gets the same `{value, label}` shape for user-defined choices without `ChoiceField`: [`resolve_selection_value()`](https://github.com/netbox-community/netbox/blob/abccf4e03650fecd366e676b392729320258c49e/netbox/extras/models/customfields.py#L387-L399) for read (shared by REST and GraphQL, per netbox-community/netbox#20897) and a [plain membership test](https://github.com/netbox-community/netbox/blob/abccf4e03650fecd366e676b392729320258c49e/netbox/extras/models/customfields.py#L1196-L1204) for write. `ChoiceField` appears to be for developer-defined enums where values are controlled.

Every test fixture here uses alpha slugs, so tests pass either way.

jnovinger (changes requested):

netbox_custom_objects/field_types.py
bctiemann: Good catch — confirmed. Core's `ChoiceField.to_internal_value()` coerces `"1"` → `int(1)` / `"true"` → `True` before the `self._choices` membership check, which is keyed by the raw string. It's actually worse than just round-tripping: a fresh `POST {"priority": "1"}` fails outright with `400 "1 is not a valid choice."`.

Fixed in c6bd376 by adding `SafeChoiceField` (`api/serializers.py`), a subclass of core's `ChoiceField` that keeps the same `{value, label}` read representation but overrides `to_internal_value()` to drop the numeric/boolean coercion step — everything else (including `to_representation()`) is inherited unchanged. `SelectFieldType`/`MultiSelectFieldType.get_serializer_field()` now use it (local imports to avoid a circular import with `field_types.py`).

Verified empirically against your exact repro:
```
POST priority="1" -> 201 {'value': '1', 'label': 'One'}
GET priority -> {'value': '1', 'label': 'One'}
PATCH priority='1' (round-tripped from GET) -> 200 {'value': '1', 'label': 'One'}
```

Also added regression tests covering numeric- and boolean-looking choice values on both select and multiselect, plus confirming out-of-set values are still correctly rejected (400).

bctiemann:
Pushed c6bd376 addressing Jason's finding on the choice-field coercion bug (inline reply [here](https://github.com/netboxlabs/netbox-custom-objects/pull/718#discussion_r4065083835)).

Summary: core's `netbox.api.fields.ChoiceField` coerces numeric-/boolean-looking string input to `int`/`bool` before checking choice membership, which breaks any choice set with a value like `"1"` or `"true"` — the API rejects its own GET output on write. Added `SafeChoiceField`, a `ChoiceField` subclass that keeps the `{value, label}` read shape but drops that write-side coercion, and swapped it in for `SelectFieldType`/`MultiSelectFieldType`. Verified empirically and added regression tests for both select and multiselect (numeric string, boolean-like string, and confirming out-of-set values are still rejected).

jnovinger (approved):

jnovinger (approved):

